PR #23, opened 2016-07-24
Add basic users endpoint

joshsmith:
This adds a basic endpoint for users, with tests!

I added in:
- Guardian, although it's not being used yet
- `Comeonin` with `BCrypt` for hashing passwords; later can be used to authenticate
- Users with:
  - Migration
  - Virtual `password` field in schema
  - Controller tests
- JSON API serialization with:
  - Error view
  - Changeset view
  - User view

It's important to note that by default Phoenix uses the JSON API recommended dasherized keys. We should probably switch to dasherized keys in our API as a result.

This is still missing many of the user model's attributes and validations, slugged route creation on signup, etc. Also missing some model tests.

I figure it's better to keep this PR smaller than not, so many of these things can be added later.


test/controllers/user_controller_test.exs
```diff
@@ -0,0 +1,118 @@
...
+      "data" => %{
+        "type" => "user",
+        "attributes" => attrs,
+        "relationships" => relationships
+      }
+    }
+
+    id = json_response(conn, 201)["data"]["id"]
+    assert id
+    assert Repo.get(User, id)
```
npendery, line 63: Should we also assert that the `conn` path is currently at `users/{:id}` for the created user? The test title sounds like it should check that the page rendered is for the created resource. Aren't both the current assertions just checking that it was created?

joshsmith, line 63: I'm not sure. This is what the generators provided so I'm guessing that's considered a standardized approach?

npendery, line 63: Okay cool! It might also just be how Im thinking of the word render as returning a specific page/url. So expecting to see an assertion verifying that this is the case.

joshsmith, line 63: I think by `render` they mean the `json_response`. I think the integration testing here is a little different than we're used to. Still adjusting.

barruumrex, line 63: Yeah. These tests are meant to specifically test the controller. Testing that the router sends requests to the correct controllers can also be done. This page is a bit outdated but you can see how they do it using the conn method (which I'm pretty sure changed to build_conn since then).  https://semaphoreci.com/community/tutorials/test-driven-apis-with-phoenix-and-elixir

joshsmith, line 63: @barruumrex is there a need to do that? I guess we really need to wrap our heads around how isolated we should be testing things and how far up the stack we need to go.

There are also big differences between model testing here than there would be in Rails. Like testing the Repo only when you have side effects and doing straight "model" unit tests otherwise.

Definitely requires adjusting.

barruumrex, line 63: Oh. Sorry. I should have been more clear. I actually don't generally test the router. I was just pointing out how it would be done if you did want to do that. 

Leaving the controller test specifically testing the controller seems to be pretty standard.

joshsmith, line 63: @barruumrex ah, okay! I got you. I think you're right then.

joshsmith:
@barruumrex How does the rest of the PR look to you? I'd like to add new things separately, but if things need fixed here, then I'll obviously do that in this PR.

Also would really like to see about how to transition from our Rails-esque Postgres DB to the Phoenix-esque one. I think there are bound to be differences in naming, etc. Need to think that through more.

rjsamson:
Looks pretty solid to me 👍

barruumrex:
Looks pretty good to me. The controller testing feels fat to me (the tests go all the way down to the db and back), but that's not really a problem at this stage in the project. Definitely something we can put off until it becomes a problem in practice. 

As far as converting the database, I don't expect to see much issue with that. Ecto is pretty flexible and shouldn't have difficulty dealing with an AR style db. However, I don't have direct experience with making that transition.

PR #44, opened 2016-08-02
Split out test.watch into watch container

joshsmith:
This splits out a `watch` container and makes `test` still use the `test` `MIX_ENV` so the database gets spun up correctly.


rigelstpierre:
Looks good. 🐑 🇮🇹

Issue #42, opened 2016-08-01
Add Users API


joshsmith:
Part of #26.

This can be split out into multiple issues to reflect the endpoints in the API docs: http://docs.codecorpsapidevelop.apiary.io/#reference/categories
- [x] Create
- [ ] Filter
- [x] Retrieve
- [x] Check email availability
- [x] Check username availability
- [x] Update

I think we should get rid of `/user/` and `/me` endpoints. Let's not map those over and instead use the normal `users/:id`, even for our authenticated user.

begedin:
~~@JoshSmith Is this up to date? From what I can tell, on the development branch, "update", "retrieve" and "create" all work. I clicked those in, but feel free to revert if there's something missing there.~~

Nevermind. I only just figured out we're actually missing some of the fields.

PR #47, opened 2016-08-05
Added "email_available" and "username_available" endpoints to Users API

begedin:
Part of a solution for #42 

Still new at this, so I'm not really sure if the placement for these is correct, but I wanted to start with a smaller problem to deal with.

### DIFF
--- a/test/controllers/user_controller_test.exs
+++ b/test/controllers/user_controller_test.exs
@@ -137,4 +137,53 @@ defmodule CodeCorps.UserControllerTest do
     assert response(conn, 204)
     refute Repo.get(User, user.id)
   end
+
+  describe "email_available" do
+    test "returns valid and available when email is valid and available", %{conn: conn} do
+      resp = get conn, user_path(conn, :email_available, %{email: "available@mail.com"})
+      json = json_response(resp, 200)
+      assert json["available"]
+      assert json["valid"]
+
+    end
+
+    test "returns valid but inavailable when email is valid but taken", %{conn: conn} do
+      insert_user(%{email: "used@mail.com"})
+      resp = get conn, user_path(conn, :email_available, %{email: "used@mail.com"})
+      json = json_response(resp, 200)
+      refute json["available"]
+      assert json["valid"]
+    end
+
+    test "returns as available but invalid when email is invalid", %{conn: conn} do
+      resp = get conn, user_path(conn, :email_available, %{email: "not_an_email"})
+      json = json_response(resp, 200)
+      assert json["available"]
+      refute json["valid"]
+    end
+  end
+
+  describe "username_available" do
+    test "returns as valid and available when username is valid and available", %{conn: conn} do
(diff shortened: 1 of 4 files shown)


test/controllers/user_controller_test.exs
```diff
@@ -115,4 +115,42 @@ defmodule CodeCorps.UserControllerTest do
...
+    resp = get conn, user_path(conn, :username_available, %{username: "used"})
+    json = json_response(resp, 200)
+    refute json["available"]
+    assert json["valid"]
+
+    resp = get conn, user_path(conn, :username_available, %{username: ""})
+    json = json_response(resp, 200)
+    assert json["available"]
+    refute json["valid"]
+  end
```
joshsmith, line 188: Following our conversation earlier, I think I would wrap the current tests as `describe` blocks for the endpoint and then separate each of the three tests (valid/available, invalid/available, valid/unavailable) into their own `test` blocks.

web/models/user.ex
```diff
@@ -40,4 +52,19 @@ defmodule CodeCorps.User do
...
+
+  defp check_username_valid(struct, username) do
+    put_in struct[:valid], String.length(username) >= 1 && String.length(username) <= 39
+  end
+
+  defp check_used(struct, column, value) do
+    query = from u in "users", where: field(u, ^column) == ^value, select: field(u, ^column)
+    available = CodeCorps.Repo.all(query) |> Enum.empty?
+
+    put_in struct[:available], CodeCorps.Repo.all(query) |> Enum.empty?
```
joshsmith: This looks repeated. Did you mean to use `available` here?

web/router.ex
```diff
@@ -22,7 +22,10 @@ defmodule CodeCorps.Router do
   scope "/", CodeCorps, host: "api." do
     pipe_through :api
 
+    get "/users/email_available", UserController, :email_available
+    get "/users/username_available", UserController, :username_available
     resources "/users", UserController, except: [:new, :edit]
+
```
joshsmith: Random extra space. I would also maybe place the `get` below the `resources "/users"`.

begedin: If it goes bellow, then the matcher matches with "GET /users/:id", where id is "email_available" or "username_available"

test/controllers/user_controller_test.exs
```diff
@@ -115,4 +115,55 @@ defmodule CodeCorps.UserControllerTest do
     assert response(conn, 204)
     refute Repo.get(User, user.id)
   end
+
+  describe "email_available" do
+    test "returns { available: true, valid: true } when email is valid and available", %{conn: conn} do
```
joshsmith: The `returns { available: true, valid: true }` seems to do extra work of describing internals of the test, instead of just generalizing its conditions.

web/router.ex
joshsmith: Ah, I see.

PR #48, opened 2016-08-05
Add first_name, last_name, twitter, website, biography, to user, enable updating them

begedin:
Part of a solution for #42 

I'm doing this in smaller chunks, for easier reviews. There's a lot of work to be done.

### DIFF
--- a/priv/repo/migrations/20160805132301_add_user_profile_fields.exs
+++ b/priv/repo/migrations/20160805132301_add_user_profile_fields.exs
@@ -1,0 +1,13 @@
+defmodule CodeCorps.Repo.Migrations.AddUserProfileFields do
+  use Ecto.Migration
+
+  def change do
+    alter table(:users) do
+      add :first_name, :string
+      add :last_name, :string
+      add :website, :string
+      add :twitter, :string
+      add :biography, :string
+    end
+  end
+end
--- a/web/models/user.ex
+++ b/web/models/user.ex
@@ -2,10 +2,15 @@ defmodule CodeCorps.User do
   use CodeCorps.Web, :model
 
   schema "users" do
-    field :username, :string
-    field :email, :string
-    field :password, :string, virtual: true
+    field :biography, :string
     field :encrypted_password, :string
+    field :email, :string
+    field :first_name, :string
+    field :last_name, :string
+    field :password, :string, virtual: true
+    field :twitter, :string
+    field :username, :string
+    field :website, :string
 
     timestamps()
   end
@@ -32,6 +37,15 @@ defmodule CodeCorps.User do
     |> put_pass_hash()
   end
 
+  def update_changeset(struct, params) do
+    struct
+    |> changeset(params)
+    |> cast(params, [:first_name, :last_name, :twitter, :biography, :website])
+    |> prefix_url(:website)
+    |> validate_format(:website, ~r/\A((http|https):\/\/)?[a-z0-9]+([\-\.]{1}[a-z0-9]+)*\.[a-z]{2,}(([0-9]{1,5})?\/.*)?#=\z/ix)
+    |> validate_format(:twitter, ~r/\A[a-zA-Z0-9_]{1,15}\z/)
+  end
+
   defp put_pass_hash(changeset) do
     case changeset do
       %Ecto.Changeset{valid?: true, changes: %{password: pass}} ->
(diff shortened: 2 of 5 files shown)


priv/repo/migrations/20160805132301_add_user_profile_fields.exs
```diff
@@ -0,0 +1,13 @@
...
+  def change do
+  	alter table(:users) do
+      add :first_name, :string
+      add :last_name, :string
+      add :website, :string
+      add :twitter, :string
+      add :biography, :string
+  	end
+  end
+end
```
joshsmith, line 13: Formatting looks a little wonky here.

web/models/user.ex
```diff
@@ -7,6 +7,13 @@ defmodule CodeCorps.User do
     field :password, :string, virtual: true
     field :encrypted_password, :string
 
+    field :first_name, :string
+    field :last_name, :string
+
+    field :twitter, :string
+    field :biography, :string
+    field :website, :string
+
```
joshsmith, line 14: Any reason we're breaking up the fields into groups? Also may consider alpha ordering.

web/models/user.ex
```diff
@@ -32,6 +39,14 @@ defmodule CodeCorps.User do
     |> put_pass_hash()
   end
 
+  def update_changeset(struct, params) do
+    struct
+    |> changeset(params)
+    |> cast(params, [:first_name, :last_name, :twitter, :biography, :website])
+    |> validate_format(:website, ~r/\A((http|https):\/\/)?[a-z0-9]+([\-\.]{1}[a-z0-9]+)*\.[a-z]{2,}(([0-9]{1,5})?\/.*)?#=\z/ix)
+    |> validate_format(:twitter, ~r/\A[a-zA-Z0-9_]{1,15}\z/)
+  end
```
joshsmith, line 47: What does `validate_format` do when the values are blank? In other words, if I try to update my profile but never added a website or Twitter, will it fail?

joshsmith:
Are we able to add in a URL prefixer into the mix in this PR? We can open a separate issue if you'd like. Should be a fairly simple thing like `put_pass_hash()`.

web/models/user.ex
begedin, line 14: To me, it's easier to grasp which fields we have when they're grouped together in some sensible way. Alphabetising makes sense when there are too many fields to do any other type of ordering.

joshsmith, line 14: If you're going for consistency I think the simplest consistency works best, otherwise there's always the question of when we reach too many, and then having to do a large refactoring to meet those constraints.

I find if I'm reading/scanning code, then alpha sub-ordering almost always makes it easier.

web/models/user.ex
begedin, line 47: `validate_format` passes if the string is blank or the parameter is not in the map. 

Additionally
- If the parameter is a blank string, it will update the model with that blank string
- If the parameter is missing from the map, it will not update the model at all

To make a parameter required, we would use `validate_required` instead.

I think the current behavior should be ok. If the user makes no changes to their twitter or website, nothing will happen to it on update. If they explicitly clear the textbox, it will get updated to a blank string. I think it should be ok for the user to allow unsetting their profile info.

I can write a validation, but I don't think it's necessary, since this is documented, default behavior of the framework.

joshsmith, line 47: Got it, thanks!

begedin:
Added some tests to user mode to test the new changeset and added the URL prefixing behavior as well. Should be good for another review.

web/models/user.ex
```diff
@@ -40,4 +54,17 @@ defmodule CodeCorps.User do
...
+    |> update_change(key, &do_prefix_url/1)
+  end
+
+  defp do_prefix_url(nil), do: nil
+
+  defp do_prefix_url("http://" <> rest), do: "http://" <> rest
+
+  defp do_prefix_url("https://" <> rest), do: "https://" <> rest
+
+  defp do_prefix_url(value), do: "http://" <> value
```
joshsmith, line 65: Related functions should be grouped together without blank lines: https://github.com/rrrene/elixir-style-guide

joshsmith, line 65: That would mean even the first `do_prefix_url` comes right after `prefix_url`.

web/models/user.ex
```diff
@@ -32,6 +37,15 @@ defmodule CodeCorps.User do
     |> put_pass_hash()
   end
 
+  def update_changeset(struct, params) do
+    struct
+    |> changeset(params)
+    |> cast(params, [:first_name, :last_name, :twitter, :biography, :website])
+    |> validate_format(:website, ~r/\A((http|https):\/\/)?[a-z0-9]+([\-\.]{1}[a-z0-9]+)*\.[a-z]{2,}(([0-9]{1,5})?\/.*)?#=\z/ix)
+    |> validate_format(:twitter, ~r/\A[a-zA-Z0-9_]{1,15}\z/)
+    |> prefix_url(:website)
```
joshsmith: Do we want `prefix_url` to happen before the validation?

begedin: I don't think it really matters. All functions will run either way. Validation doesn't break the process, it just adds to the `errors` `Keyword` in the changeset. Preference, really.

joshsmith:
Really minor stylistic note here and one question, good to merge after change without re-review.

PR #50, opened 2016-08-05
Add posts

### LINKED ISSUE #31: Add Posts API
Part of #26.

- [x] List all
- [x] Create
- [x] Retrieve
- [x] Update

snewcomer:
Closes #31 when finished.

This is unfinished, see comments below.

Blocked by organizations/projects.

### DIFF
--- a/priv/repo/migrations/20160805210821_create_post.exs
+++ b/priv/repo/migrations/20160805210821_create_post.exs
@@ -1,0 +1,24 @@
+defmodule CodeCorps.Repo.Migrations.CreatePost do
+  use Ecto.Migration
+
+  def change do
+    create table(:posts) do
+      add :number, :integer
+      add :title, :string
+      add :post_type, :string
+      add :state, :string
+      add :status, :string
+      add :body, :string
+      add :markdown, :string
+      add :likes_count, :integer
+      add :comments_count, :integer
+      add :user_id, references(:users, on_delete: :nothing)
+
+      timestamps()
+    end
+    create index(:posts, [:user_id])
+
+    create index(:posts, [:number], unique: true)
+
+  end
+end
--- a/test/models/post_test.exs
+++ b/test/models/post_test.exs
@@ -1,0 +1,47 @@
+defmodule CodeCorps.PostTest do
+  use CodeCorps.ModelCase
+
+  alias CodeCorps.Post
+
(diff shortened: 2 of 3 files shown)


joshsmith:
A few things worth pointing out:
- `number` should be unique to a Post and a Project, not unique across the whole system. The `number` will need to be incremented and avoid the problem of a race condition.
- we can leave out likes and comments counts for right now
- status should be an enum of some sort. Not sure how to do in Elixir.
- state is a state machine. Again not sure how to implement yet.
- markdown is used to render the body to HTML
- post depends on project, so this PR may be blocked until we implement that, which itself depends on organizations. We can still prep as much as we can
- the markdown to HTML rendering pipeline in Rails renders the body with both syntax highlighting and github flavored markdown

Thanks for kicking this off. Let me know if you have questions about the above.

joshsmith:
For the sequencing of posts, we used `sequenced` in Rails, which should be fairly trivial to port over to Elixir/Phoenix. Here's the [relevant code](https://github.com/djreimer/sequenced/blob/master/lib/sequenced/generator.rb) that generates the ids.

joshsmith:
I mentioned to @begedin in #62 that we can probably use [`earmark`](https://github.com/pragdave/earmark) to accomplish the markdown and github-flavored markdown portions of this.

Syntax highlighting I have yet to figure out, but we may need to move over to the client, unfortunately.

joshsmith:
We could replace `aasm` transitions from Rails with changeset pattern matching, something like:

`def transition_changeset(%Model{state: "from_this_state"} = model, "to_this_state")`

And then make a catchall that logs, complains, or just simply ignores transitions that we don't want.

Probably want to pass in a changeset and return that so we can pipe with other changesets in the model.

joshsmith:
I believe that `earmark` is already implemented elsewhere so we can take what we learned there for the markdown portion.

joshsmith:
Looks like for doing the `number` generation we can add a unique index on the `[:repo_id, :issue_number]` columns to ensure that no two issues within one repo can have the same number. Then we can use [GenRetry](https://hexdocs.pm/gen_retry/GenRetry.html) to retry any failure to insert because of a duplicate index.

joshsmith:
Closing in favor of #77. Will move relevant conversation from here over to there.

PR #54, opened 2016-08-07
Add JWT

joshsmith:
This is a first spike of JWT using Guardian.
- [x] Verify this works with `ember-simple-auth`
- [x] Verify the `secret_key` approach we're using makes sense

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -32,12 +32,10 @@ config :plug, :mimes, %{
 }
 
 config :guardian, Guardian,
-  allowed_algos: ["HS512"], # optional
-  verify_module: Guardian.JWT,  # optional
   issuer: "CodeCorps",
   ttl: { 30, :days },
   verify_issuer: true, # optional
-  secret_key: "DELETE_ME",
+  secret_key: System.get_env("GUARDIAN_SECRET_KEY"),
   serializer: CodeCorps.GuardianSerializer
 
 # Import environment specific config. This must remain at the bottom
--- a/test/controllers/auth_controller_test.exs
+++ b/test/controllers/auth_controller_test.exs
@@ -1,0 +1,55 @@
+defmodule CodeCorps.AuthControllerTest do
+  use CodeCorps.ConnCase
+
+  setup do
+    conn =
+      %{build_conn | host: "api."}
+      |> put_req_header("accept", "application/vnd.api+json")
+      |> put_req_header("content-type", "application/vnd.api+json")
+
+    {:ok, conn: conn}
+  end
+
+  defp build_payload(email, password) do
+    %{
+      "username" => email,
+      "password" => password
+    }
+  end
+
+  test "authenticates and returns JWT and user ID when data is valid", %{conn: conn} do
+    user = insert_user()
+    conn = post conn, auth_path(conn, :create), build_payload(user.email, user.password)
+
+    response = json_response(conn, 201)
+    assert response["token"]
+    assert response["user_id"] == user.id
+  end
+
+  test "does not authenticate and renders errors when the password is wrong", %{conn: conn} do
+    user = insert_user()
+    conn = post conn, auth_path(conn, :create), build_payload(user.email, "wrong password")
+
+    response = json_response(conn, 401)
+    [error | _] = response["errors"]
+    assert error["detail"] == "Your password doesn't match the email #{user.email}."
+    assert error["id"] == "UNAUTHORIZED"
+    assert error["title"] == "401 Unauthorized"
+    assert error["status"] == 401
+    refute response["token"]
+    refute response["user_id"]
+  end
+
+  test "does not authenticate and renders errors when the user doesn't exist", %{conn: conn} do
+    conn = post conn, auth_path(conn, :create), build_payload("notauser@test.com", "password")
+
+    response = json_response(conn, 401)
+    [error | _] = response["errors"]
+    assert error["detail"] == "We couldn't find a user with the email notauser@test.com."
+    assert error["id"] == "UNAUTHORIZED"
+    assert error["title"] == "401 Unauthorized"
+    assert error["status"] == 401
+    refute response["token"]
+    refute response["user_id"]
+  end
+end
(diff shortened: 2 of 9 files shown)


web/controllers/auth_controller.ex
```diff
@@ -0,0 +1,46 @@
...
+    user = Repo.get_by(User, email: email)
+    cond do
+      user && checkpw(password, user.encrypted_password) ->
+        {:ok, user}
+      user ->
+        {:error, :unauthorized}
+      true ->
+        dummy_checkpw()
+        {:error, :not_found}
+    end
```
begedin, line 44: This could maybe be reworked without a condition. 

``` Elixir
defp login_by_email_and_pass(conn, %{"email" => email, "password" => password}) do
  User
  |> Repo.get_by(email: email)
  |> do_login_by_email_and_pass(password)
end

defp do_login_by_email_and_pass(nil, _) do
  dummy_checkpw()
  {:error, :not_found}
end

defp do_login_by_email_and_pass(user, password) do
  checkpw(password, user.encrypted_password)
 |> process_login_attempt(user)
end

defp process_login_attempt(true, user), do: {:ok, user}
defp process_login_attempt(false, _), do: {:error, :unauthorized}
```

I know the code is larger, but it should be more elixir-like.

joshsmith:
This is good for someone to review now. We can implement `canary` separately.

test/controllers/auth_controller_test.exs
```diff
@@ -0,0 +1,58 @@
+defmodule CodeCorps.AuthControllerTest do
+  use CodeCorps.ConnCase
+
+  @valid_attrs %{description: "You want to improve software tools and infrastructure.", name: "Technology"}
+  @invalid_attrs %{name: nil}
```
begedin: These two seem to be redundant here

web/auth/guardian_serializer.ex
```diff
@@ -0,0 +1,12 @@
+defmodule CodeCorps.GuardianSerializer do
+  @behaviour Guardian.Serializer
+
+  alias CodeCorps.Repo
+  alias CodeCorps.User
+
+  def for_token(user = %User{}), do: { :ok, "User:#{user.id}" }
+  def for_token(_), do: { :error, "Unknown resource type" }
+
+  def from_token("User:" <> id), do: { :ok, Repo.get(User, id) }
```
begedin, line 10: Nice one!

begedin:
Good to merge after the cleanup of that redundant code. 👍

joshsmith:
@begedin by redundant do you mean unused?

begedin:
@JoshSmith Yeah, sorry, no idea why I used that term, twice. 

I guess you could call it...

...redundant

joshsmith:
Dad joke.

web/auth/guardian_serializer.ex
green-arrow, line 10: Just a quick comment here: `from_token/1` is invoked overtime you deserialize a JWT coming into the API, which means we're automatically incurring a round trip DB hit at the start of every call. Do we need the whole user object?

web/views/auth_view.ex
```diff
@@ -0,0 +1,27 @@
+defmodule CodeCorps.AuthView do
+  use CodeCorps.Web, :view
+
+  def render("show.json", %{token: token, user_id: user_id}) do
+    %{
+      token: token,
+      user_id: user_id,
```
green-arrow, line 7: Is `user_id` specifically needed separately from the client? If not, we should be able to just decode the JWT on the client and retrieve the users details from that.

config/config.exs
```diff
@@ -32,12 +32,10 @@ config :plug, :mimes, %{
 }
 
 config :guardian, Guardian,
-  allowed_algos: ["HS512"], # optional
-  verify_module: Guardian.JWT,  # optional
   issuer: "CodeCorps",
   ttl: { 30, :days },
   verify_issuer: true, # optional
-  secret_key: "DELETE_ME",
+  secret_key: System.get_env("GUARDIAN_SECRET_KEY"),
```
green-arrow, line 38: We may want to look more into this. The problem that we're going to run into is that when you build a release, this config gets compiled down into a `sys.config` file, which cannot read from `System.get_env/1`. This is true of both `distillery` and `exrm` releases. The recommended way to do runtime configuration in `distillery` is here:

https://github.com/bitwalker/distillery/blob/master/docs/Runtime%20Configuration.md

The problem with this is that we're not actually consuming this value ourselves, it's being consumed by Guardian. However, they do have an example using an anonymous function. So maybe we could do something like this:

``` elixir

config :guardian, Guardian,
  secret_key: fn ->
    System.get_env("GUARDIAN_SECRET_KEY")
  end
```

We can probably leave this as-is for now, but will most likely need to revisit it when we get to the deploy step.

web/views/auth_view.ex
joshsmith, line 7: Right now it is needed, but we can refactor the Ember side to pull it from the JWT if we'd like.

config/config.exs
joshsmith, line 38: Okay, maybe we should an issue so we remember that when we try to deploy next week?

web/auth/guardian_serializer.ex
joshsmith, line 10: I think we do need this, don't we? Regardless of whether we need all the user's information, we have to hit the database to see if that user exists.

This is from the README of Guardian itself.

Is there something I'm missing here?

green-arrow, line 10: I'm not sure. If you can decode the JWT, then you were able to successfully authenticate the user and generate the JWT in the first place. Unless we're worried that a user may get deleted from the system after they've authenticated to the system, I don't know if it's necessary.

FWIW, I think the Guardian docs are just kind of showing what's possible with the `from_token/1` method.

config/config.exs
joshsmith, line 38: @green-arrow did you want to create that issue?

web/auth/guardian_serializer.ex
joshsmith, line 10: @green-arrow wouldn't we need the user's information still though for something like `canary`. Sure, we could encode role and access information into the token, but it's entirely possible the token does not have the same level of information about the user as our database. For most authenticated requests, won't we want that level of information about the user?

PR #56, opened 2016-08-09
Initial Skills API implementation

### LINKED ISSUE #37: Add Skills API
Part of #26.

- [ ] Create
- [ ] Filter
- [ ] Autocomplete
- [ ] Retrieve

npendery:
Closes #37

### DIFF
--- a/priv/repo/migrations/20160805203929_create_skill.exs
+++ b/priv/repo/migrations/20160805203929_create_skill.exs
@@ -1,0 +1,16 @@
+defmodule CodeCorps.Repo.Migrations.CreateSkill do
+  use Ecto.Migration
+
+  def change do
+    create table(:skills) do
+      add :title, :string
+      add :description, :string
+      add :original_row, :integer
+      add :slug, :string
+
+      timestamps()
+    end
+
+    create index(:skills, [:slug], name: :index_skills_on_slug, unique: true)
+  end
+end
--- a/priv/repo/migrations/20160805203929_create_skill.exs
+++ b/priv/repo/migrations/20160805203929_create_skill.exs
@@ -3,14 +3,14 @@ defmodule CodeCorps.Repo.Migrations.CreateSkill do
 
   def change do
     create table(:skills) do
-      add :title, :string
+      add :title, :string, null: false
       add :description, :string
       add :original_row, :integer
-      add :slug, :string
+      add :slug, :string, null: false
 
       timestamps()
     end
 
-    create index(:skills, [:slug], name: :index_skills_on_slug, unique: true)
+    create index(:skills, ["lower(username)"], name: :index_skills_on_slug, unique: true)
   end
 end
--- a/priv/repo/migrations/20160805203929_create_skill.exs
+++ b/priv/repo/migrations/20160805203929_create_skill.exs
@@ -6,11 +6,10 @@ defmodule CodeCorps.Repo.Migrations.CreateSkill do
       add :title, :string, null: false
       add :description, :string
       add :original_row, :integer
-      add :slug, :string, null: false
 
       timestamps()
     end
 
-    create index(:skills, ["lower(username)"], name: :index_skills_on_slug, unique: true)
+    create index(:skills, ["lower(title)"], name: :index_skills_on_title, unique: true)
   end
 end
--- a/test/controllers/skill_controller_test.exs
+++ b/test/controllers/skill_controller_test.exs
@@ -7,8 +7,7 @@ defmodule CodeCorps.SkillControllerTest do
   @valid_attrs %{
     description: "Elixir is a functional, concurrent, general-purpose programming language that runs on the Erlang virtual machine (BEAM).",
     original_row: 1,
-    slug: "elixir-in-phoenix",
-    title: "Elixir in Phoenix"
+    title: "Elixir"
   }
   @invalid_attrs %{}
 
@@ -39,7 +38,6 @@ defmodule CodeCorps.SkillControllerTest do
     assert data["attributes"]["title"] == skill.title
     assert data["attributes"]["description"] == skill.description
     assert data["attributes"]["original_row"] == skill.original_row
-    assert data["attributes"]["slug"] == skill.slug
   end
 
   test "does not show resource and instead throw error when id is nonexistent", %{conn: conn} do
--- a/test/controllers/skill_controller_test.exs
+++ b/test/controllers/skill_controller_test.exs
@@ -4,7 +4,7 @@ defmodule CodeCorps.SkillControllerTest do
   alias CodeCorps.Skill
   alias CodeCorps.Repo
 
-  @valid_attrs %{description: "some content", original_row: 42, slug: "some content", title: "some content"}
+  @valid_attrs %{description: "Knowledge of the Elixir Language", original_row: 1, slug: "elixir-in-phoenix", title: "Elixir in Phoenix"}
   @invalid_attrs %{}
 
   setup do
--- a/test/controllers/skill_controller_test.exs
+++ b/test/controllers/skill_controller_test.exs
@@ -1,0 +1,72 @@
+defmodule CodeCorps.SkillControllerTest do
+  use CodeCorps.ConnCase
+
+  alias CodeCorps.Skill
+  alias CodeCorps.Repo
+
+  @valid_attrs %{description: "some content", original_row: 42, slug: "some content", title: "some content"}
+  @invalid_attrs %{}
+
+  setup do
+    conn =
+      %{build_conn | host: "api."}
+      |> put_req_header("accept", "application/vnd.api+json")
+      |> put_req_header("content-type", "application/vnd.api+json")
+
+    {:ok, conn: conn}
+  end
+
+  defp relationships do
+    %{}
+  end
+
+  test "lists all entries on index", %{conn: conn} do
+    conn = get conn, skill_path(conn, :index)
+    assert json_response(conn, 200)["data"] == []
+  end
+
+  test "shows chosen resource", %{conn: conn} do
+    skill = Repo.insert! %Skill{}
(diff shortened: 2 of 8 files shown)


web/models/skill.ex
```diff
@@ -0,0 +1,55 @@
...
+      %Ecto.Changeset{changes: %{title: title}} ->
+        slug = Inflex.parameterize(title)
+        put_change(changeset, :slug, slug)
+      _ ->
+        changeset
+    end
+
+  end
+
+  def reserved_routes do
```
npendery: Not sure where we want this stored...but pretty sure it should be in some file with constants

priv/repo/migrations/20160805203929_create_skill.exs
```diff
@@ -0,0 +1,16 @@
+defmodule CodeCorps.Repo.Migrations.CreateSkill do
+  use Ecto.Migration
+
+  def change do
+    create table(:skills) do
+      add :title, :string
+      add :description, :string
+      add :original_row, :integer
+      add :slug, :string
```
joshsmith: I think many of these probably need to be `NOT NULL`.

priv/repo/migrations/20160805203929_create_skill.exs
```diff
@@ -0,0 +1,16 @@
...
+
+      timestamps()
+    end
+
+    create index(:skills, [:slug], name: :index_skills_on_slug, unique: true)
```
joshsmith: Is this case-insensitive? It should be, but doesn't look like it is. Does it need `lower`?

test/controllers/skill_controller_test.exs
```diff
@@ -0,0 +1,72 @@
+defmodule CodeCorps.SkillControllerTest do
+  use CodeCorps.ConnCase
+
+  alias CodeCorps.Skill
+  alias CodeCorps.Repo
+
+  @valid_attrs %{description: "Knowledge of the Elixir Language", original_row: 1, slug: "elixir-in-phoenix", title: "Elixir in Phoenix"}
```
joshsmith: I would simplify to our API Blueprint's language for consistency.

web/controllers/skill_controller.ex
```diff
@@ -0,0 +1,45 @@
...
+  def index(conn, params) do
+    skills =
+      case params do
+        %{"filter" => filter} ->
+          ids = String.split(filter, ",") |> Enum.map(fn(n) -> String.to_integer(n) end)
+          query = Skill |> where([p], p.id in ^ids)
+          Repo.all(query)
+        %{"query" => query} ->
+          text_query = Skill |> where([p], ilike(p.title, ^"#{query}%"))
+          Repo.all(text_query)
```
joshsmith, line 18: I like the use of `ilike` here but think you're maybe doing too much with it. We should consider waiting until we reintroduce Searchkick. What do you think?

web/controllers/skill_controller.ex
```diff
@@ -0,0 +1,45 @@
...
+  alias CodeCorps.Skill
+  alias JaSerializer.Params
+
+  plug :scrub_params, "data" when action in [:create]
+
+  def index(conn, params) do
+    skills =
+      case params do
+        %{"filter" => filter} ->
+          ids = String.split(filter, ",") |> Enum.map(fn(n) -> String.to_integer(n) end)
```
joshsmith: I wonder if this can be simplified some.

`%{"id" => [1,2]}`

web/models/skill.ex
```diff
@@ -0,0 +1,55 @@
...
+  def update_slug(changeset) do
+    case changeset do
+      %Ecto.Changeset{changes: %{title: title}} ->
+        slug = Inflex.parameterize(title)
+        put_change(changeset, :slug, slug)
+      _ ->
+        changeset
+    end
+
+  end
```
joshsmith: We should see what @green-arrow is going to do here.

web/models/skill.ex
joshsmith: Why don't you just `validate_slug` from our `SlugValidator` module instead? Wouldn't that get you what you want?

joshsmith: I guess there might be an issue with its usage, for example I'm not sure "C++" would work. I'm curious what those slugs look like right now.

web/views/skill_view.ex
```diff
@@ -0,0 +1,6 @@
+defmodule CodeCorps.SkillView do
+  use CodeCorps.Web, :view
+  use JaSerializer.PhoenixView
+
+  attributes [:title, :description, :original_row, :slug, :inserted_at, :updated_at]
```
joshsmith: Is there any reason to serialize the `original_row` here?

joshsmith:
It looks like we're not even serializing `slug` in the Rails app for this model. Maybe to simplify we could punt on the slug? What do you think?

web/controllers/skill_controller.ex
joshsmith, line 18: Can you add an issue for the `elasticsearch` requirement, explaining how it works currently and how it needs to work? Or I could fill in details if needed.

web/models/skill.ex
npendery: C++ would just be `c++`
And yeah ill just use the `SlugValidator`

web/controllers/skill_controller.ex
npendery, line 18: Yeah, was looking into it briefly but definitely need more time to learn how we would implement `Tirexs` into this search. Used to working with a higher level client like `Chewy`

joshsmith, line 18: Maybe we should keep it until we get `elasticsearch` figured out. I also am not so great with it. I ultimately used `searchkick` in Rails because I couldn't get precisely what I wanted mapped well.

Maybe we can find someone with deeper ElasticSearch skills.

priv/repo/migrations/20160805203929_create_skill.exs
npendery: I saw that you had this for users...would you mind explaining the usage for this?

joshsmith: The idea is to differentiate between `code-corps` and `CODE-CORPS` which should resolve to the same thing. I don't believe uniqueness will deal with these constraints.

web/controllers/skill_controller.ex
npendery, line 18: Yeah sounds good. Since this is such a simple indexing it probably wont be hard, but also dont know if its necessary. 
The current query should work well though since responses are now in µs just by being in Elixir? Just need to have an observer on the Ember side text field and it will send a request with every update and return an index with the relevant skills.

web/controllers/skill_controller.ex
npendery: Could add a method to do these steps to break it out for clarity, but not sure how else we are going to convert a string of csv's to a list of numbers...

web/views/skill_view.ex
npendery: ¯ \ _ (ツ)_/¯
What is it used for?

web/controllers/skill_controller.ex
joshsmith: I would maybe ask something on Stack Overflow to see if someone else has a consistently good way of doing this. This seems like a wheel that probably _shouldn't_ need reinvented given that there are plenty of people using Phoenix with coalesced ids from Ember.

web/views/skill_view.ex
joshsmith: In the Rails app for tracking imports from the CSV of skills:

`Skill.find_or_create_by(original_row: line["Original Row"])`

priv/repo/migrations/20160805203929_create_skill.exs
npendery: There should never be an issue with that because of the Inflex dep `parameterize` method that creates the slug will always be lowercase

joshsmith: I still believe the correct thing to do is ensure this at the database level for consistency. What if we remove Phoenix? Remove `Inflex`? Remove `parameterize`?

I wouldn't rely on application-level logic for your data when your RDBMS can provide that additional sanity check.

web/models/skill.ex
npendery: What is he doing?

joshsmith: That PR is merged now, but check the `Organization` model in `develop`.

test/controllers/skill_controller_test.exs
```diff
@@ -0,0 +1,75 @@
...
+  end
+
+  defp relationships do
+    %{}
+  end
+
+  test "lists all entries on index", %{conn: conn} do
+    conn = get conn, skill_path(conn, :index)
+    assert json_response(conn, 200)["data"] == []
+  end
```
joshsmith, line 30: Should we also be testing the `filter` and `query` cases here?

web/controllers/skill_controller.ex
```diff
@@ -0,0 +1,49 @@
...
+
+  def index(conn, params) do
+    skills =
+      case params do
+        %{"filter" => filter} ->
+          ids = coalesce_id_string(filter)
+          query = Skill |> where([p], p.id in ^ids)
+          Repo.all(query)
+        %{"query" => query} ->
+          text_query = Skill |> where([p], ilike(p.title, ^"%#{query}%"))
```
joshsmith, line 17: I'm guessing this doesn't suffer from any SQL injection vulnerabilities?

joshsmith:
Looks like you have a failing test, and a couple tests cases we should add into the controller.

Also a quick question on this one.

test/controllers/skill_controller_test.exs
joshsmith, line 30: This is now tested.

joshsmith:
Now awaiting review.

Issue #27, opened 2016-08-01
Add Categories API


joshsmith:
Part of #26.

This can be split out into multiple issues to reflect the endpoints in the API docs: http://docs.codecorpsapidevelop.apiary.io/#reference/categories
- [x] Create
- [x] List all
- [x] Retrieve
- [x] Update

joshsmith:
Closed by #58.

PR #58, opened 2016-08-10
Adding index, create, show, update endpoints for categories

green-arrow:
Addresses #27

### DIFF
--- a/priv/repo/migrations/20160808143454_create_category.exs
+++ b/priv/repo/migrations/20160808143454_create_category.exs
@@ -10,6 +10,6 @@ defmodule CodeCorps.Repo.Migrations.CreateCategory do
       timestamps()
     end
 
-    create index(:categories, [:slug], unique: true, name: :index_categories_on_slug)
+    create index(:categories, ["lower(slug)"], unique: true, name: :index_categories_on_slug)
   end
 end
--- a/priv/repo/migrations/20160808143454_create_category.exs
+++ b/priv/repo/migrations/20160808143454_create_category.exs
@@ -1,0 +1,15 @@
+defmodule CodeCorps.Repo.Migrations.CreateCategory do
+  use Ecto.Migration
+
+  def change do
+    create table(:categories) do
+      add :name, :string, null: false
+      add :slug, :string, null: false
+      add :description, :string
+
+      timestamps()
+    end
+
+    create index(:categories, [:slug], unique: true, name: :index_categories_on_slug)
+  end
+end
--- a/test/controllers/category_controller_test.exs
+++ b/test/controllers/category_controller_test.exs
@@ -3,7 +3,7 @@ defmodule CodeCorps.CategoryControllerTest do
 
   alias CodeCorps.Category
 
-  @valid_attrs %{description: "The technology category", name: "Technology", slug: "technology"}
+  @valid_attrs %{description: "You want to improve software tools and infrastructure.", name: "Technology"}
   @invalid_attrs %{name: nil}
 
   setup do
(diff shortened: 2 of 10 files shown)


priv/repo/migrations/20160808143454_create_category.exs
```diff
@@ -0,0 +1,15 @@
...
+  def change do
+    create table(:categories) do
+      add :name, :string, null: false
+      add :slug, :string, null: false
+      add :description, :string
+
+      timestamps()
+    end
+
+    create index(:categories, [:slug], unique: true, name: :index_categories_on_slug)
```
joshsmith: Is this case-insensitive? I think you may need a `lower` here.

test/controllers/category_controller_test.exs
```diff
@@ -0,0 +1,86 @@
+defmodule CodeCorps.CategoryControllerTest do
+  use CodeCorps.ConnCase
+
+  alias CodeCorps.Category
+
+  @valid_attrs %{description: "The technology category", name: "Technology", slug: "technology"}
```
joshsmith: Would reword as:

> You want to improve software tools and infrastructure.

For `description`.

test/models/category_test.exs
```diff
@@ -0,0 +1,18 @@
+defmodule CodeCorps.CategoryTest do
+  use CodeCorps.ModelCase
+
+  alias CodeCorps.Category
+
+  @valid_attrs %{description: "The technology category", name: "Technology", slug: "technology"}
```
joshsmith: Would also rename `description` here as above.

joshsmith:
Looks great!

One thing we should probably change is not allowing `slug` to be set directly but instead having `slug` set on creation by using `Inflex.parameterize/1`, as in the private `generate_slug/1` in the `Organization` model.

priv/repo/migrations/20160808143454_create_category.exs
green-arrow: I didn't notice that in the schema / migrations in the ruby project, so I didn't add it here. Does ruby do that by default?

Also, I believe that if we create the index via `lower`, we'll have to make sure that whenever we query for the slug we always lower case it in the query. If that's correct I'll change it here :)

joshsmith:
Another thing of note (not handled in previous endpoints) is that the create/update endpoints should only be allowed for admin users. There's no concept right now of an admin user.

green-arrow:
@JoshSmith how many times are generated slugs used in the database? If it's more than just here and in the organizations, I'm going to abstract out that `generate_slug/1` method to a model helper module.

joshsmith:
Similarly, it might be nice to bring over the seeds we had in Rails. Can we use such seeds in production, too?

``` ruby
categories = [
  {
    name: "Arts",
    description: "You want to improve the arts."
  },
  {
    name: "Economy",
    description: "You want to improve finance and the economic climate."
  },
  {
    name: "Education",
    description: "You want to improve literacy, schools, and training."
  },
  {
    name: "Environment",
    description: "You want to improve your environment."
  },
  {
    name: "Government",
    description: "You want to improve government responsiveness."
  },
  {
    name: "Health",
    description: "You want to improve prevention and treatment."
  },
  {
    name: "Justice",
    description: "You want to improve your judicial system."
  },
  {
    name: "Politics",
    description: "You want to improve elections and voting."
  },
  {
    name: "Public Safety",
    description: "You want to improve crime prevention and safety."
  },
  {
    name: "Science",
    description: "You want to improve tools for advancing science."
  },
  {
    name: "Security",
    description: "You want to improve tools like encryption."
  },
  {
    name: "Society",
    description: "You want to improve our communities."
  },
  {
    name: "Technology",
    description: "You want to improve software tools and infrastructure."
  },
  {
    name: "Transportation",
    description: "You want to improve how people travel."
  },
]
```

joshsmith:
@green-arrow go ahead and abstract it out. I'm not positive on where else it's used, but you can assume (for SEO purposes) that it may be reused often.

green-arrow:
@JoshSmith I think I addressed all comments in the last commit. Let me know if you see anything else that needs changed :)

joshsmith:
This looks good to me. Merge when ready. 👍

PR #62, opened 2016-08-10
Add projects API initial commit

begedin:
This should now have everything except for conversion of base64 into an image and upload to AWS


priv/repo/migrations/20160810124357_create_project.exs
```diff
@@ -0,0 +1,19 @@
+defmodule CodeCorps.Repo.Migrations.CreateProject do
+  use Ecto.Migration
+
+  def change do
+  	create table(:projects) do
```
joshsmith: I'm not sure why your first tab is so off. Do we have `.editorconfig` here?

priv/repo/migrations/20160810124357_create_project.exs
```diff
@@ -0,0 +1,19 @@
...
+      add :description, :string
+      add :icon_large_url, :string
+      add :icon_thumb_url, :string
+      add :long_description_body, :string
+      add :long_description_markdown, :string
+      add :slug, :string
+      add :title, :string
```
joshsmith: Any of these need to be `NOT NULL`? I think at least `title` and `slug`, yes?

joshsmith: Does `slug` need to have an index?

test/controllers/project_controller_test.exs
```diff
@@ -0,0 +1,58 @@
...
+    data = json_response(conn, 200)["data"]
+    assert data["id"] == "#{project.id}"
+    assert data["type"] == "project"
+    assert data["attributes"]["title"] == "Test project"
+    assert data["attributes"]["description"] == "Test project description"
+    assert data["attributes"]["long-description-markdown"] == "A markdown **description**"
+    assert data["relationships"]["organizations"]["id"] == organization.id
+  end
+
+  test "shows chosen resource retrieved by slug"
```
joshsmith: Do we want to retrieve this by `slug`? Is it needed? I think I missed your points on this re: the Ember app before, but not positive now. Do we still need it?

I would like to avoid the same API structure we currently have if possible. The routing was really weird and I think not necessary in a number of cases.

web/models/project.ex
```diff
@@ -0,0 +1,68 @@
...
+    field :title, :string
+
+    belongs_to :organization, CodeCorps.Organization
+
+    timestamps()
+  end
+
+  def changeset(struct, params) do
+    struct
+  	|> cast(params, [:title, :description, :long_description_markdown])
```
joshsmith: Weird spacing issue here, too.

web/models/project.ex
```diff
@@ -0,0 +1,68 @@
...
+    |> validate_required(:title)
+    |> generate_slug()
+    |> validate_slug(:slug)
+    |> add_project_icons(Map.get(params, :base64_icon_data))
```
joshsmith, line 32: Why this and not just use the `:base64_icon_data` casted?

web/models/project.ex
```diff
@@ -0,0 +1,68 @@
...
+  defp generate_slug(changeset) do
+    slug =
+      get_change(changeset, :title)
+      |> do_generate_slug
+
+    changeset
+    |> put_change(:slug, slug)
+  end
+  defp do_generate_slug(nil), do: nil
+  defp do_generate_slug(value), do: Inflex.parameterize(value)
```
joshsmith: This seems different than the other approaches?

web/models/project.ex
```diff
@@ -0,0 +1,68 @@
...
+
+  defp render_markdown_to_html(changeset) do
+    html =
+      changeset
+      |> get_change(:long_description_markdown)
+      |> do_render_markdown_to_html
+
+    changeset
+    |> put_change(:long_description_body, html)
+  end
```
joshsmith, line 60: Is there a reason we're using `get_change` here instead of passing in the field specifically?

web/views/project_view.ex
```diff
@@ -0,0 +1,11 @@
+defmodule CodeCorps.ProjectView do
+  use CodeCorps.Web, :view
+  use JaSerializer.PhoenixView
+
+  attributes [
+  	:slug, :title, :description, :icon_thumb_url, :icon_large_url,
+  	:long_description_body, :long_description_markdown,
+  	:inserted_at, :updated_at]
```
joshsmith, line 8: The line breaking here seems a little odd.

joshsmith:
@begedin on the markdown generation side was thinking we could use something like [earmark](https://github.com/pragdave/earmark)

joshsmith:
@begedin think some of my comments from earlier were missed. Would you mind looking again?

priv/repo/migrations/20160810124357_create_project.exs
begedin: We don't. Created #65

priv/repo/migrations/20160810124357_create_project.exs
begedin: Agreed on both parts. I don't think anything else needs to be non-nullable, though.

web/models/project.ex
begedin: Didn't realise we now have the function in a separate module. Switched to using that now.

web/models/project.ex
begedin, line 60: `get_change` will return a `nil` if the field has a value but hasn't actually been changed. In return, `do_render_markdown_to_html` will then simply return `nil`. After this, the code did have a bug, so I resolved that and added specs to cover it.

From now on, if a changeset actually has a markdown change, then the body will also rerender. If not, nothing will happen.

web/views/project_view.ex
begedin, line 8: What do you mean? Indentation?

If it's grouping, if I know there will be at least 3 lines, I prefer to group them in a more meaningful way, instead of just breaking at fixed points. In this case, common fields are on the same line.
- `:slug`, `:title`, and `:description` are on the same line
- `:icon_thumb_url`, `:icon_large_url` are on the same line

etc.

joshsmith:
👍

PR #63, opened 2016-08-10
Add CORS

joshsmith:
This adds CORS using `corsica` to be sure we can actually make requests to the server with our Ember application.

Testing inside the PageController test.

### DIFF
--- a/lib/code_corps/endpoint.ex
+++ b/lib/code_corps/endpoint.ex
@@ -38,6 +38,6 @@ defmodule CodeCorps.Endpoint do
     key: "_code_corps_key",
     signing_salt: "RJebPcfU"
 
-  plug Corsica, origins: "http://localhost:4200"
+  plug Corsica, origins: Application.get_env(:code_corps, :allowed_origins)
   plug CodeCorps.Router
 end
--- a/lib/code_corps/endpoint.ex
+++ b/lib/code_corps/endpoint.ex
@@ -38,5 +38,6 @@ defmodule CodeCorps.Endpoint do
     key: "_code_corps_key",
     signing_salt: "RJebPcfU"
 
+  plug Corsica, origins: "http://localhost:4200"
   plug CodeCorps.Router
 end
--- a/test/controllers/page_controller_test.exs
+++ b/test/controllers/page_controller_test.exs
@@ -5,4 +5,20 @@ defmodule CodeCorps.PageControllerTest do
     conn = get conn, "/"
     assert html_response(conn, 200) =~ "Welcome to Phoenix!"
   end
+
+  test "allows CORS from http://localhost:4200" do
+    conn = build_conn()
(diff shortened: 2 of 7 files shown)


joshsmith:
If anyone has a good idea for how to write a test for this, feel free to run with it. This is blocked until that time.

_EDIT_: now testing inside the PageController test.

joshsmith:
This could use a review from anyone now.

lib/code_corps/endpoint.ex
```diff
@@ -38,5 +38,6 @@ defmodule CodeCorps.Endpoint do
     key: "_code_corps_key",
     signing_salt: "RJebPcfU"
 
+  plug Corsica, origins: "http://localhost:4200"
```
noslouch: The allowed origins should probably be defined in the env config files and hen accessed here. I think something like:

``` elixir
config :code_corps, allowed_origins: {"http://localhost:4200"} 
```

in config/dev.ex and then accessed here like

``` elixir
Application.get_env :code_corps, :allowed_origins
```

would do the trick

test/controllers/page_controller_test.exs
```diff
@@ -5,4 +5,20 @@ defmodule CodeCorps.PageControllerTest do
...
+    assert "http://localhost:4200" in get_resp_header(conn, "access-control-allow-origin")
+  end
+
+  test "does not allow CORS for http://google.com" do
+    conn = build_conn()
+      |> put_req_header("origin", "http://google.com")
+      |> get("/")
+
+    refute "http://google.com" in get_resp_header(conn, "access-control-allow-origin")
+  end
```
noslouch, line 23: Should this also test for pre-flight requests?

joshsmith, line 23: I don't know if we need to do more than a smoke test to see if the plug is there. What do you think?

lib/code_corps/endpoint.ex
joshsmith: Good call. Any idea if we should be testing all the envs then or not?

test/controllers/page_controller_test.exs
noslouch, line 23: Fair enough. I guess this is mostly testing that we've got CORS enabled and Corsica can do the rest. 

Though I wonder if there's value is faking different envs to test the allowed origins? Might be overkill.

joshsmith, line 23: Yeah that's the way I'd lean, against the overkill. Churn is low on this anyway. Just wanted some basic testing.

lib/code_corps/endpoint.ex
noslouch: Ha see my other comment. Might be overkill to update a config and then update a test for that config. But then again when a new subdomain is added and it breaks the app we'll know who to blame.

PR #71, opened 2016-08-15
Add coalesce id support to users index endpoint

### LINKED ISSUE #42: Add Users API

begedin:
Closes #42, so I guess this is the answer to the ultimate question about the life, the universe and everything?

SImilar to how I did #70. Some changes overlap, because they were copy-paste


joshsmith:
Thanks for all the 🐟

PR #77, opened 2016-08-16
Add API for posts

joshsmith:
Not even closed to finish yet, but provides some skeleton work here.
- `number` should be unique to a Post and a Project, not unique across the whole system. The `number` will need to be incremented and avoid the problem of a race condition.
- `status` should be an enum of some sort. Maybe use something like what was used in `Skill`.
- `state` is a state machine.
- markdown is used to render the body to HTML. Reference `Preview` for the markdown generation.
- we have no server-side syntax highlighting

For the sequencing of posts, we used `sequenced` in Rails, which should be fairly trivial to port over to Elixir/Phoenix. Here's the [relevant code](https://github.com/djreimer/sequenced/blob/master/lib/sequenced/generator.rb) that generates the ids.

To generate the `number`, we could add a unique index on the `[:project_id, :number]` columns to ensure that no two issues within one repo can have the same number. Then we can use [GenRetry](https://hexdocs.pm/gen_retry/GenRetry.html) to retry any failure to insert because of a duplicate index.

We could replace `aasm` transitions from Rails with changeset pattern matching, something like: `def transition_changeset(%Model{state: "from_this_state"} = model, "to_this_state")` And then make a catchall that logs, complains, or just simply ignores transitions that we don't want. Probably want to pass in a changeset and return that so we can pipe with other changesets in the model.
- [x] Add markdown to html generation
- [ ] Ensure number can be incremented concurrently without race condition (needs test)
- [ ] Make state into an actual state machine


begedin:
Started figuring out the sequence generation. Will update if I'm done/done for the day.

begedin:
Added basic sequence generation. Concurrency isn't handled yet. Done for now, so if someone wants to take over, go ahead.

Also referencing #31 for easier navigation.

joshsmith:
I removed `state` from this because I think we should hold off on adding the state machine piece until later. It's not a huge priority.

Let's open a separate, follow-on issue for this.

joshsmith:
I also added a failing test to start hitting `projects/:id/posts`; I did add a working `index` function.

I also can't find any way to test concurrency. We should also add a follow-up issue for this.

joshsmith:
Similarly, we should add a follow-up issue for adding in pagination and `post_type` filters, which is sorely missing from this, as well.

joshsmith:
@begedin if you want to take this one over to finish up that `Not yet implemented` failing test, review my changes, and then merge if you feel ready, all good with me. This PR is getting long in the tooth.

PR #78, opened 2016-08-16
[ROLE-SKILLS]: add index, filter, create, delete actions

### LINKED ISSUE #35: Add Role Skills API
Part of #26.

- [ ] Create

joshsmith:
Closes #35.


joshsmith:
Only thing different than user skill is the type is included in the relationship data and a few extra assertions.

Should maybe port these over to user skill. @begedin?

begedin:
Not a bad idea to port. Looks good to go otherwise, so your choice if you wanna do it here or separately. 👍

joshsmith:
Separately. Can you open an issue?

begedin:
Created issue #81

PR #79, opened 2016-08-16
Add seeds and fix bug in ProjectController

joshsmith:
This adds seeds for users, organizations, and projects.

There was also a bug with the index action on projects where the organizations were not properly preloaded.

This PR now makes the Phoenix app as compatible with the Ember app as we can be, given missing `state` for user onboarding, missing images, missing posts, etc.

### DIFF
--- a/priv/repo/seeds.exs
+++ b/priv/repo/seeds.exs
@@ -1,5 +1,75 @@
 alias CodeCorps.Repo
 alias CodeCorps.Category
+alias CodeCorps.Organization
+alias CodeCorps.Project
+alias CodeCorps.User
+
+# Users
+
+users = [
+  %{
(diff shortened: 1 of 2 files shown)


joshsmith:
I'm also not super satisfied with the edge case handling of the seeds here, including creating relationships. We should improve on this some.

joshsmith:
@begedin mind giving a review?

begedin:
Looks good. I was being silly and used "update branch" from the UI here. Hope that doesn't make the squash problematic. I forgot it sometimes does. 👍

begedin:
Also, now that we have `:ex_machina`, it's also possible to use it in dev mode for seeding, so we should look into that.

PR #82, opened 2016-08-17
Add PostComment API

### LINKED ISSUE #28: Add Comments API
Part of #26.

- [ ] Create
- [ ] List all
- [ ] Filter
- [ ] Update

npendery:
Closes #28

### DIFF
--- a/priv/repo/migrations/20160817220118_create_comment.exs
+++ b/priv/repo/migrations/20160817220118_create_comment.exs
@@ -1,0 +1,17 @@
+defmodule CodeCorps.Repo.Migrations.CreateComment do
+  use Ecto.Migration
+
+  def change do
+    create table(:comments) do
+      add :body, :text, null: false
+      add :markdown, :text, null: false
+      add :user_id, references(:users, on_delete: :nothing)
+      add :post_id, references(:posts, on_delete: :nothing)
+
+      timestamps()
+    end
+
+    create index(:comments, [:user_id])
+    create index(:comments, [:post_id])
+  end
+end
--- a/test/models/comment_test.exs
+++ b/test/models/comment_test.exs
@@ -1,0 +1,32 @@
+defmodule CodeCorps.CommentTest do
+  use CodeCorps.ModelCase
+
+  alias CodeCorps.Comment
+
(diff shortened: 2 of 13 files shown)


priv/repo/migrations/20160817220118_create_comment.exs
```diff
@@ -0,0 +1,18 @@
+defmodule CodeCorps.Repo.Migrations.CreateComment do
+  use Ecto.Migration
+
+  def change do
+    create table(:comments) do
+      add :body, :text
+      add :markdown, :text
+      add :state, :string
```
joshsmith: I would make all of these `null: false`.

web/models/comment.ex
```diff
@@ -0,0 +1,33 @@
+defmodule CodeCorps.Comment do
+  use EctoStateMachine,
```
joshsmith: Looks like you left this out of `mix.exs`.

web/controllers/comment_controller.ex
```diff
@@ -0,0 +1,59 @@
...
+
+  def delete(conn, %{"id" => id}) do
+    comment = Repo.get!(Comment, id)
+
+    # Here we use delete! (with a bang) because we expect
+    # it to always work (and if it does not, it will raise).
+    Repo.delete!(comment)
+
+    send_resp(conn, :no_content, "")
+  end
```
joshsmith: Should remove the `delete` endpoint.

test/models/comment_test.exs
```diff
@@ -0,0 +1,18 @@
+defmodule CodeCorps.CommentTest do
+  use CodeCorps.ModelCase
+
+  alias CodeCorps.Comment
+
+  @valid_attrs %{body: "some content", markdown: "some content", state: "some content"}
```
joshsmith: This looks like it it should not be valid since the `state` is not in the set of states we expect.

joshsmith:
Yes, I think even after seeing I'm still partial to calling this `Comment` and not `PostComment`.

joshsmith:
I'll finish this up.

rjsamson:
Looks good to me. 👍

PR #87, opened 2016-08-18
Add basic organization memberships api

### LINKED ISSUE #30: Add Organization Memberships API
Part of #26.

- [ ] List all
- [ ] Create
- [ ] Filter
- [ ] Retrieve
- [ ] Update
- [ ] Delete

begedin:
Closes #30

There's the whole who can upgrade a user to what, thing, but that should arrive with proper authorisation. 

The API defined in apiary is covered with this PR


priv/repo/migrations/20160818132546_add_organization_membership.exs
```diff
@@ -0,0 +1,20 @@
...
+      add :organization_id, references(:organizations, on_delete: :nothing), null: false
+      add :member_id, references(:users, on_delete: :nothing), null: false
+
+      timestamps()
+    end
+
+    create index(
+      :organization_memberships,
+      [:member_id, :organization_id],
+      name: :organization_memberships_member_id_index,
```
joshsmith: Is there a reason you chose a specific name here?

test/controllers/organization_memberships_controller_test.exs
```diff
@@ -0,0 +1,179 @@
...
+  alias CodeCorps.User
+
+  import CodeCorps.Factories
+
+  def filter_params(records) do
+    ids =
+    records
+    |> Enum.map(fn(r) -> r.id end)
+    |> Enum.join(",")
+    %{filter: %{id: ids}}
```
joshsmith, line 16: I think style is a bit off here.

```
ids = 
  records
 |> ...
```

test/controllers/organization_memberships_controller_test.exs
```diff
@@ -0,0 +1,179 @@
...
+
+  test "filters resources on index", %{conn: conn} do
+    membership_1 = insert(:organization_membership)
+    membership_2 = insert(:organization_membership)
+    insert(:organization_membership)
+
+    params = filter_params([membership_1, membership_2])
+    path = conn |> organization_membership_path(:index)
+
+    data = conn |> get(path, params) |> json_response(200) |> Map.get("data")
```
joshsmith, line 36: Could the `path` and `data` be combined into pipes given that `conn` is the link between them and `data` is what we want at the end?

test/models/organization_membership_test.exs
```diff
@@ -0,0 +1,63 @@
...
+  alias CodeCorps.OrganizationMembership
+
+  describe "update_changeset" do
+    @valid_attrs %{role: "admin"}
+    @invalid_attrs %{}
+
+    test "changeset with valid attributes" do
+      changeset = OrganizationMembership.update_changeset(%OrganizationMembership{}, @valid_attrs)
+
+      assert changeset.valid?
```
joshsmith, line 14: Would remove this space.

test/support/factories.ex
```diff
@@ -0,0 +1,27 @@
...
+    %CodeCorps.User{
+      username: sequence(:username, &"user#{&1}"),
+      email: sequence(:email, &"email-#{&1}@example.com"),
+      password: "password"
+    }
+  end
+
+  def organization_factory do
+    %CodeCorps.Organization{
+      name: sequence(:username, &"Organiazation #{&1}"),
```
joshsmith: Spelling.

test/test_helper.exs
```diff
@@ -1,3 +1,5 @@
+{:ok, _} = Application.ensure_all_started(:ex_machina)
```
joshsmith, line 3: Can we document what this means?

web/controllers/organization_membership_controller.ex
```diff
@@ -0,0 +1,82 @@
+defmodule CodeCorps.OrganizationMembershipController do
+  use CodeCorps.Web, :controller
+
+  import CodeCorps.ControllerHelpers
+
+  alias JaSerializer.Params
+
+  alias CodeCorps.OrganizationMembership
```
joshsmith, line 7: Would group `alias`es together. Do we know what preferred ordering of `use` `import` and `alias` is?

web/controllers/organization_membership_controller.ex
```diff
@@ -0,0 +1,82 @@
...
+  def create(conn, %{"data" => data = %{"type" => "organization-membership", "attributes" => _params}}) do
+    changeset = %OrganizationMembership{} |> OrganizationMembership.create_changeset(Params.to_attributes(data))
+
+    case Repo.insert(changeset) do
+      {:ok, membership} ->
+
+        conn
+        |> put_status(:created)
+        |> put_resp_header("location", organization_membership_path(conn, :show, membership))
+        |> render("show.json-api", data: membership |> Repo.preload([:member, :organization]))
```
joshsmith: The piping here reads a little ambiguously. Should we handle this differently?

web/models/organization_membership.ex
```diff
@@ -0,0 +1,41 @@
...
+  Represents a membership of a user in an organization.
+  """
+
+  use CodeCorps.Web, :model
+
+  alias CodeCorps.Organization
+  alias CodeCorps.User
+
+
+  schema "organization_memberships" do
```
joshsmith, line 11: Extra space above this line.

web/models/organization_membership.ex
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  @doc """
+  Builds a changeset based on the `struct` and `params`, for updating a record.
+  """
+  def update_changeset(struct, params \\ %{}) do
+    struct
+    |> cast(params, [:role])
+    |> validate_required([:role])
+  end
```
joshsmith, line 38: I think we're missing this pattern from other models. We should apply there. Can we look and open issues for any that require it?

web/router.ex
```diff
@@ -50,6 +52,8 @@ defmodule CodeCorps.Router do
     resources "/user-skills", UserSkillController, except: [:update]
     resources "/role-skills", RoleSkillController, except: [:update]
 
+    resources "/organization-memberships", OrganizationMembershipController, only: [:index, :show, :create, :update, :delete]
```
joshsmith: Would move this up to keep alphabetical ordering.

joshsmith:
Some minor comments here that could use some adjustment and responses. Generally good to merge without need for much discussion, unless you feel it's required.

priv/repo/migrations/20160818132546_add_organization_membership.exs
begedin: Not really. Do we have a scheme we could go with?

test/controllers/organization_memberships_controller_test.exs
begedin, line 36: If there is a way, I don't know it. Is there a way to combine two pipes into one?

joshsmith, line 36: @begedin wouldn't it be:

```
data = conn
  |> organization_membership_path(:index)
  |> get(path, params)
  |> json_response(200)
  |> Map.get("data")
```

Roughly?

web/controllers/organization_membership_controller.ex
begedin, line 7: Based on https://github.com/levionessa/elixir_style_guide#modules, which, according to stack overflow is the most commonly used style guide, it's

```
@moduledoc
use
import
alias
require
@type
@module_attribute
```

priv/repo/migrations/20160818132546_add_organization_membership.exs
joshsmith: Is it required? I thought you don't need to be explicit with the name on indexes?

web/controllers/organization_membership_controller.ex
joshsmith, line 7: Okay it looks like they group tightly together and prefer alpha subordering. Let's do that here.

web/models/organization_membership.ex
begedin, line 38: Which pattern do you mean? `validate_required` ?

web/router.ex
begedin: I can do that, but the slug stuff needs to be on the bottom, so I'll keep it there regardless of ordering.

web/models/organization_membership.ex
joshsmith, line 38: No, the `update_changeset` that doesn't include the `ids`, for our join models.

web/router.ex
joshsmith: Yes, that one I know.

priv/repo/migrations/20160818132546_add_organization_membership.exs
begedin: Wasn't aware of that. I'll give it a try

test/controllers/organization_memberships_controller_test.exs
begedin, line 36: No. `conn |> organization_membership_path(:index)` returns a string (the path). That then gets piped further, so we end up actually calling `get("the_path", path<undefined>, params)`.

web/models/organization_membership.ex
begedin, line 38: Ah. Makes sense. I'll make a pass through the existing code and write an issue.

test/controllers/organization_memberships_controller_test.exs
joshsmith, line 36: Okay, got it.

joshsmith:
@begedin this looks good to merge now. Want to squash and 🚢 🇮🇹 ?

web/models/organization_membership.ex
begedin, line 38: Issue is only with projects - #93 

Any other model that doesn't differentiate also doesn't have an `:update` endpoint, so it's not a problem.

joshsmith, line 38: Okay, thanks! 👍

PR #89, opened 2016-08-18
[POST]: WIP update related data

snewcomer:
Trying to get an initial update of related data (post -> project_id).  Noticed that `preview` uses `put_assoc` for updating the previews user.

Any thoughts @begedin on how to get the update working without `put_assoc`?  Also plan on adding more assertions for related data once we figure that out.

Failing line is right here: https://github.com/code-corps/code-corps-phoenix/pull/89/files#diff-8f3513cf21a0b7b4702df8db8016d51fR155

### DIFF
--- a/test/controllers/post_controller_test.exs
+++ b/test/controllers/post_controller_test.exs
@@ -98,8 +98,12 @@ defmodule CodeCorps.PostControllerTest do
       }
     }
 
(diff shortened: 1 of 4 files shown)


joshsmith:
This is missing the `user` relationship, too, isn't it?

snewcomer:
yep was just getting started with projects first

PR #92, opened 2016-08-18
Add skills seed

wolverineks:


priv/repo/seeds.exs
```diff
@@ -1,5 +1,37 @@
...
+  %{
+    title: "Ruby",
+  },
+  %{
+    title: "Ruby on Rails",
+  },
+  %{
+    title: "Docker",
+  },
+]
```
joshsmith, line 24: Super picky of me, but would you mind alphabetizing? Reads better.

joshsmith:
Once you make that minor change, would you mind rebasing off `develop` and squashing? I would also generally avoid merging `develop` into this branch and instead choosing to rebase, generally, given our preference for squashing prior to merge.

joshsmith:
Also, you have write privileges (once you accept the invite), so would also prefer you branch directly off our `develop` instead of your own fork. Makes it easier for us to jump in and help, too.

PR #95, opened 2016-08-18
Add roles seed

wolverineks:


joshsmith:
@wolverineks would you mind labelling as `awaiting review` when this is ready to review?

PR #100, opened 2016-08-19
Add user photo

joshsmith:
This is very much not working yet. I can't figure out at all how to actually handle receiving the file from Ember. We expect a base64 encoded image right now. I cannot seem to make this work. It looks like `Plug` automatically parses for files and handles those directly.

Possible that we need to rewrite the Ember side to handle the image uploads as form data.


begedin:
I got this to a point where, when running a test, arc tries to connect and then times out. If I export environment variables we use in code-corps-api, then I get

```
test/models/user_test.exs:101
     ** (EXIT from #PID<0.497.0>) an exception was raised:
         ** (FunctionClauseError) no function clause matching in String.contains?/2
             (elixir) lib/string.ex:1712: String.contains?({:system, "S3_BUCKET"}, ".")
             (ex_aws) lib/ex_aws/s3/request.ex:40: ExAws.S3.Request.host_and_bucket/2
             (ex_aws) lib/ex_aws/s3/request.ex:26: ExAws.S3.Request.url/3
             (ex_aws) lib/ex_aws/s3/request.ex:12: ExAws.S3.Request.request/5
             lib/arc/storage/s3.ex:14: Arc.Storage.S3.put/3
             (elixir) lib/task/supervised.ex:94: Task.Supervised.do_apply/2
             (elixir) lib/task/supervised.ex:45: Task.Supervised.reply/5
             (stdlib) proc_lib.erl:240: :proc_lib.init_p_do_apply/3
```

I guess the bucket URL is wrong, or the instructions for arc are bad. If I replace the bucket config line with a direct value instead of a Key, I get a different error

```
test/models/user_test.exs:101
     ** (MatchError) no match of right hand side value: {:error, [{:http_error, 403, "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<Error><Code>InvalidAccessKeyId</Code>
....
```

At this point, my guess is, it should work, and what's missing is proper configuration on my part.

joshsmith:
I can't quite figure out why this build is failing even after adding in the env vars to CI.

test/models/user_test.exs
```diff
@@ -97,5 +97,20 @@ defmodule CodeCorps.UserTest do
...
+    test "uploads base64photo data to aws" do
+      # 1x1 black pixel gif
+      photo_data = "data:image/gif;base64,R0lGODlhAQABAIAAAAUEBAAAACwAAAAAAQABAAACAkQBADs="
+      user = insert_user
+      attrs = %{base64_photo_data: photo_data}
+
+      changeset = User.update_changeset(user, attrs)
+
+      IO.inspect(System.get_env("CLOUDFRONT_DOMAIN"))
+      IO.inspect(changeset)
```
begedin: Leftover inspects?

begedin:
Looks good other than the leftover inspects. :+1:

PR #101, opened 2016-08-19
Add users seed

wolverineks:
Changed the test user password from "test" to "test123" to meet validations


npendery:
Isn't this #79?

joshsmith:
The important part of this is the change to the password length.

Issue #109, opened 2016-08-21
Refactor tests to not rely on ENV vars


joshsmith:
We have some tests, like photo uploads, that rely on secret ENV vars. We should allow people to run tests excluding these.

This will require updating the README and adding some config to the tests.

joshsmith:
@green-arrow what else do we need to do to consider this closed?

green-arrow:
I think we should be able to consider this closed since we have a pattern down now for excluding those kinds of tests.

PR #115, opened 2016-08-22
Added user_roles migration, model, endpoint.

green-arrow:
Also added a tag to add to all endpoints that require interaction with AWS S3 that are ignored in tests by default.

To run all tests, including those which interface with S3, run `mix test --include s3`

Endpoint issue: #40
Contributes to #109 (exclude tests that rely on ENV vars)

### DIFF
--- a/priv/repo/migrations/20160822011624_create_user_role.exs
+++ b/priv/repo/migrations/20160822011624_create_user_role.exs
@@ -1,0 +1,14 @@
+defmodule CodeCorps.Repo.Migrations.CreateUserRole do
+  use Ecto.Migration
+
+  def change do
+    create table(:user_roles) do
+      add :user_id, references(:users, on_delete: :nothing)
+      add :role_id, references(:roles, on_delete: :nothing)
+
+      timestamps()
+    end
+
+    create index(:user_roles, [:user_id, :role_id], unique: true)
+  end
+end
--- a/test/controllers/user_controller_test.exs
+++ b/test/controllers/user_controller_test.exs
@@ -126,6 +126,7 @@ defmodule CodeCorps.UserControllerTest do
       assert user.biography == "Just a test user"
     end
 
+    @tag :requires_env
     test "uploads a photo to S3", %{conn: conn} do
       user = insert_user()
       photo_data = "data:image/gif;base64,R0lGODlhAQABAIAAAAUEBAAAACwAAAAAAQABAAACAkQBADs="
--- a/test/test_helper.exs
+++ b/test/test_helper.exs
@@ -2,6 +2,9 @@
 # Needs to be called before ExUnit.start
 {:ok, _} = Application.ensure_all_started(:ex_machina)
 
+# By default, exclude S3 tests
+# To run S3 tests, run `mix test --include requires_env`
+ExUnit.configure exclude: [:requires_env]
 ExUnit.start
 
 Ecto.Adapters.SQL.Sandbox.mode(CodeCorps.Repo, :manual)
--- a/test/controllers/organization_controller_test.exs
+++ b/test/controllers/organization_controller_test.exs
@@ -74,6 +74,7 @@ defmodule CodeCorps.OrganizationControllerTest do
     assert organization.id == slugged_route.organization_id
   end
 
+  @tag :requires_env
   test "uploads a icon to S3", %{conn: conn} do
     organization = insert_organization()
     icon_data = "data:image/gif;base64,R0lGODlhAQABAIAAAAUEBAAAACwAAAAAAQABAAACAkQBADs="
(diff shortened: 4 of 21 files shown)


test/test_helper.exs
```diff
@@ -2,6 +2,9 @@
 # Needs to be called before ExUnit.start
 {:ok, _} = Application.ensure_all_started(:ex_machina)
 
+# By default, exclude S3 tests
+# To run S3 tests, run `mix test --include s3`
+ExUnit.configure exclude: [:s3]
```
joshsmith: I'm a little concerned about making this the default. I'd like to go the other way I think, and add the `exclude` command in the `README`.

We almost want to encourage people to get their own AWS account, too, so they can actually play around with some of that if they'd like to.

What do you think? It's scary to me to rely on a default that ignores some tests.

priv/repo/migrations/20160822011624_create_user_role.exs
```diff
@@ -0,0 +1,14 @@
...
+
+  def change do
+    create table(:user_roles) do
+      add :user_id, references(:users, on_delete: :nothing)
+      add :role_id, references(:roles, on_delete: :nothing)
+
+      timestamps()
+    end
+
+    create index(:user_roles, [:user_id, :role_id], unique: true, name: :index_projects_on_user_id_role_id)
```
joshsmith: Is the name strictly necessary or will it be auto-generated?

test/controllers/user_controller_test.exs
```diff
@@ -126,6 +126,7 @@ defmodule CodeCorps.UserControllerTest do
       assert user.biography == "Just a test user"
     end
 
+    @tag :s3
```
joshsmith: Is there a more generalizable tag we could use here, like `requires_env` or something?

joshsmith:
I would add a `UserRoleViewTest` to this (you can see examples scattered throughout the codebase, including my open PR, which needs review _wink wink_).

I would also consider doing a `UserViewTest` and ensuring that the `has_many` roles through is present there.

Minor comments on this overall, and generally looks good. Still might want to re-review after changes, so maybe add a commit and don't squash so we can review only new changes.

test/test_helper.exs
green-arrow: So the reason I made this excluded by default is that these tests should be in the minority. The majority of the time, we won't be interacting with tests involving S3 / other secret env variables. Since it's likely most people won't mess with those tests, I think it makes sense to exclude by default and call it out in the README.

Also, as long as CI is set up to include those tests, we'll be sure to catch anything before it gets merged.

If we do want to include these by default, just let me know and I'll change it. Just wanted to give my two cents :)

priv/repo/migrations/20160822011624_create_user_role.exs
green-arrow: I believe the name will be auto-generated. This was taken from the `user_skills` implementation. Do we prefer to use the auto-generated names?

test/controllers/user_controller_test.exs
green-arrow: Do we see there being a distinction between interacting with third party services / general / other secret env variables? I'm just wondering how granular we want to make it?

I guess we could always make it generically `requires_env` now then get more granular in the future if we ever need to.

test/test_helper.exs
joshsmith: Okay, I just worry about missing this on CI. Already I have to remember to include it there in the config.

priv/repo/migrations/20160822011624_create_user_role.exs
joshsmith: I prefer auto-generated, yes.

test/controllers/user_controller_test.exs
joshsmith: I think generic > granular now.

test/test_helper.exs
green-arrow: Understandable. I can make it included by default. I'll make sure to point that out in the README as well.

joshsmith: I mean, I can add to CI now if you'd like. I definitely see what you're saying. Maybe let's try it this way and see if it becomes an issue for folks later. Then we can fix.

green-arrow:
@JoshSmith for the `UserViewTest`, adding that relationship causes some other tests (like `UserController` to break because the relationship has not yet been loaded. I don't see an endpoint like `/users/:id/roles`, so do we want to force pre-fetching that relationship?

joshsmith:
@green-arrow yes, so there's no specific endpoint for it, but we'll preload the roles to avoid the N+1 hit while still having their ids present in the `relationships` for the view. It's necessary for the onboarding, for example, and elsewhere throughout the app.

joshsmith:
To be clear, though, we're not `include`-ing these. Ember will go out and hit them with coalesced requests to the `user_roles` endpoint.

green-arrow:
Gotcha, sounds good

green-arrow:
@JoshSmith - I believe this is ready for review again

joshsmith:
👍

PR #119, opened 2016-08-22
Project skill

snewcomer:


snewcomer:
@JoshSmith So fyi to get skills as a m2m, I needed to project_skills as a has_many.  But the basic end scenario that projects need to have skill data that comes down in the relationship block and no other relationship data, correct?

https://github.com/code-corps/code-corps-phoenix/pull/119/files#diff-509ee0fad23784b36015246631733114R28

npendery:
Just a thought, but we could simplify the codebase by using Ecto's [many to many](https://hexdocs.pm/ecto/Ecto.Schema.html#many_to_many/3). Then it is just a table in the DB and we wouldnt need the model or controller for `ProjectSkill`

joshsmith:
@npendery the reason we're keeping those joins, largely, is in the event we need to add additional data to it. For example, a `ProjectSkill` could be marked as `required` or something. I'd prefer the flexibility even if it means a slight disadvantage in terms of complexity.

snewcomer:
Sounds reasonable.  Lmk if I should change anything.

npendery:
Wouldn't those extra requirements just sit in either Project or Skill though? I might be totally wrong on this but the single purpose for join tables is to connect two tables with a single id for each? Any other logic should be handled in either of the two tables that are being joined.

joshsmith:
@npendery no, that's not accurate. Look at `OrganizationMembership` for example. The only to understand the role level of someone in an organization is through that particular relation. `Organization` does not have that responsibility and nor does `User`. It's only the join – the membership itself – that contains that information.

Important to also note that we're still simplifying the logic on the Ember side by not explicitly using the joins that don't have extra non-relational data.

npendery:
Yup okay makes sense, thanks for the breakdown!

joshsmith:
Of course! It's rather odd at first glance and I hadn't really dealt with something like that much before this project, except for maybe on CookAcademy stuff, but it's a nice (and actually pretty simple) approach to dealing with some complex relationship data.

snewcomer:
@JoshSmith So a few things:
1.  Any examples for setting up the m2m relationship in a controller test (these assertions need to be added)?  Should I be using the factories instead?
2.  What indexes / other stuff needs to be added to the migration file for this m2m relationship?

joshsmith:
@snewcomer I think there are some in the `user_` joins.

The kinds of indexes you see there are similar to the ones you'd see in those cases, I believe.

snewcomer:
@JoshSmith my only two thoughts were 1.  What I needed to add to the `repo/migrations` file for m2m on the project model.  I see indexes for belongs_to but not for m2m.  Maybe that is not necessary 2.  Was going to change over to the factories, but bailed on that...unless you want that as a req.

snewcomer:
I'll work on switching over to factories tonight

joshsmith:
@snewcomer I would look to other code as examples, e.g.:

```
defmodule CodeCorps.Repo.Migrations.CreateUserRole do
  use Ecto.Migration

  def change do
    create table(:user_roles) do
      add :user_id, references(:users, on_delete: :nothing)
      add :role_id, references(:roles, on_delete: :nothing)

      timestamps()
    end

    create index(:user_roles, [:user_id, :role_id], unique: true)
  end
end
```

You could effectively replace `user` with `project` and `role` with `skill` and get exactly what you need on the join table.

joshsmith:
I would also probably add indexes for `user_id` and `role_id` individually, too, which appears to be missing. Can you create an issue to look for and add missing indexes?

snewcomer:
Will do tonight.

joshsmith:
You'll need to rebase off `develop`, too, before you dive back into this.

test/controllers/project_skill_controller_test.exs
```diff
@@ -0,0 +1,135 @@
...
+  setup do
+    conn = %{build_conn | host: "api."}
+      |> put_req_header("accept", "application/vnd.api+json")
+      |> put_req_header("content-type", "application/vnd.api+json")
+
+    {:ok, conn: conn}
+  end
+
+  defp attributes do
+  end
```
joshsmith: I would either remove this altogether or just use `%{}`

test/models/project_skill_test.exs
```diff
@@ -0,0 +1,61 @@
+defmodule CodeCorps.ProjectSkillTest do
+  use CodeCorps.ModelCase
+
+  alias CodeCorps.ProjectSkill
+
+  @valid_attrs %{}
+  @invalid_attrs %{}
```
joshsmith: If we're not using these then remove them.

test/support/test_helpers.ex
```diff
@@ -0,0 +1,115 @@
...
+  def insert_comment(attrs \\ %{}) do
+    changes = Map.merge(%{
+      markdown: "some content",
+    }, attrs)
+
+    %Comment{}
+    |> Comment.create_changeset(changes)
+    |> Repo.insert!
+  end
+end
```
joshsmith, line 115: This whole file needs to be deleted.

test/views/project_view_test.exs
```diff
@@ -57,6 +58,22 @@ defmodule CodeCorps.ProjectViewTest do
...
+            ]
+          },
+          "project-skills" => %{
+            data: [
+              %{
+                id: project_skill.id |> Integer.to_string,
+                type: "project-skill"
+              }
+            ]
+          },
```
joshsmith, line 76: I would alpha order the keys.

web/models/project.ex
```diff
@@ -25,6 +25,8 @@ defmodule CodeCorps.Project do
 
     has_many :project_categories, CodeCorps.ProjectCategory
     has_many :categories, through: [:project_categories, :category]
+    has_many :project_skills, CodeCorps.ProjectSkill
+    has_many :skills, through: [:project_skills, :skill]
```
joshsmith, line 30: I may suggest separating this `has_many` two-fer group to make it a little easier on the eyes.

web/models/skill.ex
```diff
@@ -6,6 +6,8 @@ defmodule CodeCorps.Skill do
     field :description, :string
     field :original_row, :integer
 
+    has_many :project_skills, CodeCorps.ProjectSkill
+    has_many :projects, through: [:project_skills, :project]
     timestamps()
```
joshsmith, line 12: Would add a line break above `timestamps()`.

web/router.ex
```diff
@@ -61,6 +61,7 @@ defmodule CodeCorps.Router do
     resources "/user-roles", UserRoleController, only: [:create, :delete]
     resources "/user-skills", UserSkillController, only: [:index, :show, :create, :delete]
     resources "/role-skills", RoleSkillController, only: [:index, :show, :create, :delete]
+    resources "/project-skills", ProjectSkillController, only: [:index, :show, :create, :delete]
```
joshsmith: Would reorder this alpha by resources, and cleanup other things while you're in there (like `/role-skills`).

web/views/project_skill_view.ex
```diff
@@ -0,0 +1,8 @@
+defmodule CodeCorps.ProjectSkillView do
+  use CodeCorps.Web, :view
+  use JaSerializer.PhoenixView
+
+  has_one :project, serializer: CodeCorps.ProjectView
+  has_one :skill, serializer: CodeCorps.SkillView
+
```
joshsmith: Remove the empty line here.

web/views/project_view.ex
```diff
@@ -11,6 +11,8 @@ defmodule CodeCorps.ProjectView do
 
   has_many :project_categories, serializer: CodeCorps.ProjectCategoryView
   has_many :categories, serializer: CodeCorps.CategoryView
+  has_many :project_skills, serializer: CodeCorps.ProjectSkillView
+  has_many :skills, serializer: CodeCorps.SkillView
```
joshsmith, line 16: Same grouping remark as above.

joshsmith:
Looks excellent! Really minor feedback. Once implemented you can feel free to rebase, squash, and merge at will.

Great work @snewcomer 👍

Issue #26, opened 2016-08-01
Migrate Rails API to Phoenix


joshsmith:
This is a journey issue that collects all the basic API endpoints to achieve parity with the soon-to-be deprecated Rails API.
## Minor features
- [x] **OrganizationMemberhship** missing ability to filter by role. This is currently an option for the "filter by organization slug" index endpoint on the rails api #132 _in progres_
- [x] **RoleSkill** - missing the `cat` enum field #129 
- [ ] Duplicate validation across all sluggable resources that clash
- [ ] Attachment validations (size, type, etc.)
## Issues
- [ ] **Skill** auto complete behavior - #60 among other issues
- [x] `&Category.changeset/2` should not permit slug as a parameter. #133
## Documentation
- [ ] Rewrite documentation to account for API changes
## Testing
- [ ] Auth view
- [ ] Changeset view
- [ ] Comment view
- [ ] Error helpers view
- [x] Organization membership view #124 _in progress_
- [x] Organization view #124 _in progress_
- [x] Post view _in progress_
- [ ] Preview view
- [ ] Project category view
- [x] Role skill view _in progress_
- [x] Role view _in progress_
- [x] Skill view _in progress_
- [ ] Slugged route view
## Maintenance & refactoring
- [ ] **Something to consider:** Write an ex_machina strategy which would allow us simple building of JSON_API payloads for tests -https://github.com/thoughtbot/ex_machina#custom-strategies
## Low priority features
- [ ] Find a way to test concurrency/possible race conditions for post numbering
- [ ] Obscenity validations
- post and comment images
- post likes
- mentions
- notifications
- user relationships (followers)
## Comments API #28, #82
- [x] Rename to `Comment` from `PostComment`
## A `User` has roles - _in progress_ #115
- [x] `UserRoles` API #40 
- [x] Add relationships to `User` - has many `UserRoles` and `Roles`
- [x] Add relationships to `Role` - has many `UserRoles` and `Users`
## A `Project` has categories - _in progress_ #105
- [x] `ProjectCategories` API #33 , #105
- [x] Add relationships to `Project` - has many `ProjectCategories` and `Categories`
- [x] Add relationships to `Category` - has many `ProjectCategories` and `Projects`
## A `User` has categories - _in progress_ #118
- [x] `UserCategories` API #39 
- [x] Add relationships to `User` - has many `UserCategories` and `Categories`
- [x] Add relationships to `Category` - has many `UserCategories` and `Users`
[...]

PR #126, opened 2016-08-23
Add roles for organization membership

joshsmith:
Part of #26.

This adds missing roles and role validation with tests.

Also minor removal of an unused `import` in an unrelated file that was bothering me.

### DIFF
--- a/web/models/organization_membership.ex
+++ b/web/models/organization_membership.ex
@@ -18,12 +18,23 @@ defmodule CodeCorps.OrganizationMembership do
   end
 
   @doc """
+  Builds a changeset based on the `struct` and `params`.
+  """
+  def changeset(struct, params \\ %{}) do
+    struct
+    |> cast(params, [:role])
+    |> validate_required([:role])
+    |> validate_inclusion(:role, roles)
+  end
+
+  @doc """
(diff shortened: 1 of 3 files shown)


joshsmith:
If someone could review this, that would be awesome.

web/models/organization_membership.ex
```diff
@@ -18,23 +18,37 @@ defmodule CodeCorps.OrganizationMembership do
...
   def create_changeset(struct, params \\ %{}) do
     struct
-    |> cast(params, [:member_id, :organization_id, :role])
-    |> validate_required([:member_id, :organization_id, :role])
+    |> changeset(params)
+    |> cast(params, [:member_id, :organization_id])
+    |> validate_required([:member_id, :organization_id])
     |> assoc_constraint(:member)
     |> assoc_constraint(:organization)
+    |> validate_inclusion(:role, roles)
```
npendery: `validating _inclusion` twice, here and line 27 under changeset

npendery:
Besides my one comment, 🐑 🇮🇹

web/models/organization_membership.ex
joshsmith: Good catch! Fixed now.

PR #134, opened 2016-08-24
Add user validations for email and username uniqueness

### LINKED ISSUE #128: User validations - uniqueness of email and username, etc
We need to add user validations that don't exist.

For #26.

joshsmith:
Closes #128.

Adds user validations for both email and username uniqueness and tests for them.

### DIFF
--- a/test/models/user_test.exs
+++ b/test/models/user_test.exs
@@ -22,25 +22,47 @@ defmodule CodeCorps.UserTest do
     assert {:email, {"has invalid format", []}} in changeset.errors
   end
 
-  test "registration_changeset does not accept long usernames" do
-    attrs = Map.put(@valid_attrs, :username, String.duplicate("a", 40))
-    changeset = User.registration_changeset(%User{}, attrs)
-    assert {:username, {"should be at most %{count} character(s)", count: 39}} in changeset.errors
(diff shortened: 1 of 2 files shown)


npendery:
Good to go! :shipit:

test/models/user_test.exs
```diff
@@ -22,25 +22,47 @@ defmodule CodeCorps.UserTest do
...
+      changeset = User.registration_changeset(%User{}, user_2_attrs)
+      {:error, changeset} = Repo.insert(changeset)
+      refute changeset.valid?
+      assert changeset.errors[:email] == {"has already been taken", []}
+    end
+
+    test "does not allow duplicate usernames, regardless of case" do
+      user_1_attrs = %{email: "user_1@email.com", password: "password", username: "duplicate"}
+      user_2_attrs = %{email: "user_2@email.com", password: "password", username: "DUPLICATE"}
+      insert_user(user_1_attrs)
```
npendery, line 60: Do we want to use factories for these inserts for consistency?

joshsmith: Yes but I'll wait for the PR exclusive to that one. That should be next priority anyway.

PR #139, opened 2016-08-24
Add role search to organization memberships

### LINKED ISSUE #132: Add ability to filter by role to organization memberships controlller
Right now missing ability to filter by role. This is currently an option for the "filter by organization slug" index endpoint on the Rails api:

```
    def filter_params
      filter_params = {}
      filter_params[:role] = params[:role] if params[:role]
      filter_params[:id] = id_params if params.fetch(:filter, {})[:id]
      filter_params
    end
```

For #26.

npendery:
Closes #132

### DIFF
--- a/test/controllers/organization_memberships_controller_test.exs
+++ b/test/controllers/organization_memberships_controller_test.exs
@@ -5,15 +5,6 @@ defmodule CodeCorps.OrganizationMembershipControllerTest do
   alias CodeCorps.Organization
   alias CodeCorps.User
 
-  def filter_params(records) do
-    ids =
-      records
-      |> Enum.map(fn(r) -> r.id end)
-      |> Enum.join(",")
-
-    %{filter: %{id: ids}}
-  end
-
   setup do
     conn =
       %{build_conn | host: "api."}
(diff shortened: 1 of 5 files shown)


web/controllers/organization_membership_controller.ex
```diff
@@ -19,6 +19,17 @@ defmodule CodeCorps.OrganizationMembershipController do
...
+  def index(conn, %{"role" => roles}) do
+    roles = roles |> coalesce_string
+    memberships =
+      OrganizationMembership
+      |> preload([:organization, :member])
+      |> where([p], p.role in ^roles)
+      |> Repo.all
+
+    render(conn, "index.json-api", data: memberships)
+  end
```
joshsmith: It looks like the problem with this approach is that the two could be combined: you may be filtering either `ids` or `roles`.

I'm not sure what the actual requests look like that are generating the filtering, and if they're both used, but this does look like a bit of an explosion of endpoints and some duplication, as well as possibly missing combinations of request types.

joshsmith:
You'll need to rebase off `develop` now.

web/controllers/organization_membership_controller.ex
```diff
@@ -8,23 +8,51 @@ defmodule CodeCorps.OrganizationMembershipController do
...
+          ids = id_list |> coalesce_id_string
+          OrganizationMembership
+            |> where([om], om.member_id in ^ids)
+        %{"role" => roles} ->
+          roles = roles |> coalesce_string
+          OrganizationMembership
+            |> where([om], om.role in ^roles)
+        _ ->
+          OrganizationMembership
+      end
```
green-arrow: I would personally simplify this into a couple different methods:

``` elixir
def index(conn, params) do
  memberships = OrganizationMembership
  |> with_filters(params)
  |> preload([:organization, :member])
  |> Repo.all

  render(conn, "index.json-api", data: memberships)
end

defp with_filters(query, params) do
  query
  |> with_organization_filter(params)
  |> with_role_filter(params)
  |> with_member_filter(params)
end

defp with_organization_filter(query, %{"organization_id" => organization_id}) do
  query |> where([om], om.organization_id == ^organization_id)
end
defp with_organization_filter(query, _), do: query

defp with_role_filter(query, %{"roles" => roles}) do
  roles = roles |> coalesce_string
  query |> where([om], om.role in ^roles)
end
defp with_role_filter(query, _), do: query

defp with_member_filter(query, %{"filter" => id_list}) do
  ids = id_list |> coalesce_id_string
  query |> where([om], om.member_id in ^ids)
end
defp with_member_filter(query, _), do: query
```

joshsmith:
Looks like you have just one failure:

```
  1) test filters resources on index (CodeCorps.OrganizationMembershipControllerTest)
     test/controllers/organization_memberships_controller_test.exs:26
     ** (FunctionClauseError) no function clause matching in String.split/3
```

npendery:
Yeah, still need to edit/add tests

test/controllers/organization_memberships_controller_test.exs
```diff
@@ -23,36 +14,112 @@ defmodule CodeCorps.OrganizationMembershipControllerTest do
...
+
+    test "filters resources by member id", %{conn: conn} do
+      member_1 = insert(:user)
+      member_2 = insert(:user)
+      member_3 = insert(:user)
+      membership_1 = insert(:organization_membership, member: member_1)
+      membership_2 = insert(:organization_membership, member: member_2)
+      insert(:organization_membership, member: member_3)
+
+
```
begedin: Bit nitpicky, but mind ditching the extra empty line here? One is enough.

begedin:
Other than that extra empty line, looks good to go. I would like to raise the question of code organization, though. I think we might have gotten to the point where `model_helpers` and `controller_helpers` aren't the correct names anymore. 

Basically, everything in our controller helpers and most of model helpers fits together, maybe into something like a `CodeCorps.Filtering` module?. 

@JoshSmith Do we want to create an issue for that?

I think this is good to merge otherwise. If we want to reorganize, it should go with the separate issue.

joshsmith:
@begedin yes let's please create a separate issue.

PR #142, opened 2016-08-25
Update icon URLs to not use relative paths

joshsmith:
This PR updates icon URLs to not use relative paths. Maybe could use regression tests, but probably not a huge deal.


npendery:
Nice, just learned some new things by reviewing! 👍

PR #144, opened 2016-08-25
Add deserializer for JSON API requests

joshsmith:
Very simple.

This adds deserialization to take something like `first-name` in the JSON attributes hash keys and turn it into `first_name` for use in Phoenix itself.

Not sure how to test, or if it's really necessary right now.


npendery:
🚢 🇮🇹

PR #146, opened 2016-08-25
Add posts to project and add post view test

### LINKED ISSUE #130: Add has_many posts to project
`Project` right now is missing the `has_many :posts`. We need to add to both the model and the view.

For #26.

joshsmith:
Closes #130.

Simply adds posts to projects, expands on post some and tests it.

### DIFF
--- a/test/views/post_view_test.exs
+++ b/test/views/post_view_test.exs
@@ -1,0 +1,63 @@
+defmodule CodeCorps.PostViewTest do
+  use CodeCorps.ConnCase, async: true
+
+  alias CodeCorps.Repo
+
+  import Phoenix.View, only: [render: 3]
+
+  test "renders all attributes and relationships properly" do
+    post = insert(:post)
+    comment = insert(:comment, post: post)
+
+    post =
+      CodeCorps.Post
+      |> Repo.get(post.id)
+      |> CodeCorps.Repo.preload([:comments, :project, :user])
+
+    rendered_json =  render(CodeCorps.PostView, "show.json-api", data: post)
+
+    expected_json = %{
+      data: %{
+        attributes: %{
+          "body" => post.body,
+          "inserted-at" => post.inserted_at,
+          "markdown" => post.markdown,
+          "number" => post.number,
+          "post-type" => post.post_type,
+          "status" => post.status,
+          "title" => post.title,
+          "updated-at" => post.updated_at,
+        },
+        id: post.id |> Integer.to_string,
+        relationships: %{
+          "comments" => %{
+            data: [
+              %{
+                id: comment.id |> Integer.to_string,
+                type: "comment"
+              }
+            ]
+          },
+          "project" => %{
+            data: %{
+              id: post.project_id |> Integer.to_string,
+              type: "project"
+            }
+          },
+          "user" => %{
+            data: %{
+              id: post.user_id |> Integer.to_string,
+              type: "user"
+            }
+          }
+        },
+        type: "post",
+      },
+      jsonapi: %{
+        version: "1.0"
+      }
+    }
+
+    assert rendered_json == expected_json
+  end
+end
--- a/web/controllers/post_controller.ex
+++ b/web/controllers/post_controller.ex
@@ -6,16 +6,12 @@ defmodule CodeCorps.PostController do
 
   plug :scrub_params, "data" when action in [:create, :update]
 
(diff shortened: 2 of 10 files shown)


joshsmith:
Ready for review.

web/controllers/post_controller.ex
```diff
@@ -7,15 +7,18 @@ defmodule CodeCorps.PostController do
...
 
   def index(conn, %{"project_id" => project_id}) do
-    posts = Repo.all(
-      from c in Post,
-      where: c.project_id == ^project_id,
-      select: c
-    )
+    posts =
+      Post
+      |> Post.for_project(project_id)
```
npendery: I would move this to the model helpers we have been adding as a project filter

web/controllers/post_controller.ex
```diff
@@ -36,22 +41,28 @@ defmodule CodeCorps.PostController do
...
-    post = Repo.one(
-      from c in Post,
-      where: c.number == ^number,
-      where: c.project_id == ^project_id,
-      select: c
-    )
+    post =
+      Post
+      |> Post.for_project(project_id)
+      |> Post.with_number(number)
```
npendery: include this as `show_filters` and add `with_number` to model helpers

web/controllers/post_controller.ex
joshsmith: Which ones?

npendery: [These from my PR](https://github.com/code-corps/code-corps-phoenix/pull/139/files#diff-a8651bfb80ad2539a4f87ff71b4aa05cR16)

npendery:
@JoshSmith the PR with the `ModelHelper` additions is in if you want to add `with_number` and `for_project` to that list

test/views/post_view_test.exs
```diff
@@ -0,0 +1,64 @@
+defmodule CodeCorps.PostViewTest do
+  use CodeCorps.ConnCase, async: true
+
+  alias CodeCorps.Repo
+
+  # Bring render/3 and render_to_string/3 for testing custom views
+  import Phoenix.View
```
HeroicEric: might be a good idea to only import the functions needed

``` elixir
import Phoenix.View, only: [render: 3, render_to_string: 3]
```

joshsmith:
Could use a review on my latest commit.

web/models/model_helpers.ex
```diff
@@ -15,6 +15,7 @@ defmodule CodeCorps.ModelHelpers do
 
   def id_filter(query, %{"filter" => %{"id" => id_list}}) do
     ids = id_list |> coalesce_id_string
+    IO.inspect ids
```
noslouch: probably don't want this log statement here

joshsmith: What if I do?

noslouch: then whomever is watching `stdout` in production will see lots of numbers 👻

web/controllers/project_controller.ex
```diff
@@ -50,7 +50,7 @@ defmodule CodeCorps.ProjectController do
 
     case Repo.insert(changeset) do
       {:ok, project} ->
-        project = Repo.preload(project, [:categories, :organization, :skills])
+        |> Repo.preload([:categories, :organization, :posts, :skills])
```
noslouch: this is dying on semaphore. I don't think the `case` clause heads pass a value to its body, so I think it needs to be something like...

``` elixir
project =
  project
  |> Repo.preload([:categories, :organization, :posts, :skills])
  ...
```

and then again the same thing around line 74

joshsmith: Yeah, this was dirty copy/paste in a merge. Will fix.

begedin:
I did some digging on using `Repo.preload`, vs just `preload`

It looks like just `preload` is actually us using the weird linq/lambda style query synthax, so it works just off of a model module. `Repo.preload`, on the other hand is a method on the `Repo` module, which expects a struct or list of structs as the first argument, so it has to be called after `Repo.get*/Repo.all`. From what I can tell in the documentation and various other places, `Repo.preload` is the preferred approach and more commonly used. We should fallback to `preload` if we need some sort of query interpolation.

I corrected the controller so tests pass now. I also modified the index method a bit to simplify the query and possibly shorten it, making it more performant.

PR #148, opened 2016-08-26
Add role skills, tests for role, skill, role skill view

### LINKED ISSUE #111: Add Skill Roles API
We need to add roles to skills.

joshsmith:
Closes #111.

Adds `has_many` and `belongs_to` for `RoleSkill`, `Role`, and `Skill` to set up relationships. Also adds view tests to be sure they're working appropriately.

Removes some unnecessary calls to `import CodeCorps.Factories` already imported by `ConnCase`.

### DIFF
--- a/web/controllers/role_controller.ex
+++ b/web/controllers/role_controller.ex
@@ -7,7 +7,10 @@ defmodule CodeCorps.RoleController do
   plug :scrub_params, "data" when action in [:create]
 
   def index(conn, _params) do
-    roles = Repo.all(Role)
+    roles =
+      Role
+      |> Repo.all
+      |> Repo.preload([:skills])
     render(conn, "index.json-api", data: roles)
   end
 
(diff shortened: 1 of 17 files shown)


joshsmith:
This is ready for review.

web/controllers/role_controller.ex
```diff
@@ -7,7 +7,10 @@ defmodule CodeCorps.RoleController do
   plug :scrub_params, "data" when action in [:create]
 
   def index(conn, _params) do
-    roles = Repo.all(Role)
+    roles =
+      Role
+      |> Repo.all
+      |> Repo.preload([:skills])
```
npendery, line 13: Does the `preload` need to happen before the `Repo.all` call? Or should we at least put it in front since thats how we do it for all other models?

npendery:
LGTM! :shipit:

Issue #59, opened 2016-08-10
Restrict certain endpoints to admins only


joshsmith:
We have a lot of `create` and `update` actions on endpoints that should only be possible for site-wide admins to hit.

This is blocked by:
- not having any concept of an admin for users
- not having a library for permissions, AKA `canary`

A non-exhaustive list of endpoints and actions that need to be restricted to admins:
- organizations - create
- roles - create, update
- role skills - create, delete
- skills - create, update

joshsmith:
This task is related to #26.

joshsmith:
Closed #149.

PR #149, opened 2016-08-26
Spike canary

joshsmith:
This is a first spike of what `canary` might look like.

Way unfinished but would love thoughts on this now.

Essentially having trouble finding a way to bring in the `current_user` that `canary` needs for requests. Trying to not have the server choke when the token is `nil`.

Had to split apart those endpoints that need authentication from those that do not. Added a `:with_token` pipeline for those that need authed.
### Notes by begedin:

The way canary works, we can't really look at what changes are happening to the record, especially on create. Due to that, organization membership permissions need to be simplified a bit. There is also a list of things that should be converted to tasks, post merge
- For creation of an organization membership, only a "pending" value should be allowed for the role field. This should be done trough the changeset, not permissions.
- More incoming

### DIFF
--- a/web/models/abilities.ex
+++ b/web/models/abilities.ex
@@ -1,0 +1,91 @@
+defmodule Canary.Abilities do
+  alias CodeCorps.Category
+  alias CodeCorps.Comment
+  alias CodeCorps.Organization
+  alias CodeCorps.OrganizationMembership
+  alias CodeCorps.Post
+  alias CodeCorps.Preview
+  alias CodeCorps.Project
+  alias CodeCorps.ProjectCategory
+  alias CodeCorps.ProjectSkill
+  alias CodeCorps.Role
+  alias CodeCorps.RoleSkill
+  alias CodeCorps.Skill
+  alias CodeCorps.User
+  alias CodeCorps.UserCategory
+  alias CodeCorps.UserRole
+  alias CodeCorps.UserSkill
+
+  alias CodeCorps.CategoryPolicy
+  alias CodeCorps.CommentPolicy
+  alias CodeCorps.OrganizationPolicy
+  alias CodeCorps.OrganizationMembershipPolicy
+  alias CodeCorps.PostPolicy
+  alias CodeCorps.PreviewPolicy
+  alias CodeCorps.ProjectPolicy
+  alias CodeCorps.ProjectCategoryPolicy
+  alias CodeCorps.ProjectSkillPolicy
+  alias CodeCorps.RolePolicy
+  alias CodeCorps.RoleSkillPolicy
+  alias CodeCorps.SkillPolicy
+  alias CodeCorps.UserPolicy
+  alias CodeCorps.UserCategoryPolicy
+  alias CodeCorps.UserRolePolicy
+  alias CodeCorps.UserSkillPolicy
+
+  defimpl Canada.Can, for: User do
+    # NOTE: Canary sets an :unauthorized and a :not_found handler on a config level
+    # The problem is, it will still go through the authorization process first and only call the
+    # not found handler after the unauthorized handler does its thing. This means that our
+    # unauthorized handler will halt the connection and respond, so the not_found handler
+    # will never do anything
+    #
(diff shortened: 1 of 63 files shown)


joshsmith:
The tests are failing, rightly, because of `401` errors. That's a step in the right direction, at least.

web/router.ex
```diff
@@ -11,9 +11,13 @@ defmodule CodeCorps.Router do
...
   pipeline :api do
     plug :accepts, ["json-api", "json"]
+    plug JaSerializer.Deserializer
+  end
+
+  pipeline :with_token do
     plug Guardian.Plug.VerifyHeader
     plug Guardian.Plug.LoadResource
-    plug JaSerializer.Deserializer
+    plug CodeCorps.Plug.CurrentUser
```
green-arrow, line 21: I think that instead of a custom plug, we can use `Guardian.Plug.EnsureAuthenticated`. This should cause the resource to be loaded (`guardian_serializer` module) as well as make sure a valid token exists.

web/models/abilities.ex
```diff
@@ -0,0 +1,9 @@
+defmodule Canary.Abilities do
+  alias CodeCorps.{User}
+
+  defimpl Canada.Can, for: User do
+    def can?(current_user, action, user = %User{}) when action in [:update] do
+      current_user.id == user.id
+    end
```
begedin: I'd like to point out that calling this method the elixir way would look like

``` Elixir
current_user |> can? :update, user
```

Which is just awesome!

joshsmith: Are you suggesting I change this? This function is from the canary docs.

joshsmith: No, you're not. I shouldn't read PR comments 90 seconds after waking up.

joshsmith:
@begedin I like everything I'm seeing here. Nothing I would change, other than getting the tests passing.

joshsmith:
Looks like tests are failing because of all the now expected authentication required. Wonder how we should fix in short-term?

joshsmith:
_Twenty twenty twenty-four failures to gooooo, I want them terminated!_

begedin:
@JoshSmith Everything is in policy modules now and the tests should pass. There's still cleanup left and some features we have are missing due to how canary works. I believe it's fixable, but perhaps more suited for separate issues.

I will add a list of todos into this PR, which we can either tackle here or in separate issues, as soon as I get the time. Need to run now.

joshsmith:
@begedin add and create whatever related issues you can when you have time.

begedin:
TODO:
- [x] Make it so validation for organization memberships only allows a "pending" role on create - #158 
- [x] Create issue (#176): Find a way to provide a changeset, or params to `load_and_authorize_resource`when authorising the `:create` action. We need it in some cases, because in order to decide if the user is allowed to perform the action, we need to see what type of resource is being created. It is needed for: 
  - `organization_membership` - User can only **create** their own pending membership and they should not be able to do it if they're already a member. The part of it being their own is handled by not allowing a `member_id` as a parameter. Instead, we just assign the current user. For **update**, we've simplified the permissions compared to rails, but ideally, we need to check what the current membership role is, and what we want it to be, then we want to decide if the user can do it based on their own role.
  - `posts` - Any user, regardless of membership, can create ideas and issues. Other posts can only be created by contributors or higher.
  - `project, project_category, project_skill` - We need to pass in the changeset so we can know which organization the record is being created for, since only some roles can do it.
  - `user_category, user_role, user_skill` - These might actually be fine. We need to check to see if the user is modifying their own record on create. We could handle that without permissions, by doing it on the changeset.
- [x] Issue exists: Some of these permissions have a "true if user is admin" clause, while some do not. Need to probably add that case for all permissions - #59 
- [x] Create Issue: Policies absolutely need tests - #174 
- [x] Create issue to discuss: There are some policies we have in rails, which we do not in phoenix, such as `post_image`, `comment_image`, etc. There are no endpoints for those, so I don't think we need them. - #175

joshsmith:
@begedin this was merged.

begedin:
Didn't realize. Sorry, still early. Either way, I'll finalize the list here and turn it into issues.

PR #152, opened 2016-08-26
Change index filters across Models

npendery:
Update the way we filter when hitting the index action for models

### DIFF
--- a/test/controllers/skill_controller_test.exs
+++ b/test/controllers/skill_controller_test.exs
@@ -33,10 +33,12 @@ defmodule CodeCorps.SkillControllerTest do
     elixir = insert(:skill, title: "Elixir")
     phoenix = insert(:skill, title: "Phoenix")
     insert(:skill, title: "Rails")
-    conn = get conn, "skills/?filter[id]=#{elixir.id},#{phoenix.id}"
-    data = json_response(conn, 200)["data"]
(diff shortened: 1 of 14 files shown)


thomascchen:
👍

test/controllers/skill_controller_test.exs
```diff
@@ -33,10 +33,12 @@ defmodule CodeCorps.SkillControllerTest do
     elixir = insert(:skill, title: "Elixir")
     phoenix = insert(:skill, title: "Phoenix")
     insert(:skill, title: "Rails")
-    conn = get conn, "skills/?filter[id]=#{elixir.id},#{phoenix.id}"
-    data = json_response(conn, 200)["data"]
+    params = %{"filter" => %{"id" => "#{elixir.id},#{phoenix.id}"}}
+    path = conn |> skill_path(:index, params)
+    data = conn |> get(URI.decode(path)) |> json_response(200) |> Map.get("data")
```
HeroicEric, line 38: I think this could also be done without the `URI.decode`ing, like:

``` elixir
params = %{"filter" => %{"id" => "#{elixir.id},#{phoenix.id}"}}
conn = get conn, skill_path(conn, :index, params)
data = json_response(conn, 200)["data"]
```

PR #156, opened 2016-08-29
Fix project changesets

### LINKED ISSUE #93: Project endpoint allows changing the organization through the update call
This should not be happening. A project should not be reassignable once it exists.

begedin:
Fixes #93 and adds some stuff on top
- add update_changeset which does not allow organization id
- add specs for slugs and update/create changesets


joshsmith:
👍

PR #160, opened 2016-08-30
Add state transitions to user

venkatd:


lib/code_corps/onboarding/user_transition.ex
```diff
@@ -0,0 +1,10 @@
+defmodule CodeCorps.Onboarding.UserTransition do
```
joshsmith: Wonder if `CodeCorps.Transition.UserState` is a better name?

Then we can also have a `Transition.PostStatus` or `PostStatus`.

web/models/user.ex
```diff
@@ -28,6 +28,9 @@ defmodule CodeCorps.User do
     field :username, :string
     field :website, :string
 
+    field :state, :string
```
joshsmith, line 31: Maybe add a `default` of `signed_up` here.

web/models/user.ex
```diff
@@ -78,11 +81,22 @@ defmodule CodeCorps.User do
   def update_changeset(struct, params) do
     struct
     |> changeset(params)
-    |> cast(params, [:first_name, :last_name, :twitter, :biography, :website, :base64_photo_data])
+    |> cast(params, [:first_name, :last_name, :twitter, :biography, :website, :base64_photo_data, :state_transition])
```
joshsmith, line 84: Think maybe we should `validate_inclusion` on the `state_transition` so the developer can track down the error. Even though `apply_state_transition()` does add one, that might be worthwhile to do.

web/models/user.ex
joshsmith, line 31: Oh I see that was there in the migration. Idk if you put that in `field`. My bad.

lib/code_corps/onboarding/user_transition.ex
```diff
@@ -0,0 +1,10 @@
...
+
+  def next(current, nil), do: nil
+  def next("signed_up", "edit_profile"), do: {:ok, "edited_profile"}
+  def next("edited_profile", "select_categories"), do: {:ok, "selected_categories"}
+  def next("selected_categories", "select_roles"), do: {:ok, "selected_roles"}
+  def next("selected_roles", "select_skills"), do: {:ok, "selected_skills"}
+  def next(current, invalid_transition), do: {:error, "invalid transition #{invalid_transition} from #{current}"}
```
joshsmith: Do we have a decent way to unit test these?

test/controllers/user_controller_test.exs
```diff
@@ -207,6 +207,35 @@ defmodule CodeCorps.UserControllerTest do
...
+      conn = put authenticate(conn, user), user_path(conn, :update, user), %{
+        "data" => %{
+          "type" => "user",
+          "id" => user.id,
+          "attributes" => %{"password" => "password", "state_transition" => "yehaww"}
+        }
+      }
+
+      %{"errors" => errors} = json_response(conn, 422)
+      assert errors["state_transition"] == ["invalid transition yehaww from signed_up"]
```
joshsmith, line 237: You always did have the best fake data.

lib/code_corps/onboarding/user_transition.ex
venkatd: Yeah that's a better name.

lib/code_corps/onboarding/user_transition.ex
venkatd: It should be as simple as testing the return value of `next` a bunch of times. I personally don't think there's a lot of value in these sorts of tests (code is very simple and you have to maintain an additional test file). If someone else wants to take them on at a later time I can create an issue for it.

test/controllers/user_controller_test.exs
venkatd: This line took me the longest to write.

web/models/user.ex
venkatd: The check seems redundant to me but I went ahead and added it anyway.

What's your reasoning behind it? If someone gets the transition name wrong, won't they go to the transition file?

joshsmith: I mean if an Ember dev gets it wrong, just seems a little more explicit maybe?

joshsmith: Wait, I missed that the error is added to `state_transition`, so this _is_ redundant. Sorry about that.

PR #164, opened 2016-08-31
Fix unused variable warning

HeroicEric:
Since the `user` variable is never used, the compiler throws a warning.
The `_` signifies that we don't care about the value of the variable so
the compiler will ignore it.

### DIFF
--- a/web/policies/organization_membership_policy.ex
+++ b/web/policies/organization_membership_policy.ex
@@ -52,7 +52,7 @@ defmodule CodeCorps.OrganizationMembershipPolicy do
     Organization
     |> Repo.get(membership.organization_id)
   end
-  defp fetch_membership(user, nil), do: nil
+  defp fetch_membership(_user, nil), do: nil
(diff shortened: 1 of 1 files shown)


web/policies/organization_membership_policy.ex
```diff
@@ -52,7 +52,7 @@ defmodule CodeCorps.OrganizationMembershipPolicy do
     Organization
     |> Repo.get(membership.organization_id)
   end
-  defp fetch_membership(user, nil), do: nil
+  defp fetch_membership(_, nil), do: nil
```
joshsmith: Would consider doing `_user` instead, just for clarity's sake. Will have same benefit but explicit about what the variable would be.

joshsmith:
👍  if you agree with suggestion.

joshsmith:
@HeroicEric should have mentioned, but better not to use merge and instead rebase off `develop`.

joshsmith:
You might want to rebase now and squash those commits together, then push up.

Issue #85, opened 2016-08-18
Add in pagination to all endpoints that need it


joshsmith:
We're missing pagination from a number of endpoints, namely the posts. We should identify the others, and then add pagination to those requests.

joshsmith:
Belongs to #26.

joshsmith:
See this possibility https://github.com/AgilionApps/ja_serializer#pagination

noslouch:
adding pagination to other endpoints is relatively straightforward. As you can see there's more ceremony around seeding the db in dev and test environments than there is in actually adding pagination support to the controller.

snewcomer:
@JoshSmith what other endpoints need it?  I can take one.

joshsmith:
Closed by #165.

PR #165, opened 2016-08-31
Paginate Posts

noslouch:
For #85. 

Adds two additional query param inputs to the `PostController` index action: `page[page]` and `page[page-size]`. `page[page]` defaults to 1 and `page[page-size]` defaults to 10 if not provided.

This slightly tweaks the pagination inputs from what the ember adapter is currently providing.
For the ember app to comply with this implementation, the [post adapter will need to update the name of the key in the page dict at which it sets the desired page number](https://github.com/code-corps/code-corps-ember/blob/develop/app/adapters/post.js#L14) from `number` to `page`. https://github.com/code-corps/code-corps-ember/pull/363

So given a query string that look like this:

```
?page[number]=2&status=open
------^ needs to change
```

changes to

```
?page[page]=2&status=open
```

And that should be it.

For reference, we're leveraging [Scrivener.Ecto](https://github.com/drewolson/scrivener_ecto) which is just delightful and also [ja_serializer](https://github.com/AgilionApps/ja_serializer#pagination) has out-of-the-box integration.

### DIFF
--- a/priv/repo/seeds.exs
+++ b/priv/repo/seeds.exs
@@ -5,6 +5,7 @@ alias CodeCorps.Project
 alias CodeCorps.Role
 alias CodeCorps.Skill
 alias CodeCorps.User
+alias CodeCorps.Post
 
 # Users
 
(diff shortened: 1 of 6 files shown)


priv/repo/seeds.exs
```diff
@@ -30,7 +31,7 @@ cond do
     Enum.each(users, fn user ->
       %User{}
       |> User.registration_changeset(user)
-      |> Repo.insert!()
+      |> Repo.insert!
```
noslouch, line 34: this was just an un-elixir thing I saw

joshsmith:
Looks, great @noslouch! Thanks so much. 👍 

Rebase and merge if you would.

PR #167, opened 2016-08-31
Removed unecessary function match from post_policy

begedin:
While we do lack tests for all policies, this is a quick fix to remove a function that will always match and prevent the proper function from being called.


joshsmith:
:+1:

PR #168, opened 2016-08-31
Post state and status management

### LINKED ISSUE #83: Add state to posts
We don't have `state` in posts. This is not a super high priority right now, but we should find a way to add this.

I couldn't get this working with the Ecto state machine project because I kept getting weird errors about `cast/3`. Was very odd.

I still think it may actually be simpler to take this approach:

> We could replace `aasm` transitions from Rails with changeset pattern matching, something like: `def transition_changeset(%Model{state: "from_this_state"} = model, "to_this_state")` And then make a catchall that logs, complains, or just simply ignores transitions that we don't want. Probably want to pass in a changeset and return that so we can pipe with other changesets in the model.

begedin:
Closes #83, #169

I truly believe a state machine is a complete overkill for this one. 

There are only two cases for state:
- When we are creating the post, it comes out as "published"
- When we are updating it, it becomes "edited"

Due to that, manually putting those two changes into the respective changesets is more than enough.

Also, there are only two cases for status:
- When we are creating the post, it comes out as "open".
- During update, we can set it to "open" or "closed" and nothing else.

While I was at it, I reorganized the post model tests. We are lacking in that area and we should write more, but it's outside of this PRs scope.

### DIFF
--- a/test/models/post_test.exs
+++ b/test/models/post_test.exs
@@ -12,7 +12,7 @@ defmodule CodeCorps.PostTest do
     post_type: "nonexistent"
   }
 
-  describe "&create/2" do
+  describe "create/2" do
     test "is invalid with invalid attributes" do
       changeset = Post.changeset(%Post{}, @invalid_attrs)
       refute changeset.valid?
@@ -38,7 +38,7 @@ defmodule CodeCorps.PostTest do
     end
   end
 
-  describe "&create_changeset/2" do
+  describe "create_changeset/2" do
     test "is valid with valid attributes" do
       user = insert(:user)
       project = insert(:project)
@@ -92,7 +92,7 @@ defmodule CodeCorps.PostTest do
       assert changeset |> get_field(:status) == "open"
     end
   end
-  describe "&update_changeset/2" do
+  describe "update_changeset/2" do
     test "sets state to 'edited'" do
       changeset = Post.update_changeset(%Post{}, %{})
       assert changeset |> get_change(:state) == "edited"
--- a/test/models/post_test.exs
+++ b/test/models/post_test.exs
@@ -7,11 +7,9 @@ defmodule CodeCorps.PostTest do
     title: "Test post",
     post_type: "issue",
     markdown: "A test post",
-    status: "open"
   }
   @invalid_attrs %{
-    post_type: "nonexistent",
-    status: "nonexistent"
+    post_type: "nonexistent"
   }
 
   describe "&create/2" do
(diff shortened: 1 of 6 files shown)


web/models/post.ex
```diff
@@ -10,6 +10,7 @@ defmodule CodeCorps.Post do
     field :markdown, :string
     field :number, :integer
     field :post_type, :string
+    field :state, :string
```
begedin, line 13: It's already present in the schema, as part of the initial migration

test/models/post_test.exs
```diff
@@ -14,76 +14,98 @@ defmodule CodeCorps.PostTest do
...
+    test "is valid with valid attributes" do
+      user = insert(:user)
+      project = insert(:project)
+      changeset = Post.create_changeset(%Post{}, %{
+        markdown: "some content",
+        post_type: "issue",
+        title: "some content",
+        project_id: project.id,
+        user_id: user.id,
+        status: "open"
```
joshsmith: Do we add this status ourselves from the client?

test/models/post_test.exs
```diff
@@ -14,76 +14,98 @@ defmodule CodeCorps.PostTest do
...
-    changes = Map.put(@valid_attrs, :post_type, "nonexistent")
-    changeset = Post.changeset(%Post{}, changes)
-    refute changeset.valid?
+    test "sets status to 'open'" do
+      changeset = Post.create_changeset(%Post{}, %{})
+      # open is default, so we `get_field` instead of `get_change`
+      assert changeset |> get_field(:status) == "open"
+    end
   end
+  describe "&update_changeset/2" do
```
joshsmith: Is the `&` the elixir way of describing this?

test/models/post_test.exs
begedin: Need to cleanup the attributes. The status can be provided, but the create changeset ignores the param and simply sets to "open"

joshsmith:
Does the current front-end rely on an `edited_at` field or was that the API going out ahead of the client's ability to consume and do anything with in the UI? I know we have something of that nature in Rails, but not sure if it has any downstream impact.

test/models/post_test.exs
joshsmith: This is only thing I would really suggest to change here since it's misleading.

test/models/post_test.exs
begedin: Actually, it's `function_name/arity`, sans `&`. I've corrected it

begedin:
@JoshSmith `edited_at` has no purpose yet on the client, nor does `updated_at`, which rails had.

joshsmith:
@begedin what I thought, then ignore it and we'll be sure to add to UI later. It's a worthwhile feature but non-blocking.

joshsmith:
Rebase, squash, and merge at will.

joshsmith:
👍

PR #172, opened 2016-08-31
Changes for ember compatibility

begedin:
What was done
- Changed the error format for validation errors to conform JSON API specs and guidelines. Ember expect that specific format and handles it automatically, so the alternative of customising error handling client-side would take longer.
- Made it so, when creating a post, the controller reloads the record before responding to the request, reason being, the post number is being added at the database level, so we need to reload to get that information back.
- `CommentController#index` was lacking the "filter by ID functionality", so I added it and specs to support the feature.
- `PostController.index` was missing filtering by status, so I added that as well
- Made it so slug only regenerates if the source value it's generated from changes. Simplifies some things to a degree.
- `OrganizationMembershipControler#index` was filtering by `member_id` instead of `id`. Fixed that bug.
- `CategoryController#index` was missing some preloads
- `User` was missing the default "signed_up" state upon creation, so the onboarding process would never trigger.
- Some authorization policies needed a changeset to properly determine user permissions, so I added custom authorization functionality, which accepts a changeset, namely, policies for `UserCategory`, `UserRole` and `UserSkill`

### DIFF
--- a/web/controllers/user_category_controller.ex
+++ b/web/controllers/user_category_controller.ex
@@ -4,7 +4,7 @@ defmodule CodeCorps.UserCategoryController do
   alias CodeCorps.UserCategory
   alias JaSerializer.Params
 
-  plug :load_and_authorize_resource, model: UserCategory, only: [:create, :delete]
+  plug :load_and_authorize_resource, model: UserCategory, only: [:delete]
   plug :scrub_params, "data" when action in [:create]
 
   def index(conn, params) do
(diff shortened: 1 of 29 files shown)


web/controllers/user_category_controller.ex
```diff
@@ -45,4 +51,11 @@ defmodule CodeCorps.UserCategoryController do
     UserCategory |> Repo.get!(id) |> Repo.delete!
     conn |> send_resp(:no_content, "")
   end
+
+  defp authorize(conn, changeset) do
+    conn
+    |> assign(:changeset, changeset)
+    |> Canary.Plugs.authorize_resource(model: changeset)
+  end
+  defp authorized?(conn), do: conn |> Map.get(:assigns) |> Map.get(:authorized)
```
npendery: Should these two methods be added to controller helpers since they are used in multiple controllers?

npendery:
Looks good! Just had one comment. Will look over again though later to make sure I didnt miss anything

web/controllers/user_category_controller.ex
begedin: I'm not really sure. Maybe a brand new `authorization_helpers` would make more sense.

begedin:
@npendery I've moved the two functions into `CodeCorps.AuthenticationHelpers`. It makes the most sense, really, since we already have related functions there.

PR #173, opened 2016-08-31
Automatically set pending role on membership creation. Allow update of role

### LINKED ISSUE #158: Only a "pending" value should be allowed for the role field for organization memberships
For creation of an organization membership, only a "pending" value should be allowed for the role field. This should be done trough the changeset, not permissions.

See #149.

begedin:
Closes #158 

Had to rewrite the model specs, due to updated changesets
- Now, `create_changeset` sets new membership role to "pending" automatically.
- `update_changeset` allows setting of membership role to one of predefined values only


joshsmith:
👍 looks great! 🚢

PR #177, opened 2016-08-31
Update INSTALLING.md to include copy of .env.example

monkseal:
Small change for problem I noticed while installed.


monkseal:
@JoshSmith @begedin Can we update [USAGE.md](https://github.com/code-corps/code-corps-phoenix/blob/develop/docs/USAGE.md) to include the command to run the tests? Do you need to use `docker exec`? 

When I try `mix test test/models/user_test.exs`, I get:

```
12:19:05.036 [error] GenServer #PID<0.208.0> terminating
** (Postgrex.Error) FATAL (invalid_authorization_specification): role "postgres" does not exist
    (db_connection) lib/db_connection/connection.ex:148: DBConnection.Connection.connect/2
    (connection) lib/connection.ex:623: Connection.enter_connect/5
    (stdlib) proc_lib.erl:247: :proc_lib.init_p_do_apply/3
```

joshsmith:
@monkseal we can do that! Test running is `docker-compose run test mix test`.

PR #186, opened 2016-09-11
Update Docker

joshsmith:
This makes some changes to Docker (removing unnecessary portions of Brunch, ecto migrations, etc).


begedin:
👍

PR #187, opened 2016-09-12
Add cat field to Role Skill

### LINKED ISSUE #129: Add cat enum field to RoleSkill
We're missing the cat enum field for `RoleSkill`.

For #26.

npendery:
Closes #129


web/models/role_skill.ex
```diff
@@ -15,8 +17,9 @@ defmodule CodeCorps.RoleSkill do
   """
   def changeset(struct, params \\ %{}) do
     struct
-    |> cast(params, [:role_id, :skill_id])
+    |> cast(params, [:cat, :role_id, :skill_id])
     |> validate_required([:role_id, :skill_id])
+    |> validate_inclusion(:cat, cats)
```
joshsmith, line 22: I would put the `cast` and validation into an `import_changeset` instead. We don't do imports yet, but that would be the ideal place for this.

I added the same migration in my not-yet-merged branch, so I can remove that when this is merged.

joshsmith:
Looks like a bunch of random commits got thrown here. May need to rebase.

npendery:
@JoshSmith yeah I saw that, working on that now. Some funky stuff happened when I rebased from origin so my local develop branch is off.

joshsmith:
It may be less effort to close and copy/paste raw files, depending on how crazy git history is.

npendery:
Yeah thats what I was thinking too, doing that now

PR #188, opened 2016-09-12
Migrate more data for deployment

joshsmith:
This changes `preview` to `previews` (an old migration).

Removes unnecessary icon URL fields.

### DIFF
--- a/priv/repo/migrations/20160815125009_create_preview.exs
+++ b/priv/repo/migrations/20160815125009_create_preview.exs
@@ -2,7 +2,7 @@ defmodule CodeCorps.Repo.Migrations.CreatePreview do
   use Ecto.Migration
 
   def change do
-    create table(:preview) do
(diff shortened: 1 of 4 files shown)


npendery:
I dont think `text` is a valid type in Ecto which is why the build fails. Here are the [valid types](https://hexdocs.pm/ecto/Ecto.Schema.html#module-types-and-casting)

joshsmith:
@npendery great minds...

priv/repo/migrations/20160911232200_add_cat_to_role_skills.exs
```diff
@@ -0,0 +1,9 @@
+defmodule CodeCorps.Repo.Migrations.AddCatToRoleSkills do
```
npendery: This was added in #189, should remove in this PR.

PR #190, opened 2016-09-12
Prep for Heroku and Circle

joshsmith:
Adds some config information for both Heroku and Circle.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,25 @@
+machine:
+  environment:
+    PATH: "$HOME/.asdf/bin:$HOME/.asdf/shims:$PATH"
+
+dependencies:
+  cache_directories:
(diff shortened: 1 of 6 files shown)


circle.yml
```diff
@@ -0,0 +1,25 @@
...
+test:
+  override:
+    - mix test
+
+deployment:
+  staging:
+    branch: develop
+    commands:
+      - git push git@heroku.com:code-corps-phoenix-staging.git $CIRCLE_SHA1:master
+      - heroku run rake db:migrate --app code-corps-phoenix-staging
```
schneidmaster: you probably want `mix ecto.migrate`

joshsmith: Nah I want Rails.

PR #196, opened 2016-09-13
Change error view

joshsmith:
Replaces `404` and `500` with proper`.json-api` views.


snewcomer:
@JoshSmith Looks good to me!  I believe the Ja.ErrorSerializer is also an option, but probably not a big deal - 

https://github.com/AgilionApps/ja_serializer/search?utf8=%E2%9C%93&q=ErrorSerializer

joshsmith:
@snewcomer can you add an issue to refactor to use that? Looks good to use.

PR #206, opened 2016-09-13
Missing command fix in INSTALLING.md

rileytaylor:
The `cp .env.example .env` command was missing the `.env` part.


npendery:
Good catch! 🚀

PR #207, opened 2016-09-14
Add analytics

### LINKED ISSUE #182: Add analytics reporting to Segment
We need to report some basic funnel analytics up to Segment like we did in the Rails API.

Part of #26.

joshsmith:
Not yet finished.

Closes #182.

This is a first spike to get analytics working.

You'll notice I needed to change the pipelines to get the `current_user` anywhere the JTW might be present, and not just in routes that require authentication and halt on `401`.

I also have our own `CodeCorps.Analytics.Segment` which calls out Segment, vs `.InMemory` which can be used in `test` and maybe even `dev` to instead be testable and not make an HTTP request unnecessarily.

We need to add the following events:
- [x] Added User Category
- [x] Added User Role
- [x] Added User Skill
- [x] Removed User Category
- [x] Removed User Role
- [x] Removed User Skill
- [x] Updated Profile
- [x] Signed In
- [x] Signed Up
- [x] Created Comment
- [x] Created Post
- [x] Edited Comment
- [x] Edited Post
- [x] Created Organization Membership
- [x] Requested Organization Membership


green-arrow:
This looks good to me. I would suggest using the in memory one in dev as well and only using the real version in production :)

joshsmith:
@green-arrow I would agree there and then just comment that out when I do want to default to Segment.

joshsmith:
Should consider if we want to track most of these events in the controller (probably?). Should happen after the event itself succeeds, in any case.

joshsmith:
For the `.track` functions, we should use pattern matching off something like a string of the event name, for example `signed_up`, or maybe an atom? We can then write the `.InMemory` version to just have a catchall rather than matching them individually.

We can also then ensure the correct data is passed as the second argument, like a `%User{}` struct in the case of `signed_up`.

Any thoughts?

lib/code_corps/analytics/segment.ex
```diff
@@ -0,0 +1,24 @@
+defmodule CodeCorps.Analytics.Segment do
+  def identify(user) do
+    %Segment.Analytics.Identify
+      {
+        userId: user.id,
```
begedin: Is there a special reason `userId` has to be camelCase?

web/plugs/current_user.ex
```diff
@@ -1,20 +1,22 @@
...
-            conn
-        end
-      {:error, _reason} ->
+    case Guardian.Plug.current_token(conn) do
+      nil ->
         conn
+      current_token ->
+        with {:ok, claims} <- Guardian.decode_and_verify(current_token),
+             {:ok, user} <- GuardianSerializer.from_token(claims["sub"]) do
+                @analytics.identify(user)
```
begedin: Here ought to be good enough, but I'm thinking making a pipeline and a plug specifically for identifying the current user might be a bit clearer (at the cost of some performance. 

It could fetch the user from the `conn`, since the `current_user` plug assigns it. No need for a database fetch or anything like that.

web/router.ex
```diff
@@ -64,7 +70,7 @@ defmodule CodeCorps.Router do
   end
 
   scope "/", CodeCorps, host: "api." do
-    pipe_through [:api, :with_token]
+    pipe_through [:api, :bearer_auth, :ensure_auth, :current_user]
```
begedin: Considering what we had above, we would 

```
pipe_through [:api, :bearer_auth, :ensure_auth, :current_user, :analytics_identify]
```

begedin (commented):
The nature of analytics sort of goes against how elixir works (function chaining), so I'm really not sure what the best option would be. I have a suggestion, though, so feel free to consider it.

lib/code_corps/analytics/segment.ex
joshsmith: Yeah, that's how the library expects it.

web/plugs/current_user.ex
joshsmith: This seems like a much better idea than what I'm doing. Will implement.

joshsmith:
I think we should add in testing of this at a later step. Don't care enough now for it to make it into this PR.

config/dev.exs
```diff
@@ -45,3 +45,5 @@ config :code_corps, allowed_origins: ["http://localhost:4200"]
 
 config :guardian, Guardian,
   secret_key: "e62fb6e2746f6b1bf8b5b735ba816c2eae1d5d76e64f18f3fc647e308b0c159e"
+
+# config :code_corps, :analytics, CodeCorps.Analytics.InMemory
```
begedin: Maybe switch to in_memory, if segment keys are missing from env?

begedin (approved):
Looks good. Not sure about the segment thing I mentioned. Just an idea.

green-arrow (commented):
This looks good to me :)

config/dev.exs
joshsmith: Yeah, can you add an issue for doing this later?

PR #208, opened 2016-09-14
Limit Skill controller index to return only 5 skills

### LINKED ISSUE #205: Add limit to skills autocomplete query
We should add a `limit` of 5 to the Skills autocomplete query until we can focus on #60 to do this feature justic.

npendery:
Closes #205

### DIFF
--- a/test/controllers/skill_controller_test.exs
+++ b/test/controllers/skill_controller_test.exs
@@ -56,6 +56,17 @@ defmodule CodeCorps.SkillControllerTest do
       assert first_result["id"] == "#{ruby.id}"
       assert second_result["id"] == "#{rails.id}"
     end
+
+    test "limit filter limits results on index", %{conn: conn} do
+      insert_list(6, :skill)
+
+      params = %{"limit" => 5}
+      path = conn |> skill_path(:index, params)
+      json = conn |> get(path) |> json_response(200)
+
+      returned_skills_length = json["data"] |> length
+      assert returned_skills_length == 5
+    end
   end
 
   describe "show" do
--- a/web/models/model_helpers.ex
+++ b/web/models/model_helpers.ex
@@ -25,6 +25,11 @@ defmodule CodeCorps.ModelHelpers do
 
   def newest_first_filter(query), do: query |> order_by([desc: :inserted_at])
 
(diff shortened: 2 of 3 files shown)


test/controllers/skill_controller_test.exs
```diff
@@ -15,11 +15,13 @@ defmodule CodeCorps.SkillControllerTest do
...
   describe "index" do
-    test "lists all entries on index", %{conn: conn} do
+    test "lists first five entries on index", %{conn: conn} do
+      insert_list(6, :skill)
       path = conn |> skill_path(:index)
       json = conn |> get(path) |> json_response(200)
 
-      assert json["data"] == []
+      returned_skills_length = json["data"] |> length
+      assert returned_skills_length == 5
```
joshsmith: I think we should be doing this `limit` only in the search case, so the `|> limit(5)` I think should be folded into that.

joshsmith: We'll need to test here that it renders all of them and test in the search case that it renders no more than 5.

npendery: Yeah, I was thinking about that but couldnt figure out how to discern between the two with only backend. This might be better tackled on the Ember side to just display 5 after it requests all matching skills?

joshsmith: I think instead of relying on `title_filter` in the `ModelHelpers` I'd make a function specifically on `Skill` that includes the `limit`. Anytime you're doing a `?query` on that endpoint you should only be returning five.

joshsmith: Either that or we should explicitly add a `limit` filter and then the Ember app can add that to the query params.

joshsmith: The latter seems more flexible now that I think about it, but we could always do the former temporarily, open an issue here and the Ember side, and circle back later if we'd like.

begedin:
Just a suggestion to consider, but how about creating a limit filter, which has a default of 5 for now? We can then later, in ember, explicitly specify a different value if necessary.

npendery:
Yeah I like the idea of the limit filter. Will also make the necessary changes on the Ember side.

joshsmith (approved):

PR #212, opened 2016-09-15
Current user based index data

### LINKED ISSUE #204: Fix user-skills and user-categories index actions to return authenticated user data
Right now the `index` actions in both `user-skills` and `user-categories` return everyone's data, instead of the authenticated user's when authenticated.

In Rails we did the following:

```
    if id_params.present?
      user_skills = UserSkill.includes(:skill, :user).where(id: id_params)
    elsif current_user.present?
      user_skills = UserSkill.includes(:skill, :user).where(user_id: current_user.id)
    else
      user_skills = []
    end
```

We need to fix this the `filter` or find some other approach to correcting the data returned from this endpoint.

npendery:
Closes #204 

On top of @JoshSmith's #207

### DIFF
--- a/web/controllers/user_skill_controller.ex
+++ b/web/controllers/user_skill_controller.ex
@@ -10,6 +10,14 @@ defmodule CodeCorps.UserSkillController do
   plug :scrub_params, "data" when action in [:create]
 
   def index(conn, params) do
+    params =
+      case conn.assigns[:current_user] do
+        nil ->
+          params
+        user ->
+          Map.put(params, :current_user_id, user.id)
+      end
+
     user_skills =
       UserSkill
       |> UserSkill.index_filters(params)
--- a/web/models/model_helpers.ex
+++ b/web/models/model_helpers.ex
@@ -17,6 +17,11 @@ defmodule CodeCorps.ModelHelpers do
 
   # filters
 
+  def current_user_filter(query, %{current_user_id: user_id}) do
+    query |> where([object], object.user_id == ^user_id)
+  end
+  def current_user_filter(query, _), do: query
+
   def id_filter(query, %{"filter" => %{"id" => id_list}}) do
     ids = id_list |> coalesce_id_string
     query |> where([object], object.id in ^ids)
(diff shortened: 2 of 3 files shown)


web/controllers/user_skill_controller.ex
```diff
@@ -10,6 +10,17 @@ defmodule CodeCorps.UserSkillController do
...
+      case conn.assigns[:current_user] do
+        nil ->
+          params
+        %CodeCorps.User{id: user_id} ->
+          Map.put(params, :current_user_id, user_id)
+        _ ->
+          params
+      end
+
+    Kernel.inspect(params) |> IO.puts
```
joshsmith: Is this supposed to be here?

web/plugs/current_user.ex
```diff
@@ -1,20 +1,22 @@
 defmodule CodeCorps.Plug.CurrentUser do
+  @analytics Application.get_env(:code_corps, :analytics)
```
JoshSmith: How did analytics code make it in here?

joshsmith (changes requested):
I think doing this on top of #207 wasn't a good idea, because you've had to undo changes in there. I would have just pulled in the auth-type changes I was making there, or do that as a separate PR and I could've rebased from there.

I would consider doing that now since I'm not sure what effect this will have on that existing PR and the commit history.

begedin:
@npendery 

Sorry about this, but it seems we'll have to close the PR without merging. We wanted to mirror the rails api here, but once we took another good look, we concluded it doesn't make sense to do it that way.

Here's what rails does:

``` Ruby
if id_params.present?
  user_skills = UserSkill.includes(:skill, :user).where(id: id_params)
elsif current_user.present?
  user_skills = UserSkill.includes(:skill, :user).where(user_id: current_user.id)
else
  user_skills = []
end
```

To put it in words, if the client provides a list of ids it wants, they get that back. If they don't, then if they are authenticated, they get all of their own records back.

The reality is, they shouldn't need to explicitly fetch their own records. They should get a list of ids with their user payload, and then simply fetch those records using these ids.

If there anything, we might eventually need a `user_filter`instead of a `current_user_filter`, meaning the client would provide a `user_id` to filter by. For now, though, we can close this and see if there's anything that needs fixing client-side.

PR #213, opened 2016-09-15
Hide email from response for all but authenticated user

### LINKED ISSUE #194: Prevent email from rendering for anyone other than the currently authed user
We need to prevent the email from rendering in the API for anyone other than the user who is currently authenticated.

Right now you can see anyone's email in the request, which is not good.

begedin:
Closes #194 

I implemented this at the view level.

We can override record attributes by defining functions of the same name as the attribute. This function then receives the record as the first parameter and the `conn` as the second, so it's simple to compare the user being serialised with the `current_user` assign in the `conn`. 

Took me a while to figure out this is possible, though.

### DIFF
--- a/web/views/user_view.ex
+++ b/web/views/user_view.ex
@@ -31,4 +31,19 @@ defmodule CodeCorps.UserView do
   def photo_thumb_url(user, _conn) do
     CodeCorps.UserPhoto.url({user.photo, user}, :thumb)
   end
+
+  @doc """
+  Returns the user email or an empty string, depending on the user
+  being rendered is the authenticated user, or some other user.
+
+  Users can only see their own emails. Everyone else's are private.
+  """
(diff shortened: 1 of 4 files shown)


web/views/user_view.ex
```diff
@@ -31,4 +31,12 @@ defmodule CodeCorps.UserView do
   def photo_thumb_url(user, _conn) do
     CodeCorps.UserPhoto.url({user.photo, user}, :thumb)
   end
+
+  def email(user, %Plug.Conn{assigns: %{current_user: current_user}}) do
```
joshsmith, line 41: I would add some `def` documentation of this.

joshsmith (changes requested):
Overall looks good. Would just add some brief documentation stating what's going on since it's non-obvious at first glance.

begedin:
@JoshSmith Need another review for this.

joshsmith (approved):

PR #215, opened 2016-09-15
Made searching by slug case insensitive

### LINKED ISSUE #203: Make find on slugs case-insensitive
Right now `code-corps` does not return the same result as `CODE-CORPS`. It should.

begedin:
Closes #203 

I could use ideas on how to organize this a bit better, but I think we might wanna do it in a separate issue, since a lot of current PRs touch on filtering.

I'm still using `Repo.get_by!`, but since the generated slug is always downcased (parametrized), I also downcase the filter.

I'm not really sure that's what we want, though. Do we want to parametrize the filter as well?

### DIFF
--- a/test/controllers/project_controller_test.exs
+++ b/test/controllers/project_controller_test.exs
@@ -61,6 +61,13 @@ defmodule CodeCorps.ProjectControllerTest do
 
       assert expected_ids == actual_ids
     end
+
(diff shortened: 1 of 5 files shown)


joshsmith:
I think simply downcasing the filter is fine here. Not sure parameterizing it is necessary.

joshsmith:
👍

joshsmith (approved):

PR #216, opened 2016-09-15
Several fixes to make onboarding and some other features work properly.

begedin:
- UserRoleController was missing index and show endpoints. I added those, but we still need to add tests
- Added routes to support above actions
- Added has_many user-skills to skill view, but I'm not absolutely sure we need that. The ember behavior on the skill onboarding page is weird. We should look into making it simpler.
- Fixed fitlering of emails for unauthenticated user. This was an issue introduced by me today. I was under the impression that conn will get a "current_user" assign when authenticated, but the current user is actually assigned to the "data" key. I changed the method signature in UserView, but we need to update associated tests and add a test on the user controller, best under the show group, to ensure an authenticated user will get an email. Apparently, the view test wasn't enough.
- **Very important** - Made it so our pipeline always loads the current user if a token is present. The endpoints that do explicitly require authentication should still support requests from authenticated users.
- [ ] Add tests for UserRoleController #index and #show. Use UserSkillController as example
- [ ] Look into making ember behavior on the skill onboarding page simpler
- [x] Update broken tests on the user view
- [x] Add test for UserController.show, where, when authenticated, the response has the email field properly populated.

### DIFF
--- a/web/views/skill_view.ex
+++ b/web/views/skill_view.ex
@@ -5,5 +5,4 @@ defmodule CodeCorps.SkillView do
   attributes [:title, :description, :inserted_at, :updated_at]
 
   has_many :role_skills, serializer: CodeCorps.RoleSkillView
-  has_many :roles, serializer: CodeCorps.RoleView
 end
--- a/test/controllers/user_controller_test.exs
+++ b/test/controllers/user_controller_test.exs
@@ -59,6 +59,14 @@ defmodule CodeCorps.UserControllerTest do
       assert data["attributes"]["password"] == nil
     end
 
+    test "renders email when authenticated", %{conn: conn} do
+      user = insert(:user)
+      path = conn |> user_path(:show, user)
+      json = conn |> authenticate(user) |> get(path) |> json_response(200)
+
+      assert json["data"]["attributes"]["email"] == user.email
+    end
+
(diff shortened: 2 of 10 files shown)


begedin:
What's next?

I think we should simplify the user view and not include the "through" relationships (skills, roles, categories). Instead, we should only include "user-skills", "user-roles", "user-categories". The client is more then capable to fetch the other stuff. Ember might need to be updated as well.

web/views/skill_view.ex
```diff
@@ -5,5 +5,5 @@ defmodule CodeCorps.SkillView do
   attributes [:title, :description, :inserted_at, :updated_at]
 
   has_many :role_skills, serializer: CodeCorps.RoleSkillView
-  has_many :roles, serializer: CodeCorps.RoleView
+  has_many :user_skills, serializer: CodeCorps.UserSkillView
```
joshsmith: Do we really want all the `user_skills` across the app loaded for each `Skill`? I wouldn't think so.

I can see adding it for a `User`, but not for a `Skill`.

joshsmith (changes requested):
Not sure how I feel about the changes made here.

joshsmith (approved):

joshsmith:
@begedin I see you commented that you weren't sure about including user skill in skill. I removed that. It's not necessary after all and appears to still work on Ember side.

PR #223, opened 2016-09-17
Add Sentry

joshsmith:
This adds Sentry error tracking.

Also includes some fixes for mime types and JSON API error rendering.

Needed to override the `hackney` dependency so not sure how to fix that, but would prefer not to be relying on it directly. Got into dependency hell without that move.


joshsmith:
This is blocked unless and until Sentry raises the errors or we can find an alternative way to do `assert_error_sent`.

joshsmith:
> If you could try Phoenix master with: `{:phoenix, "1.3.0-dev", github: "phoenixframework/phoenix", override: true}`, and let me know if everything is good, that would be great.

joshsmith:
Appears to be related to the currently failing tests: https://github.com/phoenixframework/phoenix/issues/1928

schneidmaster (approved):

PR #226, opened 2016-09-17
Change static assets to use priv/static and only robots.txt

joshsmith:
This removes `web/static` which was unused and changes to serving `priv/static` with the Heroku static buildpack. Serves only `robots.txt`.


WenInCode (approved):
LGTM

PR #227, opened 2016-09-17
Change repo name

joshsmith:
This changes the repo name everywhere it should matter.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -19,14 +19,14 @@ deployment:
   staging:
     branch: develop
     commands:
-      - git push git@heroku.com:code-corps-phoenix-staging.git $CIRCLE_SHA1:master
-      - heroku run "POOL_SIZE=2 mix ecto.migrate" --app code-corps-phoenix-staging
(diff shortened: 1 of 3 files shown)


circle.yml
```diff
@@ -19,14 +19,14 @@ deployment:
   staging:
     branch: develop
     commands:
-      - git push git@heroku.com:code-corps-phoenix-staging.git $CIRCLE_SHA1:master
-      - heroku run "POOL_SIZE=2 mix ecto.migrate" --app code-corps-phoenix-staging
+      - git push git@heroku.com:code-corps-staging.git $CIRCLE_SHA1:master
+      - heroku run "POOL_SIZE=2 mix ecto.migrate" --app code-corps-staging
```
WenInCode, line 23: I do not know how our heroku stuff is set up. I just want to confirm this is supposed to be `code-corps-staging` and not something like `code-corps-api-staging`

joshsmith: This is correct. I opted for not having `api` there since we're only ever serving the API through Heroku.

WenInCode (approved):
LGTM

PR #228, opened 2016-09-17
Add color icons

joshsmith:
This adds some pretty colors to the icons.


joshsmith:
Need to fix tests, but as I've not yet eaten today, this will wait. 🍔

PR #231, opened 2016-09-18
Redirect to API docs from root and remove warning for CORS from tests

### LINKED ISSUE #198: Remove Phoenix built in PageController
We need to get rid of the page that renders for the main api page: https://api.pbqrpbecf.org/

[Router endpoint](https://github.com/code-corps/code-corps-phoenix/blob/develop/web/router.ex#L27)

joshsmith:
Closes #198.

@noslouch instead of removing entirely or replacing with some random root response, I figured it would be a good opportunity to redirect from the root to our API docs.

Thoughts?

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -61,6 +61,9 @@ config :segment,
 # Configures random icon color generator
 config :code_corps, :icon_color_generator, CodeCorps.RandomIconColor.Generator
 
+# Set Corsica logging to output a console warning when rejecting a request
+config :code_corps, :corsica_log_level, [rejected: :warn]
+
 # Import environment specific config. This must remain at the bottom
 # of this file so it overrides the configuration defined above.
 import_config "#{Mix.env}.exs"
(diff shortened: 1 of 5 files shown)


config/config.exs
```diff
@@ -61,6 +61,8 @@ config :segment,
 # Configures random icon color generator
 config :code_corps, :icon_color_generator, CodeCorps.RandomIconColor.Generator
 
+config :code_corps, :corsica_log_level, [rejected: :warn]
```
begedin, line 65: Might not be a bad idea to add a short comment to this, explaining what it does. Might be a good idea with any config setting. Maybe something like 

```
# Set Corsica logging to output a console warning when rejecting a request
# https://hexdocs.pm/corsica/Corsica.html
```

Does't cost much initially, but saves a decent amount of googling for future developers.

begedin (commented):
Looks good, but didn't we use to redirect to the ember app from any route not explicitly listed in the router?

joshsmith:
@begedin in the Rails app we actually served up the Ember app from the app, we didn't just redirect.

PR #232, opened 2016-09-18
Fix mixed case slugs and ensure slugs are properly generated for slugged routes

joshsmith:
This fixes the bug with existing mixed case slugged routes by properly parameterizing them. Also fixes it going forward.

### DIFF
--- a/priv/repo/migrations/20160918003206_fix_mixed_case_slugs.exs
+++ b/priv/repo/migrations/20160918003206_fix_mixed_case_slugs.exs
@@ -1,0 +1,24 @@
+defmodule CodeCorps.Repo.Migrations.FixMixedCaseSlugs do
+  use Ecto.Migration
+
+  alias CodeCorps.Repo
+  alias CodeCorps.SluggedRoute
+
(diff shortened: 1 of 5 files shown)


priv/repo/migrations/20160918003206_fix_mixed_case_slugs.exs
```diff
@@ -0,0 +1,24 @@
...
+    SluggedRoute
+    |> Repo.all
+    |> Repo.preload([:user, :organization])
+    |> Enum.each(fn record ->
+      changeset =
+        SluggedRoute.changeset(record)
+        |> Ecto.Changeset.put_change(:slug, Inflex.parameterize(record.slug))
+        |> Repo.update!
+    end)
+  end
```
begedin, line 19: Could do a separate method here, then call it like 

``` Elixir
|>Enum.each(&parametrize_slug(&1)
```

I don't think that's really necessary in a migration, though.

begedin (approved):
Looks good to me. Minor comment, but feel free to decide it's worth it. I'm not sure it is for a migration.

PR #233, opened 2016-09-18
Update API docs

### LINKED ISSUE #67: Rewrite documentation to account for API changes
We're going to need to change our API blueprint documentation to account for the API changes that have been incurred during this migration to Phoenix.

joshsmith:
Closes #67 when finished.

Still in progress.
- Changes OAuth token to JWT
- Removes some unused routes
- Changes URLs from underscores to dashes
- Changes type to singularized
- Removes owner type for slugged route and adds proper relationships (may need to include an empty user/organization relationship)
- Removes reference to comment images
- Changes to `application/json` which appears to be what Ember sends

### DIFF
--- a/blueprint/api.apib
+++ b/blueprint/api.apib
@@ -3,7 +3,7 @@ HOST: https://api.codecorps.org
 
 # API Reference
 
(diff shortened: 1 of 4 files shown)


rileytaylor (approved):
Looks good, :shipit:

PR #254, opened 2016-09-25
Issue 145 add full name

schippada:
Tested the api from the browser. I updated both test and api documentation. But I was not able to run the tests. Was getting the following error. Please review the code and let me know. I will worry about running the test later if that is okay.

Here is the error I am getting when I run the tests:
could not compile dependency :arc, "mix compile" failed. You can recompile this dependency with "mix deps.compile arc", update it with "mix deps.update arc" or clean it with "mix deps.clean arc"
==> code_corps
*\* (Mix) Failed to open Hex registry (file does not exist, run `mix hex.info` to fetch it).

### DIFF
--- a/blueprint/api.apib
+++ b/blueprint/api.apib
@@ -1237,6 +1237,7 @@ This resource identifies a relationship between a User and a Skill. For example,
         + data
             + Include (User Resource Identifier)
             + attributes (object)
+                + name: `Test` (string, optional)
                 + first_name: `Test` (string, optional)
                 + last_name: `User` (string, optional)
                 + website: `https://testuser.org` (string, optional)
--- a/web/views/user_view.ex
+++ b/web/views/user_view.ex
@@ -3,7 +3,7 @@ defmodule CodeCorps.UserView do
   use JaSerializer.PhoenixView
 
   attributes [
-    :biography, :email, :first_name, :last_name,
+    :biography, :email, :name, :first_name, :last_name,
     :photo_large_url, :photo_thumb_url, :twitter,
     :username, :website, :state,
     :inserted_at, :updated_at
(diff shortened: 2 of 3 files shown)


web/views/user_view.ex
```diff
@@ -20,13 +20,18 @@ defmodule CodeCorps.UserView do
   has_many :user_skills, serializer: CodeCorps.UserSkillView
 
   def photo_large_url(user, _conn) do
+    IO.inspect user
```
joshsmith, line 23: Will want to remove this.

web/views/user_view.ex
```diff
@@ -3,7 +3,7 @@ defmodule CodeCorps.UserView do
   use JaSerializer.PhoenixView
 
   attributes [
-    :biography, :email, :first_name, :last_name,
+    :biography, :email, :name, :first_name, :last_name,
```
joshsmith, line 6: I would keep this alphabetical.

blueprint/api.apib
```diff
@@ -1237,6 +1237,7 @@ This resource identifies a relationship between a User and a Skill. For example,
         + data
             + Include (User Resource Identifier)
             + attributes (object)
+                + name: `Test` (string, optional)
```
joshsmith, line 1240: This would actually be `Test User` since it's a combination of the two.

joshsmith: I would also specify that it's read-only in the documentation and that it's simply a combination of `first_name` and `last_name`.

web/views/user_view.ex
```diff
@@ -20,13 +20,18 @@ defmodule CodeCorps.UserView do
...
+    IO.inspect user
     CodeCorps.UserPhoto.url({user.photo, user}, :large)
   end
 
   def photo_thumb_url(user, _conn) do
     CodeCorps.UserPhoto.url({user.photo, user}, :thumb)
   end
 
+  def name(user, _conn) do
+    user.first_name <> " " <> user.last_name
```
joshsmith, line 32: I would rethink this to not add in the empty space `" "` in the event that either `first_name` or `last_name` are blank. You can probably do this via pattern matching on the `name` function.

joshsmith (changes requested):
Some minor changes requested here.

joshsmith:
Hey @schippada just checking in again. I know work has been crazy. We may need to close this, though, since it's so out of sync.

PR #262, opened 2016-09-25
Setup initial stripe config, add ENV, improve README

joshsmith:
- Add stripe config
- Add more ENV
- Improve README
- Add usage docs for using Stripe locally (not pretty, but better than nothing)
- Reordered `mix.exs` to be alpha for non-core libraries and applications


rileytaylor (approved):
:shipit: 

Stripe usage docs look fine.

PR #270, opened 2016-09-26
Deploy Elixir docs to Github Pages

### LINKED ISSUE #266: Add ex_doc documentation building and publish to GitHub Pages
We should use `ex_doc` to build our inline docs and publish them to GitHub Pages.

joshsmith:
This pushes Elixir docs to Github Pages: https://code-corps.github.io/code-corps-api-github-pages/api-reference.html

We can add a `CNAME` but not necessary right now.

Could use thoughts on whether this is best way to deploy.

Closes #266.


rileytaylor (approved):
lgtm. I don't have any thoughts on other ways to deploy, this is simple enough and gets the job done without doing any manual work.

PR #274, opened 2016-09-27
Eliminate many to many from views

### LINKED ISSUE #273: Eliminate many to many (join) relationship from all views
https://github.com/code-corps/code-corps-ember/pull/466 eliminates many-to-many-relationships from all client-side models, instead opting to rely on join relationships.

The same should be done API side, with views.

begedin:
Closes #273 

I've also updated preload calls on all affected controllers, so they match what the view requires. Since they all seem to be identical across all controller actions, we should consider moving them into a helper.

Something like

``` Elixir

def preload(query, _action), do: query |>preload([:slugged_route, :organization_memberships, :user_categories, :user_roles, :user_skills])

# user_controller.ex

def index(conn, params) do
  users =
    User
    |> User.index_filters(params)
    |> preload
    |> Repo.all

  render(conn, "index.json-api", data: users)
end
```

The unused action parameter is so that, in the future, if we need it, we can use separate preloads for different actions (if the actions start using different serializers). All that is for a separate issue, though.

### DIFF
--- a/elixir_buildpack.config
+++ b/elixir_buildpack.config
@@ -1,2 +1,2 @@
 erlang_version=19.1
-elixir_version=1.3.3
+elixir_version=1.3.2
--- a/mix.exs
+++ b/mix.exs
@@ -4,7 +4,7 @@ defmodule CodeCorps.Mixfile do
   def project do
     [app: :code_corps,
      version: "0.0.1",
-     elixir: "~> 1.3.3",
+     elixir: "1.3.2",
      elixirc_paths: elixirc_paths(Mix.env),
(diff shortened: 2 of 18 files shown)


joshsmith:
@begedin some of this duplication could probably be reduced in the future by using `ja_resource` in addition to `ja_serializer`. It may not work for much of what we're doing now, but it should work for at least some of our endpoints.

joshsmith:
I am going to wait to review this until I can pull down Ember and check both manually. Assigning to self for now. Don't let me forget. 😛

joshsmith (approved):
👍

PR #277, opened 2016-09-27
Add show and index actions to ProjectCategory

joshsmith:
This fixes an issue whereby we had no `ProjectCategory` `show` and `index` actions.


dmvt (approved):
Looks good pending passing CI

PR #291, opened 2016-09-28
Render state_transition for user

joshsmith:
Closes https://github.com/code-corps/code-corps-ember/issues/445

Simply renders the virtual attribute of `state_transition` which will tell the client we are no longer undergoing a transition.


begedin (approved):
About as straightforward as it gets. I didn't realize what the actual change API-side was until now. Looks good 👍

joshsmith:
@begedin I did a bad job of explaining then. 😭 

LGTM 🐑

PR #293, opened 2016-09-29
Rename posts to tasks

### LINKED ISSUE #244: Rename posts to tasks
In advance of our redesign, we want to rename Posts and all reference to them as Tasks. This will include things like `post_type` => `task_type`, or maybe more simply: `type`.

joshsmith:
This renames every `Post` and `post` to `Task` and `task`.

Also has a relatively complex migration that someone should give a careful once-over to be sure that everything worked as expected.

Closes #244.


venkatd (commented):
Lookin' good

venkatd (approved):
Approved

PR #299, opened 2016-09-30
Add test helpers to reduce duplicated assertions

sbatson5:
Why:
There seem to be quite a few duplicated assertions.  We should work on some test helpers to make these tests a bit more idiomatic and reduce some redundancy around what we are testing.

This PR:
Adds a test helper for asserting ids in a payload
Adds a helper for 401 responses
Leverages Ex-machina methods to reduce the number of inserts

### DIFF
--- a/test/controllers/organization_memberships_controller_test.exs
+++ b/test/controllers/organization_memberships_controller_test.exs
@@ -25,96 +25,72 @@ defmodule CodeCorps.OrganizationMembershipControllerTest do
 
   describe "index" do
     test "lists all resources", %{conn: conn} do
-      membership_1 = insert(:organization_membership)
-      membership_2 = insert(:organization_membership)
+      [membership_1, membership_2] = insert_pair(:organization_membership)
 
       path = conn |> organization_membership_path(:index)
+      response = conn |> get(path) |> json_response(200)
 
-      data = conn |> get(path) |> json_response(200) |> Map.get("data")
-      assert data |> length == 2
-
-      [first_result, second_result] = data
-      assert first_result["id"] == "#{membership_1.id}"
-      assert second_result["id"] == "#{membership_2.id}"
+      assert ids_from_response(response) == [membership_1.id, membership_2.id]
     end
 
     test "lists all resources for specified organization", %{conn: conn} do
       organization = insert(:organization)
-      membership_1 = insert(:organization_membership, organization: organization)
-      membership_2 = insert(:organization_membership, organization: organization)
+      [membership_1, membership_2] = insert_pair(:organization_membership, organization: organization)
       insert(:organization_membership)
 
       path = conn |> organization_organization_membership_path(:index, organization)
+      response = conn |> get(path) |> json_response(200)
 
-      data = conn |> get(path) |> json_response(200) |> Map.get("data")
-      assert data |> length == 2
-
-      [first_result, second_result] = data
-      assert first_result["id"] == "#{membership_1.id}"
-      assert second_result["id"] == "#{membership_2.id}"
+      assert ids_from_response(response) == [membership_1.id, membership_2.id]
     end
 
     test "filters resources by membership id", %{conn: conn} do
-      membership_1 = insert(:organization_membership)
-      membership_2 = insert(:organization_membership)
+      [membership_1, membership_2] = insert_pair(:organization_membership)
       insert(:organization_membership)
 
       params = %{"filter" => %{"id" => "#{membership_1.id},#{membership_2.id}"}}
-      conn = get conn, organization_membership_path(conn, :index, params)
-      data = json_response(conn, 200)["data"]
-      assert data |> length == 2
+      response =
+        conn
+        |> get(organization_membership_path(conn, :index, params))
+        |> json_response(200)
 
-      [first_result, second_result] = data
-      assert first_result["id"] == "#{membership_1.id}"
-      assert second_result["id"] == "#{membership_2.id}"
+      assert ids_from_response(response) == [membership_1.id, membership_2.id]
     end
 
     test "filters resources by role", %{conn: conn} do
-      membership_1 = insert(:organization_membership, role: "admin")
-      membership_2 = insert(:organization_membership, role: "admin")
+      [membership_1, membership_2] = insert_pair(:organization_membership, role: "admin")
       insert(:organization_membership, role: "owner")
 
       params = %{"role" => "admin"}
-      path = conn |> organization_membership_path(:index, params)
+      response =
+        conn
+        |> get(organization_membership_path(conn, :index, params))
+        |> json_response(200)
 
-      data = conn |> get(path) |> json_response(200) |> Map.get("data")
-      assert data |> length == 2
-
-      [first_result, second_result] = data
-      assert first_result["id"] == "#{membership_1.id}"
-      assert second_result["id"] == "#{membership_2.id}"
+      assert ids_from_response(response) == [membership_1.id, membership_2.id]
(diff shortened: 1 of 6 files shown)


test/controllers/user_controller_test.exs
```diff
@@ -34,8 +34,7 @@ defmodule CodeCorps.UserControllerTest do
     end
 
     test "filters resources on index", %{conn: conn} do
-      user_1 = insert(:user, username: "user_1", email: "user_1@mail.com")
-      user_2 = insert(:user, username: "user_2", email: "user_2@mail.com")
```
sbatson5, line 38: We don't seem to match on the username or email. Are they necessary to insert?

joshsmith: @sbatson5 I don't believe they would be, no.

joshsmith:
@sbatson5 it looks like your build is failing because the forked repo does not have access to `MIX_ENV`, which is explicitly required by `sentry-elixir`.

Perhaps we need to change the `circle.yml` to explicitly run with `MIX_ENV=test`?

sbatson5:
@JoshSmith Can't seem to get this suite running here or locally. Any tips?

joshsmith:
@sbatson5 I don't think it should go in `post`. I would think it would go in `override` `test` and you would do `MIX_ENV=test mix test`. We need to play around with it.

joshsmith:
@sbatson5 since we eventually circle back to rebasing and squashing, merging `develop` into your branch can make things more painful on you at that time, as you may have to resolve merge conflicts twice. I suggest rebasing without squashing every so often instead of merging in `develop`.

joshsmith:
Quick bump on this.

test/controllers/token_controller_test.exs
```diff
@@ -77,10 +73,11 @@ defmodule CodeCorps.TokenControllerTest do
...
-      assert error["id"] == "UNAUTHORIZED"
-      assert error["title"] == "401 Unauthorized"
-      assert error["status"] == 401
+      assert_unauthorized(error)
       assert error["detail"] == "token_expired"
     end
   end
+  
+  defp assert_unauthorized(%{"id" => "UNAUTHORIZED", "title" => "401 Unauthorized", "status" => 401}), do: true
+  defp assert_unauthorized(_), do: false
```
joshsmith: This could be nice to extract out elsewhere, too!

sbatson5: Slightly messed this up, but yeah may be a nice helper. Will push up a revision

test/controllers/organization_memberships_controller_test.exs
```diff
@@ -55,23 +54,21 @@ defmodule CodeCorps.OrganizationMembershipControllerTest do
...
-      data = json_response(conn, 200)["data"]
-      assert data |> length == 2
+      response = conn
+        |> get(organization_membership_path(conn, :index, params))
+        |> json_response(200)
 
-      [first_result, second_result] = data
-      assert first_result["id"] == "#{membership_1.id}"
-      assert second_result["id"] == "#{membership_2.id}"
+      assert response["data"] |> length == 2
```
sbatson5: @JoshSmith Thoughts on whether this assertion is redundant? If we assert that we get 2 ids back below, isn't that same as asserting the number of records?

joshsmith: @sbatson5 yeah, seems redundant to me.

test/controllers/organization_memberships_controller_test.exs
```diff
@@ -25,96 +25,70 @@ defmodule CodeCorps.OrganizationMembershipControllerTest do
...
       path = conn |> organization_organization_membership_path(:index, organization)
+      response = conn |> get(path) |> json_response(200)
 
-      data = conn |> get(path) |> json_response(200) |> Map.get("data")
-      assert data |> length == 2
-
-      [first_result, second_result] = data
-      assert first_result["id"] == "#{membership_1.id}"
-      assert second_result["id"] == "#{membership_2.id}"
+      assert ids_from_response(response) == [membership_1.id, membership_2.id]
```
sbatson5, line 44: @JoshSmith Let me know if you think these tests read a bit better now. We should be able to reuse `ids_from_response` in a few places, but didn't want to go crazy if it didn't read well.

SViccari: I think `ids_from_response` is a great addition! It makes tests easier to write _and_ read 🐎

SViccari:
This looks like a great set of changes! 👍  I like how you've introduced a new test helper and demonstrated it's usefulness, without retroactively changing all the test files at once.

test/controllers/organization_memberships_controller_test.exs
joshsmith: @sbatson5 agree with @SViccari.

joshsmith:
@sbatson5 let me know when I should review this.

sbatson5:
@JoshSmith I think this is in a good place to merge. I have some other refactors but would like to tackle them in other PRs and make them a bit more digestible.

SViccari:
@sbatson5 tiny nitpick, it looks like your first commit message has a duplication :)

sbatson5:
@SViccari fixed 😛

joshsmith:
Would just rebase and squash into a single commit, and then good to go!

joshsmith (approved):
:

joshsmith:
Hah that was supposed to be a 👍 but can't edit it.

test/controllers/organization_memberships_controller_test.exs
```diff
@@ -25,96 +25,70 @@ defmodule CodeCorps.OrganizationMembershipControllerTest do
...
       insert(:organization_membership, role: "owner")
 
       params = %{"role" => "admin"}
-      path = conn |> organization_membership_path(:index, params)
-
-      data = conn |> get(path) |> json_response(200) |> Map.get("data")
-      assert data |> length == 2
+      response = conn
+        |> get(organization_membership_path(conn, :index, params))
+        |> json_response(200)
```
joshsmith, line 68: Quick stylistic note here:

```
response =
  conn
  |> get(organization_membership_path(conn, :index, params))
  |> json_response(200)
```

joshsmith: There are a couple places where I would make this change.

joshsmith:
@sbatson5 missed that comment above until a second look.

test/controllers/organization_memberships_controller_test.exs
sbatson5: Should have fixed both now 👍

test/controllers/token_controller_test.exs
```diff
@@ -77,10 +73,11 @@ defmodule CodeCorps.TokenControllerTest do
...
       [error | _] = response["errors"]
-      assert error["id"] == "UNAUTHORIZED"
-      assert error["title"] == "401 Unauthorized"
-      assert error["status"] == 401
+      assert is_unauthorized?(error)
       assert error["detail"] == "token_expired"
     end
   end
+
+  defp is_unauthorized?(%{"id" => "UNAUTHORIZED", "title" => "401 Unauthorized", "status" => 401}), do: true
```
sbatson5, line 81: Opted to leave these methods here, as I didn't see any other tests making this assertion. Happy to extract if there is another use-case, but as it seems specific to this test, will leave here for now.

joshsmith:
Will merge as soon as this build passes.

joshsmith:
🙌 it's in!

PR #300, opened 2016-10-01
INSTALLING.md remove Ruby reference

rileytaylor:
Rogue 'ruby' spotted resisting the new Elixir order. It has been dealt with.


joshsmith (approved):
👍

PR #301, opened 2016-10-01
Add test for preview view

### LINKED ISSUE #285: Write test for previews view

Le6ow5k1:
Fixes #285

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -13,7 +13,9 @@ dependencies:
     - mix compile
 test:
   override:
-    - mix test --include requires_env --cover
+    - |
+      if [ ${CIRCLE_PR_USERNAME} ]; then MIX_ENV=test mix test --cover; else MIX_ENV=test mix test --include requires_env --cover; fi
+
   post:
     - mix inch.report
 
--- a/test/support/factories.ex
+++ b/test/support/factories.ex
@@ -128,4 +128,12 @@ defmodule CodeCorps.Factories do
       skill: build(:skill)
     }
   end
+
+  def preview_factory do
+    %CodeCorps.Preview{
+      body: "Bar",
+      markdown: "Bar",
+      user: build(:user)
+    }
+  end
 end
--- a/test/views/preview_view_test.exs
+++ b/test/views/preview_view_test.exs
@@ -1,0 +1,44 @@
+defmodule CodeCorps.PreviewViewTest do
+  use CodeCorps.ConnCase, async: true
+
+  alias CodeCorps.Repo
+
+  import Phoenix.View
+
+  test "renders all attributes and relationships properly" do
+    user = insert(:user)
+    preview = insert(:preview, user: user)
+
+    preview = CodeCorps.Preview
+    |> Repo.get(preview.id)
+    |> Repo.preload([:user])
+
+    rendered_json = render(CodeCorps.PreviewView, "show.json-api", data: preview)
+
+    expected_json = %{
+      data: %{
+        id: preview.id |> Integer.to_string,
+        type: "preview",
+        attributes: %{
+          "body" => preview.body,
+          "inserted-at" => preview.inserted_at,
+          "markdown" => preview.markdown,
+          "updated-at" => preview.updated_at
+        },
+        relationships: %{
+          "user" => %{
+            data: %{
+              id: preview.user_id |> Integer.to_string,
+              type: "user"
+            }
+          }
+        }
+      },
+      jsonapi: %{
+        version: "1.0"
+      }
+    }
+
+    assert expected_json == rendered_json
+  end
+end


joshsmith:
@Le6ow5k1 thanks for the PR! Tests are failing for reasons unrelated to your pull. Forks don't have access to our ENV.

In `circle.yml`, can you change:

``` yaml
test:
  override:
    - mix test --include requires_env --cover
```

to

``` yaml
test:
  override:
    - MIX_ENV=test mix test --include requires_env --cover
```

Push that as a separate commit, if you would? Thanks!

joshsmith:
Actually, hold tight on this suggestion for a moment.

Le6ow5k1:
Sorry, already pushed. It seems that it needs a complete set of env variables.

joshsmith:
@Le6ow5k1 it's running the `requires_env` tags, which you don't have on a fork for security reasons in CI.

I'm going to need to circle back to this unfortunately. Today's my wife and my anniversary.

I do think the issue can be resolved not by removing `requires_env` (we _really_ need that on non-fork CIs), but maybe by running some Circle config that works differently on forks.

joshsmith:
We could run an if/else check on `CIRCLE_PR_USERNAME` in the `test` command so that we _only_ run the command if we assume you have the environment (it's not a fork).

joshsmith:
@Le6ow5k1 looks like you fixed it! Apologies for the difficulties there. A new package changed some behavior along with our recent change to run those require_envs on CI. We hadn't a fork push since so it wasn't reflected.

circle.yml
```diff
@@ -13,7 +13,10 @@ dependencies:
     - mix compile
 test:
   override:
-    - MIX_ENV=test mix test --cover
+    - |
+      if [ ${CIRCLE_PR_USERNAME} ]; then MIX_ENV=test mix test --cover; else MIX_ENV=test mix test --include requires_env --cover; fi
```
sbatson5, line 17: 👏 Nice solution! Could we get this as a separate PR?

test/support/factories.ex
```diff
@@ -128,4 +128,12 @@ defmodule CodeCorps.Factories do
       skill: build(:skill)
     }
   end
+
+  def preview_factory do
+    %CodeCorps.Preview{
+      body: "Foo",
+      markdown: "Bar",
```
joshsmith, line 135: For this part I would make the markdown and body match up. Also, does the factory need the body of it has the markdown? There seem to be inconsistencies with other factories.

test/views/preview_view_test.exs
```diff
@@ -0,0 +1,44 @@
...
+
+    expected_json = %{
+      data: %{
+        id: preview.id |> Integer.to_string,
+        type: "preview",
+        attributes: %{
+          "markdown" => preview.markdown,
+          "body" => preview.body,
+          "inserted-at" => preview.inserted_at,
+          "updated-at" => preview.updated_at
```
joshsmith, line 26: Would alpha order these.  Need to add this to a style guide.

circle.yml
```diff
@@ -13,7 +13,10 @@ dependencies:
     - mix compile
 test:
   override:
-    - mix test --include requires_env --cover
+    - |
+      if [ ${CIRCLE_PR_USERNAME} ]; then MIX_ENV=test mix test --cover; else MIX_ENV=test mix test --include requires_env --cover; fi
```
joshsmith, line 17: This looks great. Can we add a line to docs explaining the change? Think there's à CI section.

joshsmith (changes requested):
Minor changes here. After those will accept and ask you to rebase and squash after that. Then we'll merge!

🎈 Thanks for this.

circle.yml
SViccari: Out of curiosity, is there a benefit to having `CIRCLE_PR_USERNAME` set for some but not others? Would it be reasonable for the test suite to run the same way for everyone?

joshsmith: @SViccari the test suite can't run the same way for everyone. See https://circleci.com/docs/fork-pr-builds/#security-implications-of-running-builds-for-pull-requests-from-forks for discussion of security implications of leaking sensitive environment variables.

circle.yml
Le6ow5k1: i think there is no need for that, hopefully PR will be merged soon

test/support/factories.ex
Le6ow5k1: ~~you're right only markdown is needed~~

joshsmith (approved):

joshsmith:
@Le6ow5k1 just squash and rebase when you have a moment, then we'll merge. Thanks!

Le6ow5k1:
@JoshSmith 📦

joshsmith:
❤️ 🎈 🎉

PR #307, opened 2016-10-02
Fix doc deploy

joshsmith:
Should _finally_ fix doc deploys.


rileytaylor (approved):
lgtm :shipit: hopefully this does it... but I believe that git diff should ensure it is taken care of.

Issue #309, opened 2016-10-03
Add not null constraints everywhere needed


joshsmith:
- [x] `comments` is missing `user_id`, `task_id`
- [x] `organizations` is missing `name`, `description`, `slug`
- [x] `previews` missing `user_id`
- [x] `project_skills` missing `project_id`, `skill_id`
- [x] `projects` missing `organization_id`
- [x] `role_skills` missing `role_id`, `skill_id`
- [x] `slugged_routes` missing `slug`
- [x] `tasks` missing `title`, `state` (which is also missing a default)
- [x] `user_categories` missing `user_id`, `category_id`
- [x] `user_roles` missing `user_id`, `role_id`
- [x] `user_skills` missing `user_id`, `skill_id`

asmita005:
Hello, i would like to contribute.
where do i locate the above files to make changes to them ?

Thanks

joshsmith:
Closed by #310.

joshsmith:
@asmita005 sorry this got jumped on so quickly before I could respond! Any of the other API tasks would be good to work on, though.

PR #310, opened 2016-10-03
Not null constraints

mitchellhenke:
For #309

Unfortunately no way to add the constraint within Ecto's migration syntax that I could find.

### DIFF
--- a/priv/repo/migrations/20161003185918_add_not_null_constraints.exs
+++ b/priv/repo/migrations/20161003185918_add_not_null_constraints.exs
@@ -1,0 +1,31 @@
+defmodule CodeCorps.Repo.Migrations.AddNotNullConstraints do
+  use Ecto.Migration
+
+  def up do
+    execute "ALTER TABLE comments ALTER COLUMN user_id SET NOT NULL, ALTER COLUMN task_id SET NOT NULL"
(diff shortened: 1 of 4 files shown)


mitchellhenke:
hm, swear I ran the tests locally and they passed. will fix

joshsmith:
@mitchellhenke just parked my car on way to meeting so sorry for abruptness, but I think this is what you're looking for https://hexdocs.pm/ecto/Ecto.Migration.html#modify/3

Should be able to just add `:null` under options mentioned there in docs.

mitchellhenke:
@JoshSmith I tried this:

``` elixir
    alter table(:user_skills) do
      modify :user_id, references(:users), null: false
    end
```

but I get:

```
16:22:49.050 [info]  alter table user_skills
** (Postgrex.Error) ERROR (duplicate_object): constraint "user_skills_user_id_fkey" for relation "user_skills" already exists
```

joshsmith:
@mitchellhenke you're right there I think. For the non-references you should be able to do it without the execute (no existing constraint). But this is simple enough that I'm okay with it. We execute more SQL than that elsewhere, believe me!

joshsmith (approved):

joshsmith:
@mitchellhenke can you fetch upstream `develop` and rebase on that and squash into one commit? Then good to merge!

mitchellhenke:
@JoshSmith updated!

PR #311, opened 2016-10-03
run mix.deps before building the Docker image

SViccari:
Hi there!

First PR for this repo 😸  I'm eager to help resolve some of the issues I've encountered when trying to run this application locally with Docker and docker-compose. This PR is limited to Dockerfile changes as I'm still reading the docker-compose docs to identify the ideal way to use `volumes` (aka, mapping to include local file changes).

Why:
I ran into a number of errors when running docker-compose up and then
running the command 'docker-compose run test mix test'. The most common
error being 'could not compile dependency :dependency_name, "mix compile"
failed'.

This PR:
Copies the necessary mix files into the app directory and then fetches
all the dependencies.

Detailed error message referenced above:

```
==> arc
could not compile dependency :arc, "mix compile" failed. You can recompile this dependency with "mix deps.compile arc", update it with "mix deps.update arc" or clean it with "mix deps.clean arc"
==> code_corps
** (Mix) Failed to open Hex registry (file does not exist, run `mix hex.info` to fetch it)
```

PR #314, opened 2016-10-04
Fix source code updating issue in docker-compose

fahrradflucht:
This should fix the issues regarding code updates not showing up in the containers.

As far as my testing goes #311 is still needed to fix the dependency issues. (The issues are still present and I needed to start up twice to fix them. I didn't try the proposed fix.)

The reverted commit removed [this line](https://github.com/code-corps/code-corps-api/commit/f6011cf291292967d49d22d3c5421755b9f484b3#diff-4e5e90c6228fd48698d074241c2ba760L40) as well. It shouldn't be needed to fix the afore mentioned issue.

Now that I think about that. Could re-adding that fix the deps issue @SViccari ?

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -33,7 +33,7 @@ services:
     ports:
       - 49235:4000
     volumes:
-      - code-corps-api:/code-corps-api
+      - .:/code-corps-api
 
   test:
     <<: *api_base
@@ -52,8 +52,7 @@ services:
     ports:
       - 8081:8080
     volumes:
-      - code-corps-api:/code-corps-api
+      - .:/code-corps-api
 
 volumes:
-  code-corps-api:
   postgres-data:


docker-compose.yml
```diff
@@ -33,7 +33,7 @@ services:
     ports:
       - 49235:4000
     volumes:
-      - code-corps-api:/code-corps-api
+      - .:/code-corps-api
```
SViccari, line 36: nice! This is also what I had to do locally to register local file changes

joshsmith (approved):
👍 this is excellent. Thank you!

joshsmith:
@Fahrradflucht just needs a fetch and rebase.

fahrradflucht:
Uh I messed that one up. Fix coming. - Done

SViccari:
@Fahrradflucht That was actually one of the first things I tried, was adding `command: bash -c "mix do deps.get, compile"` to the `test` service  in the docker-compose file but I still ran into the same `mix compile failed` errors. Updating the Dockerfile is the only strategy that has worked for me.

PR #316, opened 2016-10-04
Fix seeds.exs to meet not null contraints

fahrradflucht:
5e6b235c3078f9dbacfa5295e27b5b3fc770a1d0 introduced a not null constraint for the organization description.

This adds a description to the seeded org to meet that constraint.
## 

While investigating this I noticed that the organization model doesn't require a description on the changeset. Is that intentional?

### DIFF
--- a/priv/repo/seeds.exs
+++ b/priv/repo/seeds.exs
@@ -41,7 +41,8 @@ end
 
 organizations = [
   %{
-    name: "Code Corps"
+    name: "Code Corps",
+    description: "Help build and fund public software projects for social good"
   },
 ]
 


joshsmith:
> the organization model doesn't require a description on the changeset

@Fahrradflucht no, that's not intentional. Feel free to tackle that. Thanks for noticing!

joshsmith (approved):
👍

joshsmith:
Good catch. We also really need to add in some `OrganizationMembership` relationships and similar relationships, because those are currently missing. Leads to a lot of heartache.

Feel like there should be some easier ways to seed than what we're doing right now generally. All feels a bit hacky.

PR #320, opened 2016-10-05
Add authorization based on changeset, for resources that need it

### LINKED ISSUE #176: Improve our authorization process to also check the changeset
From #149 

We need to find a way to provide a changeset, or params to `load_and_authorize_resource`when authorising the `:create` action. We need it in some cases, because in order to decide if the user is allowed to perform the action, we need to see what type of resource is being created. It is needed for: 
- `organization_membership` - User can only **create** their own pending membership and they should not be able to do it if they're already a member. The part of it being their own is handled by not allowing a `member_id` as a parameter. Instead, we just assign the current user. For **update**, we've simplified the permissions compared to rails, but ideally, we need to check what the current membership role is, and what we want it to be, then we want to decide if the user can do it based on their own role. - Related to #157, being partially handled in #158 
- `posts` - Any user, regardless of membership, can create ideas and issues. Other posts can only be created by contributors or higher, so we need to check the changeset for that.
- `project, project_category, project_skill` - We need to pass in the changeset so we can know which organization the record is being created for, since only some roles can do it.
- `user_category, user_role, user_skill` - These might actually be fine. We need to check to see if the user is modifying their own record on create. We could handle that without permissions, by doing it on the changeset.

begedin:
Work in progress, closes #176 

This is definitely still a work in progress, but I've formulated a methodology for this, which allows us to just us a `load_and_authorize_changeset` plug. I'm still working out the details, but the tests pass and it seems to be working in the organization membership controller, so that would be the one I would look at.

### DIFF
--- a/test/controllers/organization_memberships_controller_test.exs
+++ b/test/controllers/organization_memberships_controller_test.exs
@@ -110,18 +110,15 @@ defmodule CodeCorps.OrganizationMembershipControllerTest do
     end
 
     test "renders page not found when id is nonexistent", %{conn: conn} do
-      assert_error_sent 404, fn ->
-        path = conn |> organization_membership_path(:show, -1)
-        conn |> get(path)
-      end
+      path = conn |> organization_membership_path(:show, -1)
+      assert conn |> get(path) |> json_response(:not_found)
     end
   end
 
   describe "create" do
     @tag :authenticated
-    test "creates and renders resource when data is valid", %{conn: conn} do
+    test "creates and renders resource when data is valid", %{conn: conn, current_user: member} do
       organization = insert(:organization)
-      member = insert(:user)
 
       payload = build_payload |> put_relationships(organization, member)
 
@@ -141,8 +138,11 @@ defmodule CodeCorps.OrganizationMembershipControllerTest do
     end
 
     @tag :authenticated
-    test "does not create resource and renders 422 when data is invalid", %{conn: conn} do
-      payload = build_payload
+    test "does not create resource and renders 422 when data is invalid", %{conn: conn, current_user: member} do
+      # only way to trigger a validation error is to provide a non-existant organization
+      # anything else will fail on authorization level
+      organization = build(:organization)
+      payload = build_payload |> put_relationships(organization, member)
 
       path = conn |> organization_membership_path(:create)
       data = conn |> post(path, payload) |> json_response(422)
@@ -155,7 +155,7 @@ defmodule CodeCorps.OrganizationMembershipControllerTest do
     @tag :authenticated
     test "updates and renders resource when data is valid", %{conn: conn, current_user: current_user} do
       organization = insert(:organization)
-      membership = insert(:organization_membership, organization: organization)
+      membership = insert(:organization_membership, organization: organization, role: "pending")
       insert(:organization_membership, organization: organization, member: current_user, role: "owner")
 
       payload = build_payload |> put_id(membership.id) |> put_attributes(@valid_attrs)
@@ -173,23 +173,6 @@ defmodule CodeCorps.OrganizationMembershipControllerTest do
       assert membership.role == "contributor"
       assert membership.organization_id == membership.organization_id
       assert membership.member_id == membership.member_id
-    end
-
-    @tag :authenticated
-    test "doesn't update and renders 422 when data is invalid", %{conn: conn, current_user: current_user} do
-      organization = insert(:organization)
-      membership = insert(:organization_membership, organization: organization)
-      insert(:organization_membership, organization: organization, member: current_user, role: "owner")
-
-      payload =
-        build_payload
-        |> put_id(membership.id)
-        |> put_attributes(@invalid_attrs)
-
-      path = conn |> organization_membership_path(:update, membership)
-      conn = conn |> put(path, payload)
-
-      assert conn |> json_response(422)
     end
 
     test "doesn't update and renders 401 when unauthenticated", %{conn: conn} do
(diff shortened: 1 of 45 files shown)


green-arrow:
@begedin I'm curious as to your thoughts on putting authorization logic inside of the model. It seems like it may have the potential to make the models pretty bloated and change the scope of what a model handles from just DB interaction to including business logic.

In the past I've generally made a `service` layer that handles all business logic, so the flow becomes:

```
controller
      |
      V
services
      |
      V
model / repo
```

Thoughts?

begedin:
@green-arrow I'm not really sure what you mean by it.

There is a separate, policy level here, but the problem is, the user being or not being authorized for an action depends on the specificity of that action - what changes are being made, what the current state of the record is, etc.

Due to that, each policy requires an action name, the current user record and the resource the action is performed on (a model or a changeset).

This PR does not add into models, it uses them and some additional authorization.

Basically, I agree with your approach and I don't like fat models, but models aren't really growing in size here.

green-arrow:
@begedin - ah my bad, I misread the file names and thought the policy file was the model.

How do you feel about the changes logic being hidden behind the plug? Feels a little implicit to me, like just trusting this magic thing off of `conn.assigns` to be there, but maybe that's ok :)

begedin:
It does feel implicit, but it's basically the same approach that Canary uses, so this is just an additional plug, used for our specific purposes, heavily inspired by canary.

I wrote it in such a way that all assigns the plug relies on are actually assigns that the plug itself is setting, so it should be relatively self-contained.

test/controllers/organization_memberships_controller_test.exs
```diff
@@ -110,18 +110,15 @@ defmodule CodeCorps.OrganizationMembershipControllerTest do
     end
 
     test "renders page not found when id is nonexistent", %{conn: conn} do
-      assert_error_sent 404, fn ->
-        path = conn |> organization_membership_path(:show, -1)
-        conn |> get(path)
-      end
+      path = conn |> organization_membership_path(:show, -1)
+      assert conn |> get(path) |> json_response(:not_found)
```
begedin, line 114: These actually weren't rendering json, due to not using Canary's `load_resource`, which would then trigger a not-found handler.

web/controllers/organization_membership_controller.ex
```diff
@@ -19,19 +21,12 @@ defmodule CodeCorps.OrganizationMembershipController do
...
 
-  def show(conn, %{"id" => id}) do
-    membership =
-      OrganizationMembership
-      |> preload([:organization, :member])
-      |> Repo.get!(id)
-
-    render(conn, "show.json-api", data: membership)
+  def show(conn, %{"id" => _id}) do
+    render(conn, "show.json-api", data: conn.assigns.organization_membership)
```
begedin, line 24: When we use `load_and_authorize_resource` or `load_resource`, Canary will load the resource for us, by id and assign it to the conn. Since we're bound to use `load_and_authorize_resource` in many places, even without this PR, we should use it. Otherwise, we just end up loading things from the database twice.

web/models/organization_membership.ex
```diff
@@ -18,6 +18,7 @@ defmodule CodeCorps.OrganizationMembership do
 
   @doc """
   Builds a changeset based on the `struct` and `params`, for creating a record.
+  The membership role is strictly set to "pending" by the system, regardless of parameters
```
begedin, line 21: Thought I should add this here, just in case.

web/models/project_skill.ex
```diff
@@ -11,7 +11,7 @@ defmodule CodeCorps.ProjectSkill do
   @doc """
   Builds a changeset based on the `struct` and `params`.
   """
-  def changeset(struct, params \\ %{}) do
+  def create_changeset(struct, params \\ %{}) do
```
begedin, line 14: Thew new `load_and_authorize_changeset` plug automatically calls the changeset method, inferred from action name - `#{action}_changeset`. Due to that, I renamed a couple of these. 

It's a similar approach to what `Canary` uses for `load_resource`.

web/web.ex
```diff
@@ -38,6 +38,7 @@ defmodule CodeCorps.Web do
       import CodeCorps.Gettext
 
       import Canary.Plugs
+      import CodeCorps.AuthenticationHelpers, only: [load_and_authorize_changeset: 2]
```
begedin, line 41: I was actually thinking of just extending `Canary.Plugs` and importing that module alone. Not sure what the best approach would be.

joshsmith (approved):
Looks good overall. Honestly this PR is so large it's hard for me to accurately judge. But generally looks like a good approach.

web/controllers/organization_membership_controller.ex
joshsmith: This `conn.assigns.organization_membership` reads odd to me. Can you explain that?

test/controllers/organization_memberships_controller_test.exs
joshsmith: I see. This is also related to issue I raised with #324.

web/controllers/organization_membership_controller.ex
begedin: Explained it in the other PR. 

Both `load_and_authorize_resource` and `load_resource` plugs provided by `Canary` load the endpoint resource using (by default) the `id` params key. If the plugs manage to load it, it's assigned to `conn.assigns.underscored_model_name`. If it fails, the `handle_not_found` handler, which we defined, is triggered. This handler halts and renders a 404 JSON error.

When we do it manually, instead, the loading just fails and we end up with a 404 HTML error, which is a bug on it's own. 

Also, since we do use `load_and_authorize_resource` either way, that means the resource already gets loaded once. When we proceed to load it manually, we're actually end up loading it twice for endpoints that have authorization.

PR #323, opened 2016-10-06
Implement ex_machina strategy for JSON API payloads

### LINKED ISSUE #290: Write an ex_machina strategy to build JSON API payloads
Write an ex_machina strategy which would allow us simple building of JSON API payloads for tests.

See https://github.com/thoughtbot/ex_machina#custom-strategies

begedin:
Will close #290 

Tests are failing right now, due to breaking changes with a `ja_serializer` update.

This should serve as a proof of concept. The strategy has been implemented and applied to the category endpoint.

The update of `ja_serializer`
- requires updates to all views, since the relationships now need to be explicitly incuded (id, type pairs)
- requires updates to all view specs, since the serializer output now uses exclusively string keys for maps.
- removes the need for preloads on the controller level. they are now defined on the view level

Using the strategy in the category controller actually increases the code amount, but this is not the case in many other controllers, and additionally, a lot of repetition is now apparent, which could be further eliminated.


test/controllers/category_controller_test.exs
```diff
@@ -56,51 +46,67 @@ defmodule CodeCorps.CategoryControllerTest do
...
 
     @tag :authenticated
     test "does not update resource and renders 401 when not authorized", %{conn: conn} do
       category = insert(:category)
-      conn = put conn, category_path(conn, :update, category), build_payload(category.id, @invalid_attrs)
-      assert json_response(conn, 401)
+      payload = json_payload(:category, @valid_attrs)
+      path = conn |> category_path(:update, category)
+
+      assert conn |> put(path, payload) |> json_response(401)
```
joshsmith: Is there a better way to pipe the `conn` here that can be done in two lines instead of three?

test/views/category_view_test.exs
```diff
@@ -4,37 +4,33 @@ defmodule CodeCorps.CategoryViewTest do
   alias CodeCorps.Repo
 
   # Bring render/3 and render_to_string/3 for testing custom views
-  import Phoenix.View
+  import Phoenix.View, only: [render: 3]
```
joshsmith, line 4: Do we not need `render_to_string/3` here?

test/views/category_view_test.exs
```diff
@@ -4,37 +4,33 @@ defmodule CodeCorps.CategoryViewTest do
...
         },
-        type: "category",
+        "type" => "category",
       },
-      jsonapi: %{
-        version: "1.0"
+      "jsonapi" => %{
+        "version" => "1.0"
       }
     }
```
joshsmith, line 32: What changed to bring in `"` instead of atoms?

web/controllers/category_controller.ex
```diff
@@ -4,10 +4,11 @@ defmodule CodeCorps.CategoryController do
   alias CodeCorps.Category
   alias JaSerializer.Params
 
+  plug :load_resource, model: Category, only: [:show]
```
joshsmith, line 7: I'm not sure I understand really the implicit behavior here. We need `load_resource` for all resources that aren't plugged through `load_and_authorize_resource`, is that it?

web/views/category_view.ex
```diff
@@ -2,7 +2,11 @@ defmodule CodeCorps.CategoryView do
   use CodeCorps.Web, :view
   use JaSerializer.PhoenixView
 
+  def preload(query, _conn, []) do
+    query |> CodeCorps.Repo.preload([:project_categories])
+  end
```
joshsmith: Is this now recommended way according to `ja_serializer`?

web/views/category_view.ex
```diff
@@ -2,7 +2,11 @@ defmodule CodeCorps.CategoryView do
...
+
   attributes [:name, :slug, :description]
 
-  has_many :project_categories, serializer: CodeCorps.ProjectCategoryView
+  has_many :project_categories, serializer: CodeCorps.ProjectCategoryView, identifiers: :always
```
joshsmith, line 8: I know this isn't our DSL, but the `identifiers` language here is confusing.

joshsmith (commented):
Minor comments. I am happy to continue with this generally, though.

test/controllers/category_controller_test.exs
begedin: path is always the same and we always create a default category, so I could do a

``` Elixir
def request_update(conn, attrs) do
  category = insert(:category)
  path = conn |> category_path(:update, category)
  payload = json_payload(:category, attrs)
  conn |> put(path,  payload)
end
```

Then we just make a call:

``` Elixir
assert conn |> request_update(@valid_attrs) |> json_response(401)
assert conn |> request_update(@invalid_attrs) |> json_response(422)
# etc...
```

without any of the other lines.

test/views/category_view_test.exs
begedin: Ah, missed that. The comment is out of date. I don't think we ever needed it. Will correct.

test/views/category_view_test.exs
begedin: A change in `ja_serializer`. They just do it this way, always, now. All keys are strings. I guess because it's closer to actual JSON.

web/controllers/category_controller.ex
begedin: Pretty much. It's provided by `Pundit`, same as `load_and_authorize_resource`. It loads by `id` by default (we can provide a name for the `id` key), and assigns it to `conn.assings.underscored_model_name`. `load_and_authorize_resource` does the exact same thing but then also performs authorization on it. Also, if the resource with that id is not found, it automatically handles the error using `handle_not_found`, which we defined in `authentication_helpers`. That handler renders a JSON 404 error, unlike the default handler which triggers when we manually try and fail to load a resource, which renders an HTML 404 error.

web/views/category_view.ex
begedin: If you take a look at the documentation, it would seem so. They in fact recommend two methods. One for handling default preloads, and one for handling it when we specify preloads manually when calling `render` from the controller.

https://hexdocs.pm/ja_serializer/JaSerializer.Serializer.html#c:preload/3

web/views/category_view.ex
begedin: Agreed. And it took me a long time to find it to, but it seems to be the convention. It doesn't matter if we do the other changes (moving of preloads). Identifiers do not load unless we specify it this way.

begedin:
@JoshSmith Marked this as ready for review again.

The job isn't done, but all that's really left is the low hanging fruit, so I thought better to focus my attention elsewhere.

What this PR does is that it 
- Implements a strategy that generates a json_api payload for a single, or multiple unpersisted records, using an `ex_machina` factory.
- Upgrades `ja_serializer` to `0.11.0`
- Fixes issues create by breaking changes in `0.11.0`
  - Moved preloads from controllers to views
  - Added `identifiers: :always` to each `has_many` relationship in our views, so that the relationship identifier data is still rendered (no longer the case, by default). I initially thought this was needed for all relationships, but `has_one` does not require it.
  - Changed our tests to expect string keys in maps, since this is what `ja_serializer` now  renders
- Uses the newly implemented strategy on the category endpoint, as an example. This get's rid of the `build_payload` chain of helper functions from the `category_controller_test`.

Once we merge this we should create an issue to eliminate the `build_payload` helpers from other controller tests as well.

joshsmith (approved):
:

joshsmith:
- [ ] Squash and rebase
- [ ] create an issue to eliminate the build_payload helpers from other controller tests

PR #325, opened 2016-10-07
Add sentry user context plug

### LINKED ISSUE #303: Set Sentry user context
We want to set the Sentry user context like so:

``` elixir
Sentry.Context.set_user_context(%{id: user.id, email: user.email, first_name: user.first_name, last_name: user.last_name})
```

Should make a `CodeCorps.Plug.SetSentryUserContext` and some sort of module to contain that, using a similar approach to our `Analytics` module to switch between when we have Sentry enabled and when we do not. This may not be necessary, so attempt without it first (just call `Sentry.Context.set_user_context` directly and see what happens).

I would also consider condensing the plugs `CurrentUser`, `AnalyticsIdentify`, and this new plug into a single pipeline.

begedin:
Closes #303 

Merging of pipelines (suggested in the issue) is done as a separate commit, so we can rollback if wanted. Outside of that, this ought to be ready for merge, unless we want to write tests for it? Not sure how we would go about that. Probably a unit test for the plug itself, providing a conn with or without a current user.

### DIFF
--- a/web/plugs/set_sentry_user_context.ex
+++ b/web/plugs/set_sentry_user_context.ex
@@ -1,0 +1,17 @@
+defmodule CodeCorps.Plug.SetSentryUserContext do
+  def init(opts), do: opts
+
+  def call(conn, _opts), do: conn |> set_context
+
+  defp set_context(%{assigns: %{current_user: user}} = conn) do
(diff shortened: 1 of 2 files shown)


joshsmith (approved):
👍 just needs squashed.

PR #327, opened 2016-10-08
Add remaining policy tests

### LINKED ISSUE #174: Add tests for all of our authorization policies
If this turns out to bee too big for a single PR, feel free to create individual issues and tackle them separately
- [x] Category
- [x] Comment
- [x] Organization
- [x] OrganizationMembership
- [x] Preview
- [x] Project
- [x] ProjectCategory
- [x] ProjectSkill
- [x] Role
- [x] RoleSkill
- [x] Skill
- [x] Task
- [x] User
- [x] UserCategory
- [x] UserSkill
- [x] UserRole

joshsmith:
Closes #174 by adding the remaining policy tests not addressed in #320.
- Add category policy test
- Add role policy test
- Add skill policy test
- Add role skill policy test
- Add comment policy test
- Add preview policy test
- Add user policy test
- Add organization policy test
- Update abilities

### DIFF
--- a/test/policies/category_policy_test.exs
+++ b/test/policies/category_policy_test.exs
@@ -1,0 +1,29 @@
+defmodule CodeCorps.CategoryPolicyTest do
+  use CodeCorps.PolicyCase
+
+  import CodeCorps.CategoryPolicy, only: [create?: 1, update?: 1]
+
+  describe "create?" do
+    test "returns true when user is an admin" do
+      user = build(:user, admin: true)
+      assert create?(user)
+    end
+
+    test "returns false if user is not an admin" do
+      user = build(:user, admin: false)
+      refute create?(user) 
+    end
+  end
+
+  describe "update?" do
+    test "returns true when user is an admin" do
+      user = build(:user, admin: true)
+      assert update?(user)
+    end
+
+    test "returns false if user is not an admin" do
+      user = build(:user, admin: false)
+      refute update?(user)
+    end
+  end
+end
--- a/test/policies/comment_policy_test.exs
+++ b/test/policies/comment_policy_test.exs
@@ -1,0 +1,42 @@
+defmodule CodeCorps.CommentPolicyTest do
+  use CodeCorps.PolicyCase
+
+  import CodeCorps.CommentPolicy, only: [create?: 2, update?: 2]
+  import CodeCorps.Comment, only: [create_changeset: 2]
+
+  alias CodeCorps.Comment
+
+  describe "create?" do
+    test "returns true if own record" do
+      user = insert(:user)
+      changeset = %Comment{} |> create_changeset(%{user_id: user.id})
+      assert create?(user, changeset)
+    end
+
+    test "returns false if someone else's record" do
+      [user, another_user] = insert_pair(:user)
+      changeset = %Comment{} |> create_changeset(%{user_id: another_user.id})
+      refute create?(user, changeset)
+    end
+
+    test "returns false if changeset is empty" do
+      user = insert(:user)
+      changeset = %Comment{} |> create_changeset(%{})
+      refute create?(user, changeset)
+    end
+  end
+
+  describe "update?" do
+    test "returns true if own record" do
+      user = insert(:user)
+      comment = insert(:comment, user: user)
+      assert update?(user, comment)
+    end
+
+    test "returns false if someone else's record" do
+      [user, another_user] = insert_pair(:user)
+      comment = insert(:comment, user: user)
+      refute update?(another_user, comment)
+    end
+  end
+end
--- a/test/policies/organization_policy_test.exs
+++ b/test/policies/organization_policy_test.exs
@@ -1,0 +1,58 @@
+defmodule CodeCorps.OrganizationPolicyTest do
+  use CodeCorps.PolicyCase
+
+  import CodeCorps.OrganizationPolicy, only: [create?: 1, update?: 2]
+
+  defp setup_user_organization_by_role(role) do
+    user = insert(:user)
+    organization = insert(:organization)
+    insert(:organization_membership, role: role, member: user, organization: organization)
+    [user, organization]
+  end
+
+  describe "create" do
+    test "returns true when user is an admin" do
+      user = build(:user, admin: true)
+      assert create?(user)
+    end
+
+    test "returns false when user is not an admin" do
+      user = build(:user, admin: false)
+      refute create?(user)
+    end
+  end
+
+  describe "update" do
+    test "returns true when user is an admin" do
+      user = insert(:user, admin: true)
+      organization = insert(:organization)
+      assert update?(user, organization)
+    end
+
+    test "returns false when user is not member of organization" do
+      user = insert(:user)
+      organization = insert(:organization)
+      refute update?(user, organization)
+    end
+
+    test "returns false when user is pending member of organization" do
+      [user, organization] = setup_user_organization_by_role("pending")
+      refute update?(user, organization)
+    end
+
+    test "returns false when user is contributor of organization" do
+      [user, organization] = setup_user_organization_by_role("contributor")
+      refute update?(user, organization)
+    end
+
+    test "returns true when user is admin of organization" do
+      [user, organization] = setup_user_organization_by_role("admin")
+      assert update?(user, organization)
+    end
+
+    test "returns true when user is owner of organization" do
+      [user, organization] = setup_user_organization_by_role("owner")
+      assert update?(user, organization)
+    end
+  end
+end
--- a/web/helpers/authentication_helpers.ex
+++ b/web/helpers/authentication_helpers.ex
@@ -28,16 +28,21 @@ defmodule CodeCorps.AuthenticationHelpers do
   # This is partially adjusted code, taken from canary
   def load_and_authorize_changeset(conn, options) do
     action = conn.private.phoenix_action
-
     cond do
-      action in options[:only] -> conn |> load_resource(options) |> do_init_and_authorize_changeset(options[:model], action)
-      true -> conn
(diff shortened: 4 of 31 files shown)


test/policies/category_policy_test.exs
```diff
@@ -0,0 +1,29 @@
+defmodule CodeCorps.CategoryPolicyTest do
+  use CodeCorps.PolicyCase
+
+  import CodeCorps.CategoryPolicy, only: [create?: 1, update?: 1]
+
+  describe "create?" do
+    test "returns true when user is an admin" do
+      user = build(:user, admin: true)
+      assert create?(user) == true
```
sbatson5: Thoughts on just `assert create?(user)` as it looks for a truthy value?

test/policies/category_policy_test.exs
```diff
@@ -0,0 +1,29 @@
...
+    end
+
+    test "returns false if user is not an admin" do
+      user = build(:user, admin: false)
+      assert create?(user) == false
```
sbatson5: Thoughts on `refute create?(user)`?

test/policies/comment_policy_test.exs
```diff
@@ -0,0 +1,42 @@
...
+  describe "create?" do
+    test "returns true if own record" do
+      user = insert(:user)
+      changeset = %Comment{} |> create_changeset(%{user_id: user.id})
+
+      assert create?(user, changeset) == true
+    end
+
+    test "returns false if someone else's record" do
+      user = insert(:user)
```
sbatson5: Thoughts on `[user, another_user] = insert_pair(:user)`?

test/policies/comment_policy_test.exs
```diff
@@ -0,0 +1,42 @@
...
+    test "returns true if own record" do
+      user = insert(:user)
+      comment = insert(:comment, user: user)
+
+      assert update?(user, comment) == true
+    end
+
+    test "returns false if someone else's record" do
+      user = insert(:user)
+      another_user = insert(:user)
```
sbatson5: Same as above `[user, another_user] = insert_pair(:user)`

test/policies/category_policy_test.exs
sbatson5: There seem to be a few tests that explicitly check against true or false. I think we could leverage `assert` versus `refute`

test/policies/organization_policy_test.exs
```diff
@@ -0,0 +1,66 @@
...
+    test "returns false when user is not member of organization" do
+      user = insert(:user)
+      organization = insert(:organization)
+      assert update?(user, organization) == false
+    end
+
+    test "returns false when user is pending member of organization" do
+      user = insert(:user)
+      organization = insert(:organization)
+      insert(:organization_membership, role: "pending", member: user, organization: organization)
```
sbatson5: We seem to have this block repeated in a few places where only the role varies. Thoughts on extracting this to a private method?

```
defp setup_user_organization_by_role(role) do
  ... all the codes but we pass in "pending"
end
```

This saves us a few lines but adds readability to the test so we can know exactly the intention of this block.

web/policies/comment_policy.ex
```diff
@@ -1,8 +1,9 @@
 defmodule CodeCorps.CommentPolicy do
   alias CodeCorps.Comment
   alias CodeCorps.User
+  alias Ecto.Changeset
 
-  def create?(%User{}), do: true
+  def create?(%User{} = user, %Changeset{changes: %{user_id: creator_id}}), do: user.id == creator_id
```
sbatson5, line 6: I would expect a method name with a `?` to to imply it will return a bool. Thoughts on something like `can_create?`? This would probably be out of the context of this PR, but just a thought.

web/policies/organization_policy.ex
```diff
@@ -8,6 +8,7 @@ defmodule CodeCorps.OrganizationPolicy do
   def create?(%User{admin: true}), do: true
   def create?(%User{admin: false}), do: false
 
+  def update?(%User{admin: true}, %Organization{}), do: true
   def update?(%User{} = user, %Organization{} = organization), do: user |> role_is_at_least_admin(organization)
```
sbatson5, line 12: Should there be a catch all?
`def update?(_, _), do: false`

web/policies/preview_policy.ex
```diff
@@ -1,5 +1,7 @@
 defmodule CodeCorps.PreviewPolicy do
   alias CodeCorps.User
 
-  def create?(%User{} = _user), do: true
+  alias Ecto.Changeset
+
+  def create?(%User{} = user, %Changeset{changes: %{user: %{data: creator}}}), do: user.id == creator.id
```
sbatson5: What are we checking here? That the `creator` has changed in our changeset? Not sure I follow what this method is doing.

sbatson5 (commented):
Just a couple of questions. Great to see more test coverage!  In general, it seems we can just leverage `assert` and `refute` for a lot of these tests. 👍

web/policies/comment_policy.ex
joshsmith: @sbatson5 I think this is what canary expects.

web/policies/organization_policy.ex
joshsmith: @sbatson5 I think it will return false if the role's not at least an admin.

web/policies/preview_policy.ex
joshsmith: The preview uses the `current_user` and the changeset assigns that user. We're checking that the creator of the preview is the user trying to create it.

test/policies/organization_policy_test.exs
```diff
@@ -0,0 +1,58 @@
...
+
+    test "returns false when user is not member of organization" do
+      user = insert(:user)
+      organization = insert(:organization)
+      refute update?(user, organization)
+    end
+
+    test "returns false when user is pending member of organization" do
+      [user, organization] = setup_user_organization_by_role("pending")
+      refute update?(user, organization)
```
sbatson5, line 40: 👏 I think these tests read a lot better now! Very clear what we are testing.

web/policies/organization_policy.ex
sbatson5: I guess I wondering if there was ever a chance that we passed something that wasn't a `user` for the first param or `organization` for second. If we called `update?(nil, organization)`, we would get a clause failure.

joshsmith: That should never happen as something else really major would have failed in that event (i.e. there's no current user even in the `conn` to be able to check against.

web/helpers/authentication_helpers.ex
```diff
@@ -51,15 +56,23 @@ defmodule CodeCorps.AuthenticationHelpers do
...
+      |> Module.split
+      |> List.last
+      |> Macro.underscore
+      |> String.to_atom
+
     conn.assigns |> Map.get(resource_name)
   end
   defp get_resource(_conn, model, :create), do: model.__struct__
 
   defp get_changeset_method(action), do: "#{action}_changeset" |> String.to_atom
```
joshsmith, line 70: @begedin this is really constraining and non-user friendly. In many cases we don't use `create_changeset` but just `changeset`, even for the `create`.

web/helpers/authentication_helpers.ex
```diff
@@ -28,16 +28,21 @@ defmodule CodeCorps.AuthenticationHelpers do
...
+        conn
     end
   end
 
   defp do_init_and_authorize_changeset(conn, model, action) do
     changeset = init_changeset(conn, model, action)
-    conn |> assign(:changeset, changeset) |> authorize(changeset, action)
+    conn
+    |> assign(:changeset, changeset)
+    |> authorize(changeset, action)
```
joshsmith, line 45: This works really funky with `Preview` because we're doing an `assign_user` inside of the changeset. When this plug runs before the controller code itself is executed, we don't have the user that's going to be assigned in the controller yet inside the changeset that checks whether we're going to authorize it.

This is a bit 🐔 and 🍳 as a result.

joshsmith:
@begedin I'm pretty blocked on the above by `load_and_authorize_changeset`'s current functionality and have hit my limit on what I can accomplish tonight.

If you can give your thoughts on how we can refactor this to be less brittle, that would be really helpful.

web/helpers/authentication_helpers.ex
begedin: @JoshSmith I was hoping this would establish a convention, much in the same way how canary already uses a convention to load a resource. However, we could extend the code to allow for an option to specify the method name.

begedin: Also, a switch to `authy` would mean we would likely get rid of these and be more direct with which parameters we send into the various `can?` methods.

web/helpers/authentication_helpers.ex
begedin: I'm not really sure how to resolve that, then. 

We could split up `load_and_authorize_changeset` into `load_changeset` and `authorize_changeset`. Canary already does the same with `_resource_`. For that specific action, since it's a `create`, we could load manually, using a locally defined plug method, then authorize using just `authorize_changeset`

joshsmith:
Going to temporarily remove the Preview policy changes from here and address again after we open a separate issue that makes preview changesets work like our others. The `assign_user` that we're doing there is unique and probably should not be.

joshsmith:
Also going to add an issue about adding `ja_resource` before proceeding.

joshsmith:
Latest change requires a change in Ember.

joshsmith:
@sbatson5 @begedin and anyone else interested, this should be ready for a re-review

begedin (approved):
Changes look good. Seems good to go to me 👍

PR #328, opened 2016-10-08
Use json_payload strategy in controllers and eliminate use of build_payload

joshsmith:
@begedin This is a first attempt but do not understand how you had intended to handle relationships or the current user inside of the request.

### DIFF
--- a/test/controllers/category_controller_test.exs
+++ b/test/controllers/category_controller_test.exs
@@ -1,18 +1,16 @@
 defmodule CodeCorps.CategoryControllerTest do
   use CodeCorps.ApiCase
 
-  alias CodeCorps.Category
-
   @valid_attrs %{description: "You want to improve software tools and infrastructure.", name: "Technology"}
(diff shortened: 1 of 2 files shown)


joshsmith:
@begedin the current implementation of `json_payload` is pretty rigid and I can't quite determine how we're supposed to be building relationships or dealing with complex relationship graphs when using this.

I think this, along with `load_and_authorize_changeset`, do require some documentation in the README and in the inline documentation, in addition to any refactoring that happens.

begedin:
Me using the existing `@valid_attrs` in the categories controller test is probably a source of confusion. You should be able to build a payload with relationships, using 

``` Elixir
post = insert(:post)
user = insert(:user)
json_payload(:comment, post: post, user: user)
```

joshsmith:
Going to hit pause on this pending #331.

joshsmith:
Closing because I am not working on this right now.

Issue #322, opened 2016-10-05
Refactor tests in project_category_controller_test


sbatson5:
This repo has great test coverage!  Some of it is slightly redundant at times.  Here we test both the records created and the JSON returned by our API (which are the same):
https://github.com/code-corps/code-corps-api/blob/develop/test/controllers/project_category_controller_test.exs#L97-L103

I think the importance of the controller is to return formatted JSON that our front-end can consume. Can we safely assume that if the payload is correct, then our records have persisted appropriately?

We also have a few instances like this:
https://github.com/code-corps/code-corps-api/blob/develop/test/controllers/project_category_controller_test.exs#L48-L58

I think this is a great candidate for extracting to some sort of helper that is easier to read and reusable by other controller tests

sbatson5:
And this is me volunteering for it too 😛

joshsmith:
@sbatson5 I think that makes sense. It's also important to note that we have view tests. Basically the controller test needs to test not the whole payload, but just that the view renders the expected data back to us.

It really is a matter of isolating what new things have been added to the stack that aren't tested more fully and appropriately at some lower level.

Would be happy to see a PR and evaluate your exact approach here.

sbatson5:
Thanks for the context! I've only worked on a Elixir app that just serves JSON for our front-end, so I've never actually written a view test in Phoenix.

At the very least, I think we can extract some helpers.

sbatson5:
Should be good to close this issue @JoshSmith. Hopefully it counts towards that Hacktoberfest t-shirt 😉

PR #329, opened 2016-10-08
add pipable test helpers for asserting relationship attrs

sbatson5:
Addresses: https://github.com/code-corps/code-corps-api/issues/322

Wanted to add some more test helpers to help with test readability. The goal was to make pipeable assertion methods so we can chain responses and test everything we need in a block.

### DIFF
--- a/test/controllers/organization_memberships_controller_test.exs
+++ b/test/controllers/organization_memberships_controller_test.exs
@@ -23,6 +23,11 @@ defmodule CodeCorps.OrganizationMembershipControllerTest do
     }
   end
 
+  defp assert_role(data, role) do
+    assert data["attributes"]["role"] == role
+    data
+  end
+
   describe "index" do
     test "lists all resources", %{conn: conn} do
       [membership_1, membership_2] = insert_pair(:organization_membership)
@@ -99,14 +104,16 @@ defmodule CodeCorps.OrganizationMembershipControllerTest do
       membership = insert(:organization_membership, role: "admin")
 
       path = conn |> organization_membership_path(:show, membership)
-      data = conn |> get(path) |> json_response(200) |> Map.get("data")
 
-      assert data["id"] == "#{membership.id}"
-      assert data["type"] == "organization-membership"
-
-      assert data["attributes"]["role"] == "admin"
-      assert data["relationships"]["organization"]["data"]["id"] |> String.to_integer == membership.organization_id
-      assert data["relationships"]["member"]["data"]["id"] |> String.to_integer == membership.member_id
+      response =
+        conn
+        |> get(path)
+        |> json_response(200)
+        |> assert_jsonapi_relationship("organization", membership.organization.id)
+        |> assert_jsonapi_relationship("member", membership.member.id)
+        |> Map.get("data")
+        |> assert_result_id(membership.id)
+        |> assert_role("admin")
     end
 
     test "renders page not found when id is nonexistent", %{conn: conn} do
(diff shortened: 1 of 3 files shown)


test/controllers/project_category_controller_test.exs
```diff
@@ -37,25 +37,22 @@ defmodule CodeCorps.ProjectCategoryControllerTest do
...
-      assert second_result["relationships"]["project"]["data"]["type"] == "project"
-      assert second_result["relationships"]["category"]["data"]["id"] == "#{society.id}"
-      assert second_result["relationships"]["category"]["data"]["type"] == "category"
+
+      [first_result, second_result] = response |> Map.get("data")
+
+      first_result
+      |> assert_result_id(project_category_1.id)
+      |> assert_jsonapi_relationship("project", project.id)
+      |> assert_jsonapi_relationship("category", arts.id)
```
sbatson5, line 50: Since this method checks for the type, we shouldn't have to explicitly assert for type.  This method essentially tests the type and id for us.

test/support/test_helpers.ex
```diff
@@ -1,9 +1,25 @@
...
       String.to_integer(attributes["id"])
     end
   end
+
+  def assert_jsonapi_relationship(json = %{"relationships" => relationships}, relationship_name, id) do
+    assert relationships[relationship_name]["data"]["id"] == Integer.to_string(id)
+    json
+  end
+
+  def assert_jsonapi_relationship(json, relationship_name, id) do
```
sbatson5, line 16: Added 2 of these methods. One to be used when passing the relationships itself (above) and one where we can pass the whole response.

sbatson5:
@JoshSmith I think this is in a good place for reviews.

test/support/test_helpers.ex
```diff
@@ -1,9 +1,25 @@
...
+    assert json["data"]["relationships"][relationship_name]["data"]["id"] == Integer.to_string(id)
```
begedin, line 17: Remind me, will this fail if the json is completely invalid, as in, for example, no "data" key at all?

Maybe 3 signaturs would do well. One for just the "data" part, one for the entire payload (this one, but specify the structure, like in the previous method, and one which is just

``` Elixir
def assert_jsonapi_relationship(_, _, _), do: false
```

test/controllers/organization_memberships_controller_test.exs
```diff
@@ -23,6 +23,8 @@ defmodule CodeCorps.OrganizationMembershipControllerTest do
     }
   end
 
+  defp get_role(data), do: data["attributes"]["role"]
```
begedin: Maybe `assert_role(data, role)`, considering the context we use it in?

begedin (changes requested):
Really, really love this idea. @JoshSmith, what do you think? It would make our tests much cleaner and would be a good thing for new contributors to work with.

test/controllers/organization_memberships_controller_test.exs
sbatson5: And move the assertion to this method?  Yeah, I think that makes sense. 👍

test/support/test_helpers.ex
sbatson5: Yes, this should fail. Since we aren't pattern matching on the second method having `assert_jsonapi_relationships(_, _, _)` would be unreachable.  So, this would throw an error if there were no `data` attribute in the `json` list passed.  Which essentially the same as a failure.

test/controllers/organization_memberships_controller_test.exs
sbatson5: Revised in https://github.com/code-corps/code-corps-api/pull/329/commits/120f71d6be94a9955b8e526b7e4bba7bc33d9af6

joshsmith (approved):
👍

joshsmith:
Looks great! Just rebase and squash.

sbatson5:
@JoshSmith rebased and squashed 👍

PR #333, opened 2016-10-09
Add tests for user role index and show

### LINKED ISSUE #222: Add tests for UserRoleController index and show
We should add tests for the UserRoleController index and show actions. Use UserSkillController as example for these tests.

joshsmith:
Adds tests for index and show actions.

Closes #222


filipecrosk (approved):
it looks fine

PR #352, opened 2016-10-14
Switch to bodyguard - code spike

### LINKED ISSUE #181: Switch from canary to bodyguard
Another library came to my attention at ElixirConf that looks a lot more explicit in its implementation than `canary` and maps closer to `Pundit` than `cancancan`. It's named [`bodyguard`](https://github.com/schrockwell/bodyguard).

We should do a PR to switch from one to the other. Should be relatively trivial considering the similarity of `can?/3`.

begedin:
Might close #181 if we get it to a mergeable state

This is a spike to make a switch from `canary` to `bodyguard` on the categories endpoint.

I really don't like the outcome here and I had several difficulties implementing this.

I'll do a full write up first chance I get. 

For now, I should mention that I did this PR on top of #351, so that one ought to be merged first.


joshsmith:
@begedin the only change in here that I saw on top of #351 was removing ModelHelpers, so I'm guessing this has to be reopened separately.

PR #355, opened 2016-10-15
Update arc

joshsmith:
This upgrades arc to a release candidate.

Not sure if we should place in production or not, but it's a thought!


joshsmith:
Well, this certainly looks blocked.

Would like to come back and update as arc gets updated.

PR #356, opened 2016-10-15
Code spike - switch to bodyguard

### LINKED ISSUE #181: Switch from canary to bodyguard

begedin:
Missed a commit last time. This would close #181 if we decide to actually go with it.

The PR should demonstrate the issues I had and I still haven't quite figured out how to deal with authorizing delete requests. Perhaps an authorize call within the `record` callback would work.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -64,6 +64,7 @@ defmodule CodeCorps.Mixfile do
       {:arc, git: "https://github.com/stavro/arc.git", ref: "354d4d2e1b86bcd6285db3528118fe3f5db36cf5", override: true}, # Photo uploads
       {:arc_ecto, "~> 0.4.4"},
       {:canary, "~> 1.0"}, # Authorization
+      {:bodyguard, "~> 0.2.0"}, # Authorization
       {:comeonin, "~> 2.0"},
       {:corsica, "~> 0.4"}, # CORS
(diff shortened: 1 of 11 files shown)


joshsmith:
@begedin what are we waiting on in this PR?

begedin:
Really, for you to take a look and for us to decide if we should switch. As you can see, bodyguard has issues with having a nice integration just as canary does, albeit, to a lesser degree.

Authorizing a delete request, which this PR doesn't do yet, would cause further trouble and I still need to figure that out. Really, this needs a second pass. It can be done, but I need to figure out a really nice way to do it. Something that would allow us to just call `authorize` within a method chain and let everything "just work", meaning that, if authorization succeeds, the chain continues, and if it fails, the chain stops.

Let me know and I'll refocus my efforts from stripe into this.

We do need to tackle it at some point, since with canary, we still have that whole "loading a resource twice" issue.

joshsmith:
We can reprioritize post Stripe. Just wanted to know current status. Maybe mark low priority.

Issue #350, opened 2016-10-13
Test Segment Analytics tracking


joshsmith:
We reworked our analytics tracking in #332 but we need to figure out how to make this tracking testable.

vidmantas:
The infrastructure for this is merged https://github.com/code-corps/code-corps-api/commit/593f6599c494d0f6cf994e6a2522f4179779dd11 the rest of the work is to apply this for other tests where Segment is skipped to be tested.

PR #357, opened 2016-10-16
Test Segment tracking after user creation

vidmantas:
Tries to implement segment tracking testing requested in https://github.com/code-corps/code-corps-api/issues/350

Used some ideas from http://blog.plataformatec.com.br/2015/10/mocks-and-explicit-contracts/

I don't think that message sending to `self` is useful in development, so I've created a separate implementation to be plugged in for testing only.

### DIFF
--- a/config/test.exs
+++ b/config/test.exs
@@ -28,7 +28,7 @@ config :code_corps, allowed_origins: ["http://localhost:4200"]
 config :guardian, Guardian,
   secret_key: "e62fb6e2746f6b1bf8b5b735ba816c2eae1d5d76e64f18f3fc647e308b0c159e"
 
-config :code_corps, :analytics, CodeCorps.Analytics.InMemoryAPI
+config :code_corps, :analytics, CodeCorps.Analytics.TestAPI
 
 config :code_corps, :icon_color_generator, CodeCorps.RandomIconColor.TestGenerator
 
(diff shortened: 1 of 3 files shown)


joshsmith (approved):
This is fantastic. Good work @vidmantas!

joshsmith:
Did you want to merge this now or test the other calls in the context of this PR? I'd be fine with either.

begedin:
This is what I imagined we'd have. Great work, @vidmantas

vidmantas:
I've added a test for another users controller method, however before proceeding with other occurrences of Segment tracking wanted to raise a question on organizing tests: 

now we have pretty big chunks of repetitive "setup" part before assertions on some `create` and `update` test cases (e.g. when valid data is passed, lines 75-83, 95-102) . Let's try to organize them using named setups? If you want, I can provide a separate PR as a suggestion (in this case I think it would make sense to merge this one and continue for the rest of Segment calls as separate PR after there's decision on this question)

begedin:
@vidmantas We already had a few attempts at reducing test boilerplate, especially with creating records/payloads, but weren't happy with the results. Feel free to make a PR with your suggestion. I have a feeling it will be well received.

As for this, we can consider it ready, so squash/rebase and merge when you get the chance. If you also don't mind, keep the referrenced issue open and add a comment saying what else is needed (as in "the infrastructure for this has been merged, now it's just a matter of writing the individual tests."), with maybe links to the relevant code.

joshsmith:
@vidmantas I second what @begedin said!

vidmantas:
@begedin @JoshSmith alright. I just don't have write access to merge, so please do that instead of me and I'll proceed as discussed ;)

joshsmith:
Can you squash? Then we can merge.

vidmantas:
@JoshSmith ah, right, sorry. Done now :)

joshsmith:
🙌 !

PR #358, opened 2016-10-17
add test helper for building relationships

sbatson5:
# What's in this PR?

There is a lot of repeated logic in our controller tests for how we build relationships. We only have a few models that relate to other models, but they seem to be used in quite a few tests. This PR should extract it to the test helpers and (hopefully) make it flexible.
## References

Fixes No existing issues for this yet.

### DIFF
--- a/test/support/test_helpers.ex
+++ b/test/support/test_helpers.ex
@@ -22,4 +22,31 @@ defmodule CodeCorps.TestHelpers do
     assert String.to_integer(result["id"]) == id
     result
   end
+
+  def put_id(payload, id), do: put_in(payload, ["data", "id"], id)
+  def put_attributes(payload, attributes), do: put_in(payload, ["data", "attributes"], attributes)
+  def put_relationships(payload, record_1, record_2), do: put_relationships(payload, [record_1, record_2])
+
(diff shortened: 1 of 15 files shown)


test/support/test_helpers.ex
```diff
@@ -22,4 +22,31 @@ defmodule CodeCorps.TestHelpers do
     assert String.to_integer(result["id"]) == id
     result
   end
+
+  def put_id(payload, id), do: put_in(payload, ["data", "id"], id)
+  def put_attributes(payload, attributes), do: put_in(payload, ["data", "attributes"], attributes)
+
+  def put_relationships(payload, record_1, record_2), do: put_relationships(payload, [record_1, record_2])
```
BlakeWilliams, line 28: Super minor nitpick, but since this has different arity than the function below, should this be grouped up above with the other one-liners?

BlakeWilliams (commented):
One minor thing, but nice extraction!

sbatson5:
@JoshSmith rebased and squashed. Any chance this could get a 👍 ?

joshsmith (approved):

joshsmith:
🙌

Issue #341, opened 2016-10-10
Determine if organizations/:organization_id/memberships is necessary


joshsmith:
We have a `/organizations/:organization_id/memberships` that our Ember app does not appear to be hitting in tests (is not in the Mirage config), but which the API does serve.

We should determine if this route is strictly necessary, and if not, remove it.

begedin:
Looked into this:

First of all, ember fetches memberships using ID coalescing, so technically, we do not need this endpoint, nor do we need an `organization_filter` on the `OrganizationMembership.index_filters`.

That means that when we fetch an organization, it will contain a list of membership ids. Later, when needed, those will be fetched via coalesced IDs, same as any other `hasMany` relationship.

Second, I was confused as to how this even works at all, since there's only one index action in the membership controller and both `:org_id/membership` and `org_memberships/`target it.

Turns out, `some_organization_id/memberships` and `organization_memberships?organization_id=some_organization_id` is functionally equivalent. The dynamic route part get's mapped as `organization_id` and merged into the `params` map.

To explain by example, we had the following test 

``` Elixir
test "lists all resources for specified organization", %{conn: conn} do
  organization = insert(:organization)
  [membership_1, membership_2] = insert_pair(:organization_membership, organization: organization)
  insert(:organization_membership)

  path = conn |> organization_organization_membership_path(:index, organization)
  response = conn |> get(path) |> json_response(200)

  assert ids_from_response(response) == [membership_1.id, membership_2.id]
end
```

To that one, I added the following test:

``` Elixir
test "lists all resources for specified organization by standard path", %{conn: conn} do
  organization = insert(:organization)
  [membership_1, membership_2] = insert_pair(:organization_membership, organization: organization)
  insert(:organization_membership)

  path = conn |> organization_membership_path(:index, %{organization_id: organization.id})
  response = conn |> get(path) |> json_response(200)

  assert ids_from_response(response) == [membership_1.id, membership_2.id]
end
```

They both work just fine. The difference is just the path.

The conclusion would be, yes, we should be able to safely remove the `GET /:org_id/memberships`. We should, however, also be able to remove the `organization_filter` from `CodeCorps.OrganizationMembership.index_filters` as well as  from `CodeCorps.ModelHelpers` if that's the only place it's used in.

joshsmith:
@begedin should this still be open?

begedin:
It should not. Not sure why It didn't close.

PR #359, opened 2016-10-17
Remove /:organization_id/memberships path

begedin:
As discussed in #341, it's not used at all.

I would like to look at other things here as well. We have tests for filtering by role level and organization id that are written as if they need to come in specific combinations. Migth be simpler to just loosen them up and simply apply supported filters "in order".

### DIFF
--- a/lib/code_corps/helpers/query.ex
+++ b/lib/code_corps/helpers/query.ex
@@ -7,15 +7,6 @@ defmodule CodeCorps.Helpers.Query do
     query |> where([object], object.id in ^ids)
   end
 
-  def organization_filter(query, organization_id) do
-    query |> where([object], object.organization_id == ^organization_id)
-  end
-
-  def role_filter(query, roles_list) do
-    roles = roles_list |> coalesce_string
-    query |> where([object], object.role in ^roles)
-  end
-
   # skill queries
 
   def limit_filter(query, %{"limit" => count}) do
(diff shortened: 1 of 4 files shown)


joshsmith:
@begedin so not ready for review yet?

begedin:
@JoshSmith I removed everything that's unused by ember in a separate commit, so we can roll it back if you don't feel comfortable doing this. However, here's how our org memebrship mirage config looks like:

``` Javascript
/**
* Organization memberships
*/

// GET /organization-memberships
this.get('/organization-memberships', { coalesce: true });

// POST /organization-memberships
this.post('/organization-memberships');

// GET /organization-memberships/:id
this.get('/organization-memberships/:id');

// PATCH /organization-memberships/:id
this.patch('/organization-memberships/:id');

// DELETE /organization-memberships/:id
this.delete('/organization-memberships/:id');
```

We do not have any requirements for filtering, ember side.

When we retrieve an organization, it will provide us a list of membership ids. These ids will then be fetched using coalesced requests.

An average organization will not have members numbering to a point where it would be a problem to get them all and then sort/filter client-side. If we do need additional filters, we can add them later.

If anything, what we should consider is preventing an unfiltered index-request (without a list of ids), or at least paginate or limit that response, or something.

joshsmith (approved):

PR #375, opened 2016-10-19
Remove duplication from user controller tests

vidmantas:
My try to organize user controller tests to avoid duplication of actual request.

Is there any other way to group tests in ExUnit? Some kind of "scopes" or "contexts"?

### DIFF
--- a/test/controllers/user_controller_test.exs
+++ b/test/controllers/user_controller_test.exs
@@ -70,10 +70,10 @@ defmodule CodeCorps.UserControllerTest do
     end
   end
 
-  describe "create" do
-    test "creates and renders resource when data is valid", %{conn: conn} do
+  describe "create user with valid data" do
+    setup context do
       attrs = Map.put(@valid_attrs, :password, "password")
-      conn = post conn, user_path(conn, :create), %{
+      conn = post context.conn, user_path(context.conn, :create), %{
         "meta" => %{},
         "data" => %{
           "type" => "user",
@@ -82,6 +82,10 @@ defmodule CodeCorps.UserControllerTest do
         }
       }
 
+      {:ok, conn: conn}
+    end
+
+    test "creates and renders resource", %{conn: conn} do
       id = json_response(conn, 201)["data"]["id"]
       assert id
       user = Repo.get(User, id)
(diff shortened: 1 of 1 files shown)


begedin:
@vidmantas There are named setups which we could use, so you can look into that if you have the time.

vidmantas:
@begedin yeah I was thinking about named setups, but in this situation, I don't see much improvement of moving that code out (as there're no combinations of multiple setups required)
@JoshSmith wdyt?

joshsmith:
@vidmantas yeah not sure it's necessary here. Do you feel good with this being reviewed now?

vidmantas:
@JoshSmith yes please

joshsmith:
@vidmantas instead of hitting the merge button here, I think better to `git rebase`.

joshsmith:
@begedin would you mind providing a review here?

vidmantas:
@JoshSmith tried that for the first time, somehow thought it's going to do a silent rebase, sorry! Will rebase to 1 commit after we settle down.

test/controllers/user_controller_test.exs
```diff
@@ -91,20 +95,14 @@ defmodule CodeCorps.UserControllerTest do
...
-          "type" => "user",
-          "attributes" => Map.put(@valid_attrs, :password, "password"),
-          "relationships" => relationships
-        }
-      }
-      id = json_response(conn, 201)["data"]["id"] |> String.to_integer
-      assert_received {:track, ^id, "Signed Up", %{}}
+    test "calls segment tracking", %{conn: conn} do
+      id = json_response(conn, 201)["data"]["id"]
+      assert_received {:track, id, "Signed Up", %{}}
```
begedin, line 100: My guess is, the compiler is telling you the `variable id is defined but not used`.  The reason for that is that this call actually does not check if the second tupple element is equal to `id`. Instead, it redefines the `id` variable altogether.

The correct call would be 

``` Elixir
assert_received {:track, ^id, "Signed Up", %{}}
```

My guess is, you took this from another part of our code. Sadly, it was wrong in our code and I since corrected it. Sorry about that.

begedin (commented):
This is a nice amount of deduplication and would be something I would be more than glad to merge after the minor correction I commented on.

However, while working on #378 , I've developed a way to add simple helpers for creation, index, show, update, delete that auto-build our payloads.

This will very likely cause a conflict and right now, I'm really not sure what the better approach is.

@JoshSmith @vidmantas Please take a look at https://github.com/code-corps/code-corps-api/blob/e1649da303eafbb9bb1bd810820c57cba971d73f/test/support/api_case.ex and tell me what do you think. 

Also, take a look at https://github.com/code-corps/code-corps-api/blob/e1649da303eafbb9bb1bd810820c57cba971d73f/test/controllers/donation_goal_controller_test.exs to see how these new helpers would be used.

joshsmith:
@vidmantas just pinging again her to check on status of things.

vidmantas:
@JoshSmith sorry it's a difficult week for me - I promise to return and give a look later this week, latest the weekend.

joshsmith:
@vidmantas no worries! Just wanted to see if we were blocking you.

vidmantas:
@begedin I've reviewed your changes and it definitely makes sense to keep & go your way.
I can rebase and rework using new helpers, but it will happen next week only. If that's fine for you?

joshsmith:
@vidmantas we're going to be doing a lot of controller refactoring at the beginning of next week, I believe.

joshsmith:
This is really out of date with the controller refactoring we did. Sorry we got so out of sync. 😞

Issue #368, opened 2016-10-17
Create StripeCustomer model


joshsmith:
For #258 

Customer for the platform (to be copied to the connect account) and customer for the standalone account.

https://stripe.com/docs/api#customer_object

ATTRIBUTES
- `id string`
- `created timestamp`
- `currency string` The currency the customer can be charged in for recurring billing purposes.
- `default_source string` ID of the default source attached to this customer.
- `delinquent boolean` Whether or not the latest charge for the customer’s latest invoice has failed
- `email string`

dmvt:
What do you mean by 'copied to the connected account'?

joshsmith:
@dmvt https://support.stripe.com/questions/connect-publishable-key-error-with-shared-customers

joshsmith:
Closed by #377.

PR #377, opened 2016-10-19
Add StripeCard and StripeCustomer models

### LINKED ISSUE #363: Add StripeCard model
For #258 

See https://stripe.com/docs/api#card_object

ATTRIBUTES
- `id string` ID of card (used in conjunction with a customer or recipient ID)
- `brand string` Card brand. Can be Visa, American Express, MasterCard, Discover, JCB, Diners Club, or Unknown.
- `customer string` The customer that this card belongs to. This attribute will not be in the card object if the card belongs to an account or recipient instead.
- `cvc_check string` If a CVC was provided, results of the check: pass, fail, unavailable, or unchecked
- `exp_month integer`
- `exp_year integer`
- `last4 string`
- `name string` Cardholder name

begedin:
Will close #363 and #368 

Since `StripeCard#customer` and `StripeCustomer#default_source` in the original stripe payloads are clearly actually relationships between the two models (meaning a stripe customer has one default source/card and many individual cards, while a stripe card belongs to a stripe customer), I decided to create models with those relationships. 

While having string ids was relatively simple, I didn't want to complicate things too much with the foreign ids, so I just kept things explicit by defining adapters that convert between a stripe payload and a param map to be used in a changeset.

I'm not sure how the two models should relate to existing models. Should a project belong to / have one stripe customer. Should it be a user or an organization instead? I'm guessing organization, but I'm not sure how multiple projects per organization will be handled in that case.

### DIFF
--- a/lib/code_corps/stripe/adapters/stripe_card.ex
+++ b/lib/code_corps/stripe/adapters/stripe_card.ex
@@ -1,0 +1,9 @@
+defmodule CodeCorps.Stripe.Adapters.StripeCard do
+  import CodeCorps.MapUtils, only: [rename: 3]
+
+  def params_from_stripe(%{} = stripe_map) do
+    stripe_map
+    |> rename("id", "id_from_stripe")
+    |> rename("customer", "customer_id_from_stripe")
+  end
+end
--- a/lib/code_corps/stripe/adapters/stripe_customer.ex
+++ b/lib/code_corps/stripe/adapters/stripe_customer.ex
@@ -1,0 +1,7 @@
+defmodule CodeCorps.Stripe.Adapters.StripeCustomer do
+  import CodeCorps.MapUtils, only: [rename: 3]
+
+  def params_from_stripe(%{} = stripe_map) do
+    stripe_map |> rename("id", "id_from_stripe")
+  end
+end
--- a/priv/repo/migrations/20161019090945_add_stripe_customers_cards_tables.exs
+++ b/priv/repo/migrations/20161019090945_add_stripe_customers_cards_tables.exs
@@ -1,0 +1,38 @@
+defmodule CodeCorps.Repo.Migrations.AddStripeCustomersCardsTables do
+  use Ecto.Migration
+
+  def change do
+    create table(:stripe_customers) do
+      add :created, :datetime
+      add :currency, :string
+      add :delinquent, :boolean
+      add :email, :string
+      add :id_from_stripe, :string, null: false
+
+      add :user_id, references(:users, on_delete: :nothing), null: false
+
+      timestamps()
+    end
+
+    create unique_index(:stripe_customers, [:id_from_stripe])
+    create unique_index(:stripe_customers, [:user_id])
+
+    create table(:stripe_cards) do
+      add :brand, :string
+      add :customer_id_from_stripe, :string
+      add :cvc_check, :string
+      add :exp_month, :integer
+      add :exp_year, :integer
+      add :id_from_stripe, :string, null: false
+      add :last4, :string
+      add :name, :string
+
+      add :user_id, references(:users, on_delete: :nothing), null: false
+
+      timestamps()
+    end
+
+    create index(:stripe_cards, [:user_id])
+    create unique_index(:stripe_cards, [:id_from_stripe])
+  end
+end
--- a/test/models/stripe_card_test.exs
+++ b/test/models/stripe_card_test.exs
@@ -1,0 +1,41 @@
+defmodule CodeCorps.StripeCardTest do
+  use CodeCorps.ModelCase
+
+  alias CodeCorps.StripeCard
+
+  @valid_attrs %{
+    id_from_stripe: "abc123"
+  }
+
+  @invalid_attrs %{}
+
+  describe "create_changeset/2" do
+    test "reports as valid when attributes are valid" do
+      user_id = insert(:user).id
+
+      changes = Map.merge(@valid_attrs, %{user_id: user_id})
+      changeset = StripeCard.create_changeset(%StripeCard{}, changes)
+      assert changeset.valid?
+    end
+
+    test "reports as invalid when attributes are invalid" do
+      changeset = StripeCard.create_changeset(%StripeCard{}, @invalid_attrs)
+      refute changeset.valid?
+
+      assert changeset.errors[:id_from_stripe] == {"can't be blank", []}
+      assert changeset.errors[:user_id] == {"can't be blank", []}
+    end
+
+    test "ensures associations link to records that exist" do
+      attrs =  @valid_attrs |> Map.merge(%{user_id: -1})
+
+      { result, changeset } =
+        StripeCard.create_changeset(%StripeCard{}, attrs)
+        |> Repo.insert
+
+      assert result == :error
+      refute changeset.valid?
+      assert changeset.errors[:user] == {"does not exist", []}
+    end
+  end
+end
--- a/test/models/stripe_customer_test.exs
+++ b/test/models/stripe_customer_test.exs
@@ -1,0 +1,41 @@
+defmodule CodeCorps.StripeCustomerTest do
+  use CodeCorps.ModelCase
+
+  alias CodeCorps.StripeCustomer
+
+  @valid_attrs %{
+    id_from_stripe: "abc123"
+  }
+
+  @invalid_attrs %{}
+
+  describe "create_changeset/2" do
+    test "reports as valid when attributes are valid" do
+      user_id = insert(:user).id
+
+      changes = Map.merge(@valid_attrs, %{user_id: user_id})
+      changeset = StripeCustomer.create_changeset(%StripeCustomer{}, changes)
+      assert changeset.valid?
+    end
+
+    test "reports as invalid when attributes are invalid" do
+      changeset = StripeCustomer.create_changeset(%StripeCustomer{}, @invalid_attrs)
+      refute changeset.valid?
+
+      assert changeset.errors[:id_from_stripe] == {"can't be blank", []}
+      assert changeset.errors[:user_id] == {"can't be blank", []}
+    end
+
+    test "ensures associations link to records that exist" do
+      attrs =  @valid_attrs |> Map.merge(%{user_id: -1})
+
+      { result, changeset } =
+        StripeCustomer.create_changeset(%StripeCustomer{}, attrs)
+        |> Repo.insert
+
+      assert result == :error
+      refute changeset.valid?
+      assert changeset.errors[:user] == {"does not exist", []}
+    end
+  end
+end
--- a/web/models/stripe_card.ex
+++ b/web/models/stripe_card.ex
@@ -1,0 +1,25 @@
+defmodule CodeCorps.StripeCard do
+  use CodeCorps.Web, :model
+
+  schema "stripe_cards" do
+    field :brand, :string
+    field :customer_id_from_stripe, :string
+    field :cvc_check, :string
+    field :exp_month, :integer
+    field :exp_year, :integer
+    field :id_from_stripe, :string, null: false
+    field :last4, :string
+    field :name, :string
+
+    belongs_to :user, CodeCorps.User
+
+    timestamps()
+  end
+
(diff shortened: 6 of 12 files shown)


joshsmith:
A StripeCard would belong to a User and StripeCustomer. A StripeCustomer would belong to a User.

begedin:
I'm guessing a StripeCard would belong to a user **through** a StripeCustomer?

joshsmith:
@begedin that's right, if we even need to do that. It's possible we don't at first.

begedin:
@JoshSmith This ought to be ready for review.

We have a `StripeCustomer` and a `StripeCard`.

# `StripeCustomer`
- `belongs_to :default_card`, which is a `StripeCard`, the id is stored on the customer record, so can't be a `has_one`
- `has_many :cards`, which are also `StripeCard` records
- `belongs_to` a `User` record as well. Again, id is stored here, not on the user record.
- Did not add a relationship to the `User` schema yet. We can add it later, if it's convenient. Didn't want to do work we might not need.

# `StripeCard`
- `belongs_to :customer`, which is a `StripeCustomer`
- Didn't add the `belongs_to :user, through: :customer`, since, as you said, we might not end up needing it
- Added adapter which converts a stripe payload that contains a `StripeCustomer`, into a struct which we can use to insert a `StripeCustomer` record into our database. Reasoning here is that a `StripeCustomer` payload has a `default_source` field, which is a foreign key into `StripeCard`, but our convention would be to call it a `default_card_id`, or `default_source_id`. My choice was `default_card_id`. The adapter basically renames the key.
- Added adapter which converts a stripe payload that contains a `StripeCard`, into a struct which we can use to insert a `StripeCard` record into our database. Reasoning here is that a `StripeCard` payload has a `customer` field, which is a foreign key into `StripeCustomer`, but our convention would be to call it a `customer_id`. The adapter basically renames a key.

joshsmith:
Why did coverage decrease here?

begedin:
@JoshSmith No idea. Looking at the report, coverage got up or stayed the same on all affected files.

priv/repo/migrations/20161019090945_add_stripe_customers_cards_tables.exs
```diff
@@ -2,39 +2,37 @@ defmodule CodeCorps.Repo.Migrations.AddStripeCustomersCardsTables do
...
-      add :currency, :string, null: false, default: "usd"
-      add :delinquent, :boolean, null: false, default: false
-      add :email, :string, null: false
-      add :user_id, references(:users)
+      add :currency, :string
+      add :delinquent, :boolean
+      add :email, :string
+      add :id_from_stripe, :string, null: false
+
+      add :user_id, references(:users, on_delete: :nothing), null: false
```
joshsmith, line 12: Changed the id, set only pieces that should be null as null, and added a user relationship.

priv/repo/migrations/20161019090945_add_stripe_customers_cards_tables.exs
```diff
@@ -2,39 +2,37 @@ defmodule CodeCorps.Repo.Migrations.AddStripeCustomersCardsTables do
...
 
       timestamps()
     end
 
-    create unique_index(:stripe_customers, [:id])
+    create unique_index(:stripe_customers, [:id_from_stripe])
+    create unique_index(:stripe_customers, [:user_id])
```
joshsmith, line 18: Made customers unique to each user.

priv/repo/migrations/20161019090945_add_stripe_customers_cards_tables.exs
```diff
@@ -2,39 +2,37 @@ defmodule CodeCorps.Repo.Migrations.AddStripeCustomersCardsTables do
...
-      add :exp_year, :integer, null: false
-      add :last4, :string, null: false
-      add :name, :string, null: false
+      add :stripe_customer_id, references(:stripe_customers, on_delete: :nothing), null: false
 
       timestamps()
     end
 
-    create unique_index(:stripe_cards, [:id])
+    create index(:stripe_cards, [:stripe_customer_id])
```
joshsmith: Changed what needs to be null, added stripe customer id relationship.

test/models/stripe_card_test.exs
```diff
@@ -4,10 +4,24 @@ defmodule CodeCorps.StripeCardTest do
...
+    changeset = StripeCard.create_changeset(%StripeCard{}, changes)
+    assert changeset.valid?
+  end
+
+  test "changeset with invalid attributes" do
+    changeset = StripeCard.create_changeset(%StripeCard{}, @invalid_attrs)
+    refute changeset.valid?
+
+    assert changeset.errors[:id_from_stripe] == {"can't be blank", []}
+    assert changeset.errors[:stripe_customer_id] == {"can't be blank", []}
```
joshsmith: Added tests for changesets here.

test/models/stripe_customer_test.exs
```diff
@@ -4,14 +4,24 @@ defmodule CodeCorps.StripeCustomerTest do
...
+    changeset = StripeCustomer.create_changeset(%StripeCustomer{}, changes)
+    assert changeset.valid?
+  end
+
+  test "changeset with invalid attributes" do
+    changeset = StripeCustomer.create_changeset(%StripeCustomer{}, @invalid_attrs)
+    refute changeset.valid?
+
+    assert changeset.errors[:id_from_stripe] == {"can't be blank", []}
+    assert changeset.errors[:user_id] == {"can't be blank", []}
```
joshsmith: Added tests for changesets.

web/models/stripe_card.ex
```diff
@@ -1,17 +1,24 @@
...
+    belongs_to :stripe_customer, CodeCorps.StripeCustomer
 
     timestamps()
   end
+
+  def create_changeset(struct, params \\ %{}) do
+    struct
+    |> cast(params, [:id_from_stripe, :stripe_customer_id])
+    |> validate_required([:id_from_stripe, :stripe_customer_id])
+    |> assoc_constraint(:stripe_customer)
```
joshsmith: There should be some field and relationship validation here.

web/models/stripe_customer.ex
```diff
@@ -1,18 +1,24 @@
...
+    has_many :stripe_cards, CodeCorps.StripeCard, foreign_key: :stripe_customer_id
 
     timestamps()
   end
+
+  def create_changeset(struct, params \\ %{}) do
+    struct
+    |> cast(params, [:id_from_stripe, :user_id])
+    |> validate_required([:id_from_stripe, :user_id])
+    |> assoc_constraint(:user)
```
joshsmith, line 22: There should be some field and relationship validation here, as well.

joshsmith (commented):
A bunch of changes made in my last commit.

joshsmith:
I've also removed the default card relationship temporarily. We should re-discuss the best way to implement.

joshsmith:
The adapter is going to be very out of date as a result of these changes.

begedin:
@JoshSmith Removing the `default_card` relationship means we're going away from the official stripe model, where the `default_source` relationship does in fact exist. If you're ok with that, then I am to, but I thought I should make sure you're aware.

I also see that you did add the `id_from_stripe` fields. Again, I'm ok with that change as well. I wasn't sure which one to go with in the first place, so I figured I'd do the (marginally) more difficult one, because it would be easier for someone else to switch to this one if needed.

As for the changesets, I'm not sure this PR needs them at all. I mean, we're not using them anywhere yet and it just may be that these records wont get created from user input at all, and will instead be created automatically by the system, in which case, I'm not really sure validation will be neccesssary. What do you think about that?

begedin:
@JoshSmith Your changes look good from my side. I updated the tests and adapters, so once you review and approve my updates, this should be good to merge.

joshsmith:
I still think a bare minimum validation is necessary and that updates from Stripe should be handled differently than creates.

lib/code_corps/stripe/adapters/stripe_customer.ex
```diff
@@ -2,10 +2,10 @@ defmodule CodeCorps.Stripe.Adapters.StripeCustomer do
...
 
   def params_from_stripe(%{} = stripe_map) do
-    stripe_map |> rename("default_source", "default_card_id")
+    stripe_map |> rename("id", "id_from_stripe")
   end
 
   def params_to_stripe(%{} = ecto_map) do
-    ecto_map |> rename("default_card_id", "default_source")
+    ecto_map |> rename("id_from_stripe", "id")
   end
```
joshsmith: Needs removed.

lib/code_corps/stripe/adapters/stripe_card.ex
```diff
@@ -2,10 +2,14 @@ defmodule CodeCorps.Stripe.Adapters.StripeCard do
...
+    |> rename("id", "id_from_stripe")
+    |> rename("customer", "stripe_customer_id")
   end
 
   def params_to_stripe(%{} = ecto_map) do
-    ecto_map |> rename("customer_id", "customer")
+    ecto_map
+    |> rename("id_from_stripe", "id")
+    |> rename("stripe_customer_id", "customer")
   end
```
joshsmith, line 8: Needs removed.

joshsmith (commented):
Some changes we need here.

lib/code_corps/stripe/adapters/stripe_card.ex
```diff
@@ -0,0 +1,9 @@
+defmodule CodeCorps.Stripe.Adapters.StripeCard do
+  import CodeCorps.MapUtils, only: [rename: 3]
+
+  def params_from_stripe(%{} = stripe_map) do
+    stripe_map
+    |> rename("id", "id_from_stripe")
+    |> rename("customer", "stripe_customer_id")
```
joshsmith: This is not the right thing we're referring to here. We should probably remove.

joshsmith: It looks like we may want this to be the `customer_id_from_stripe`, if we want that at all.

joshsmith: Also, it looks like `customer` could really vary here. Sometimes it may be `null`. We probably don't want to be attempting anything really crazy.

begedin: On the stripe API model, `customer` is basically a foreign key, that can be null.

With my initial approach, were we simply used the original stripe object ids as string primary keys in our own models, this was just a matter of renaming `customer`, to `stripe_customer_id`, since it was an Id and we had a `belongs_to :stripe_customer` on a card model.

Now that we store stripe ids as separate fields, if we want to keep the relationship and do it somewhat straightforward, I could do one of several approaches
- Enforce a foreign key constraint from this field, into the `id_from_stripe` field on the `stripe_customers` table. I believe it is possible with ecto and postgres and should not be too difficult. We can also further rename this `customer` field into `customer_id_from_stripe` or something, to make it a bit cleaer.
- Keep the default `belongs_to :stripe_customer` and rewrite our changeset for stripe card to fetch the stripe customer by `id_from_stripe` and then assign it to the belongs_to.
- We could also just have both `stripe_customers` and `stripe_cards` simply belong to a user or an organization, the way a slugged route does . That might even make the most sense, since in the stripe API, a stripe card can belong either to a stripe customer, which belongs to a user, or to a stripe account, which belongs to a organization.

In any case, the field rename as is cannot stay around, so I'll remove it for now.

joshsmith:
@begedin let's make StripeCard simply belong to a User and not a StripeCustomer. That makes more sense to me in our implementation. But would also add the customer_id_from_stripe I think.

begedin:
@JoshSmith Should it also belong to an organization in that case?

As I said, in stripe API, a `stripe_card` can belong to a `stripe_customer`, or a `stripe_account`. Since in our case, a `stripe_customer` belongs to a user and a `stripe_account` to an organization, it would make sense.

joshsmith:
As for now we won't implement an organization making payments with a card.

begedin:
Made the changes. Coverage decrease does't make much sense to me. The files that supposedly dropped in coverage, I'm not sure how to raise it for.

joshsmith (approved):

PR #379, opened 2016-10-19
Add Stripe oauth workflow

### LINKED ISSUE #259: Add Stripe Connect OAuth workflow
Following these docs: https://hexdocs.pm/stripity_stripe/Stripe.Connect.html

We want to:

> retrieve the oauth access token or the full response, using the code received from the oauth flow return

We'll need to do `generate_button_url(csrf_token)` to yield a button for the account.

When the callback is completed, we'll need to do `oauth_token_callback(code)`.

This will involve an API endpoint for finishing the OAuth token flow, receiving the code back from the Ember front-end.

For #10.

joshsmith:
Closes #259 

Work in progress right now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 /*.ez
 /tmp
 /doc
+/bench
 
 # Generated on crash by the VM
 erl_crash.dump
--- a/blueprint/api.apib
+++ b/blueprint/api.apib
@@ -796,6 +796,26 @@ This resource identifies a relationship between a Project and a Skill. For examp
 
     + Attributes (Project Response)
 
+### Get Stripe authorization [GET /projects/{id}/stripe-auth]
+
+This endpoint allows you to get a Stripe Connect authorization URL used for creating a Stripe Connect account for the given project.
+
+See https://stripe.com/docs/connect/reference#get-authorize
+
+The URL includes a `state` CSRF token which is a Guardian generated JSON Web Token which contains the project's ID. This JWT will be used when the user returns to Code Corps in order to associate Stripe's `code` with the project that initiated the Stripe Connect authorization.
+
++ Parameters
+
+    + id: 1 (integer, required)
+
++ Response 200 (application/vnd.api+json; charset=utf-8)
+
+    + Attributes (Stripe Auth Response)
+
++ Response 401 (application/vnd.api+json; charset=utf-8)
+
+    + Attributes (JSON Web Token Invalid Response)
+
 ### Update a project [PATCH]
 
 + Request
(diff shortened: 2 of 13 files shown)


test/controllers/project_controller_test.exs
```diff
@@ -256,4 +256,23 @@ defmodule CodeCorps.ProjectControllerTest do
...
+      conn = get conn, project_path(conn, :stripe_button_url, project)
+      assert json_response(conn, 200)["url"]
+    end
+
+    @tag :requires_env
+    test "renders 404 when not authorized", %{conn: conn} do
+      project = insert(:project)
+
+      conn = get conn, project_path(conn, :stripe_button_url, project)
+      assert json_response(conn, 200)["url"]
```
begedin: Based on test title, 404. Not sure it should be 404, though. Probably 401

joshsmith: Just pushed a fix.

begedin:
@JoshSmith Had another look at this. Not really a big deal, but just something to consider...

Maybe this doesn't have to be a model at all? Why not maki it a simple ajax action client-side and then simply return a map -> to json from the API?

What are the advantages of it being a model other than a potential reduction in boilerplate? It feels like it's making things a bit less clear if we do it this way.

The way I see it, we don't really need to change anything except the view here (have it return a plain JSON payload instead of relying on `ja_serializer`) and then not try and squeeze it through ember-data client-side and instead just manage it as a simple ajax request-response.

joshsmith:
@begedin the reasoning is not just a small reduction in boilerplate but a huge one. This has to be an authorized request which means adding headers with the right JWT and so on. It makes sense to use our API as we normally would I think.

joshsmith:
@begedin if you could let me know if you agree with reasoning here then I can proceed.

joshsmith:
The reason I'm using `stripe_auth` instead of `show` is because I was using the project policy, likely because the organization membership helpers/finders were not expanded outside of that policy. This is probably superfluous with the introduction of new organization membership helpers.

joshsmith:
The StripeAuth view is missing a test.

joshsmith:
This is still missing the consumption of the tail end of the workflow, but given the size of this PR, I think it's reasonable to work on that after merging in this PR. That portion will involve using the returned token and JWT to create a `StripeAccount` for that project.

begedin:
> @begedin the reasoning is not just a small reduction in boilerplate but a huge one. This has to be an authorized request which means adding headers with the right JWT and so on. It makes sense to use our API as we normally would I think.

All of that makes sense. Really, the majority of what I think might be unnecessary would be ember-side. My question is, do we get an advantage by representing stripe auth as an ember-data model client-side. Would it not be simpler and clearer to have a service which makes an ajax request and stores the data as is, without passing it through the adapter, serializer, model layers of ember-data?

Outside of that question, the code here looks good to me. As I said, really, the only thing that would change in relation to my question is the `StripeAuthView`

joshsmith:
I tried the alternative and this is the best implementation IMO.

One thing that could make sense is using not a static ID for this but instead implementing either a time stamp as an ID or some sort of virtual UUID. But that data isn't going to hang around for long anyway, and any new request to the API will replace the data inside here.

We may at the very least want to add a `generated-at` attribute.

begedin (approved):
The changes themselves look good. We had the talk and I understand the arguments, so I'm ok with using this approach. Feel free to merge this part in, so we can continue with the rest.

PR #380, opened 2016-10-20
Add StripeAccount model

### LINKED ISSUE #362: Add StripeAccount model
For #258 

https://stripe.com/docs/api#account_object

The account is the Stripe account of an `Organization` on Code Corps. An account may have multiple `Plan`s that belong to a specific project.

ATTRIBUTES
- `id string` A unique identifier for the account
- `business_name string` The publicly visible name of the business
- `business_url string` The publicly visible website of the business
- `charges_enabled boolean` Whether or not the account can create live charges
- `country string` The country of the account
- `default_currency string` The currency this account has chosen to use as the default
- `details_submitted boolean` Whether or not account details have been submitted yet. Standalone accounts cannot receive transfers before this is true.
- `display_name string` The display name for this account. This is used on the Stripe dashboard to help you differentiate between accounts.
- `email string` The primary user’s email address
- `managed boolean` Whether or not the account is managed by your platform. Returns null if the account was not created by a platform.
- `support_email string` A publicly shareable email address that can be reached for support for this account
- `support_phone string` The publicly visible support phone number for the business
- `support_url string` A publicly shareable URL that can be reached for support for this account
- `transfers_enabled boolean` Whether or not Stripe will send automatic transfers for this account. This is only false when Stripe is waiting for additional information from the account holder.

begedin:
Closes #362 

I'm not really sure if anything else should be done here for now. The changeset is pretty much a `no_op`, as is the one test that tests it.

Really, it was just a matter of adding the migration and the model.

We have a `belongs_to :organization`, but I did not add a `has_one :stripe_account` to the other side.

### DIFF
--- a/lib/code_corps/stripe/adapters/stripe_account.ex
+++ b/lib/code_corps/stripe/adapters/stripe_account.ex
@@ -1,0 +1,7 @@
+defmodule CodeCorps.Stripe.Adapters.StripeAccount do
+  import CodeCorps.MapUtils, only: [rename: 3]
+
+  def params_from_stripe(%{} = stripe_map) do
+    stripe_map |> rename("id", "id_from_stripe")
+  end
+end
--- a/priv/repo/migrations/20161020144622_create_stripe_account.exs
+++ b/priv/repo/migrations/20161020144622_create_stripe_account.exs
@@ -1,0 +1,29 @@
+defmodule CodeCorps.Repo.Migrations.CreateStripeAccount do
+  use Ecto.Migration
+
+  def change do
+    create table(:stripe_accounts) do
+      add :business_name, :string
+      add :business_url, :string
+      add :charges_enabled, :boolean
+      add :country, :string
+      add :default_currency, :string
+      add :details_submitted, :boolean
+      add :display_name, :string
+      add :email, :string
+      add :id_from_stripe, :string, null: false
+      add :managed, :boolean
+      add :support_email, :string
+      add :support_phone, :string
+      add :support_url, :string
+      add :transfers_enabled, :boolean
+
+      add :organization_id, references(:organizations), null: false
+
+      timestamps()
+    end
+
+    create unique_index(:stripe_accounts, [:id_from_stripe])
+    create unique_index(:stripe_accounts, [:organization_id])
+  end
+end
--- a/web/models/stripe_account.ex
+++ b/web/models/stripe_account.ex
@@ -1,0 +1,35 @@
+defmodule CodeCorps.StripeAccount do
+  @moduledoc """
+  Represents a StripeAccount stored on Code Corps.
+  """
+
+  use CodeCorps.Web, :model
+
+  schema "stripe_accounts" do
+    field :business_name, :string
+    field :business_url, :string
+    field :charges_enabled, :boolean
+    field :country, :string
+    field :default_currency, :string
+    field :details_submitted, :boolean
+    field :display_name, :string
+    field :email, :string
+    field :id_from_stripe, :string, null: false
+    field :managed, :boolean
+    field :support_email, :string
+    field :support_phone, :string
+    field :support_url, :string
+    field :transfers_enabled, :boolean
+
+    belongs_to :organization, CodeCorps.Organization
+
+    timestamps()
+  end
+
+  def create_changeset(struct, params \\ %{}) do
+    struct
+    |> cast(params, [:id_from_stripe, :organization_id])
+    |> validate_required([:id_from_stripe, :organization_id])
+    |> assoc_constraint(:organization)
+  end
+end
--- a/lib/code_corps/map_utils.ex
+++ b/lib/code_corps/map_utils.ex
@@ -1,0 +1,7 @@
+defmodule CodeCorps.MapUtils do
+  def rename(map, old_key, new_key) do
+    map
+    |> Map.put(new_key, map |> Map.get(old_key))
+    |> Map.delete(old_key)
+  end
+end
--- a/test/lib/code_corps/stripe/adapters/stripe_account_test.exs
+++ b/test/lib/code_corps/stripe/adapters/stripe_account_test.exs
@@ -1,0 +1,14 @@
+defmodule CodeCorps.Stripe.Adapters.StripeAccountTest do
+  use ExUnit.Case, async: true
+
+  import CodeCorps.Stripe.Adapters.StripeAccount, only: [params_from_stripe: 1]
+
+  @stripe_map %{"id" => "str_123", "foo" => "bar"}
+  @local_map %{"id_from_stripe" => "str_123", "foo" => "bar"}
+
+  describe "params_from_stripe/1" do
+    test "converts from stripe map to local properly" do
+      assert @stripe_map |> params_from_stripe == @local_map
+    end
+  end
+end
--- a/test/models/stripe_account_test.exs
+++ b/test/models/stripe_account_test.exs
@@ -1,0 +1,41 @@
+defmodule CodeCorps.StripeAccountTest do
+  use CodeCorps.ModelCase
+
+  alias CodeCorps.StripeAccount
+
+  @valid_attrs %{
+    id_from_stripe: "abc123"
+  }
+
+  @invalid_attrs %{}
+
+  describe "create_changeset/2" do
+    test "reports as valid when attributes are valid" do
+      organization_id = insert(:organization).id
+
+      changes = Map.merge(@valid_attrs, %{organization_id: organization_id})
+      changeset = StripeAccount.create_changeset(%StripeAccount{}, changes)
+      assert changeset.valid?
+    end
+
+    test "reports as invalid when attributes are invalid" do
+      changeset = StripeAccount.create_changeset(%StripeAccount{}, @invalid_attrs)
+      refute changeset.valid?
+
+      assert changeset.errors[:id_from_stripe] == {"can't be blank", []}
+      assert changeset.errors[:organization_id] == {"can't be blank", []}
+    end
+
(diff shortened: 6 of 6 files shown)


joshsmith:
@begedin does Stripe have any `required` fields there or are they all optional?

begedin:
@JoshSmith Looked into it. `email` is required if `managed` is set to false. Otherwise, everything seems to be optional.

web/models/stripe_account.ex
```diff
@@ -0,0 +1,38 @@
...
+  @moduledoc """
+  Represents a StripeAccount stored on Code Corps.
+  """
+
+  use CodeCorps.Web, :model
+
+  @primary_key {:id, :string, []} # set primary key as string
+  @derive {Phoenix.Param, key: :id} # set method of determining primary key for route
+  schema "stripe_account" do
+    field :business_name, :string
```
SViccari, line 9: I'm curious if we want to store the `business_name` on the StripeAccount since each StripeAccount belongs to an organization, where I presume the business name and organization name are the same. But perhaps having the business name defined on the stripe account meets a feature requirement? :)

web/models/stripe_account.ex
```diff
@@ -0,0 +1,38 @@
...
+    field :business_url, :string
+    field :charges_enabled, :boolean
+    field :country, :string
```
SViccari, line 12: There's no need to change this but fwiw, the [field type defaults to string](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/schema.ex#L429), so you could leave out the specification of `:string` if you prefer.

web/models/stripe_account.ex
```diff
@@ -0,0 +1,38 @@
...
+    field :support_phone, :string
+    field :support_url, :string
+    field :transfers_enabled, :boolean
+
+    belongs_to :organization, CodeCorps.Organization
+
+    timestamps()
+  end
+
+  @doc """
```
SViccari: Is this comment generated when creating this file? Imho, I don't believe `changeset` is in need of documentation and this could be removed.

web/models/stripe_account.ex
```diff
@@ -0,0 +1,38 @@
...
+  Builds a changeset based on the `struct` and `params`.
+  """
+  def changeset(struct, params \\ %{}) do
+    struct
+    |> cast(params, [])
```
SViccari: Are all the attributes being casted as expected? I believe all parameters that are not explicitly allowed are ignored.

web/models/stripe_account.ex
```diff
@@ -0,0 +1,38 @@
...
+    |> validate_required([])
```
SViccari: Since this validation is not in use, what do you think about removing it? `validate_required` can always be added back once there is a required field to validate :)

web/models/stripe_account.ex
```diff
@@ -0,0 +1,38 @@
+defmodule CodeCorps.StripeAccount do
+  @moduledoc """
+  Represents a StripeAccount stored on Code Corps.
+  """
+
+  use CodeCorps.Web, :model
+
+  @primary_key {:id, :string, []} # set primary key as string
+  @derive {Phoenix.Param, key: :id} # set method of determining primary key for route
```
SViccari: Whoa, interesting, I haven't heard of `derive` until just now. From what I just briefly read about the use of `derive`, it seems `derive` is a great way to [set custom primary keys](http://www.phoenixframework.org/docs/ecto-custom-primary-keys). Since we're using `id` as the primary key, I'm not certain we want to use `derive` here. Same thing for `@primary_key`.

priv/repo/migrations/20161020144622_create_stripe_account.exs
```diff
@@ -0,0 +1,28 @@
+defmodule CodeCorps.Repo.Migrations.CreateStripeAccount do
+  use Ecto.Migration
+
+  def change do
+    create table(:stripe_accounts, primary_key: false) do
```
SViccari: It appears we're telling `create` to not create the traditional primary key `id` but then we add and set the `id` column below to `primary_key: true`. Perhaps we can simplify this to `create table(:stripe_accounts) do` and remove `add :id, :string, primary_key: :true` as the `id` column will be created for us.

SViccari (commented):
This is looking good! I left a few questions and comments.

SViccari:
Just curious, there has been some debate around committing changes to the `structure.sql` file when adding a migration. Has this team already had that discussion and decided to not commit the structure.sql file to source?

web/models/stripe_account.ex
begedin: You are right on the `@derive` part. We do not need it, so I will correct.

I looked into it, however, and we do need the `@primary_key` part, reason being that, by default, the `:id` column is of `:id` type. Since we use a string, we need to explicitly state that we use a string. Otherwise, `Repo.insert` will fail.

web/models/stripe_account.ex
begedin: We are attempting to simply clone the Stripe API object here for now, or, that is, clone the fields that might be relevant to us. The Stripe object's information needs to be valid for the Stripe API. Conceivably, the organization name/title used on the site, as part of the Organization record might end up being changeable in some way.

web/models/stripe_account.ex
begedin: I'm aware. However, we explicitly state the type in all of our other models, so I'm doing it here as well.

web/models/stripe_account.ex
begedin: I'm inclined to agree with you. The comment doesn't really do much for us. However, our other models have it, so would you mind creating an issue. We should either remove all of them, or rewrite them into something more useful.

web/models/stripe_account.ex
begedin: The changeset function was autogenerated but we don't really use the changeset for anything yet, so the function is basically a `no-op`. I'm actually considering removing it completely for now.

priv/repo/migrations/20161020144622_create_stripe_account.exs
begedin: We want the id to be a string, because this is a stripe object, which uses strings as ids. As far as I can tell, this is the only way to achieve that. Otherwise, by default, the `:id` will be of `:id` type, not `:string`

begedin:
@SViccari 

> Just curious, there has been some debate around committing changes to the structure.sql file when adding a migration. Has this team already had that discussion and decided to not commit the structure.sql file to source?

I'm not really sure. We were committing the equivalent file back in the rails API. @JoshSmith, any thoughts?

web/models/stripe_account.ex
joshsmith: Yes, the idea here is that the business name might be logically separate from whatever the project or organization name is. For example, Code Corps is really "Code Corps PBC", but that sounds like a peanut butter sandwich and isn't super pretty to display.

We really want the business name, as well.

web/models/stripe_account.ex
joshsmith: I don't think we want to be replacing our `id` with theirs, just have a `stripe_id` that's a `string` with an index, and our own auto-incrementing `id` that's unrelated.

web/models/stripe_account.ex
joshsmith: I think being explicit is probably a good choice.

joshsmith:
@SViccari @begedin we hadn't discussed it at all really. I'm not really up on the debate. We could create a separate discussion issue for that though?

web/models/stripe_account.ex
SViccari: I agree with @JoshSmith that adding a `stripe_id` column is the way to go. If we replace `id` with a value that's tied to a third party value, we're essentially making `id` a foreign key, which could be confusing in the future.

web/models/stripe_account.ex
SViccari: sounds good 🍹

web/models/stripe_account.ex
SViccari: I think removing them all in a separate PR sounds like a good idea 👍

web/models/stripe_account.ex
SViccari: Ah okay, since the changeset isn't in use yet, I'm in favor of removing it. I find unused code distracting as it leads me to ask "what uses this? If nothing, why is it here? Is it important?" etc.

joshsmith:
I've now removed the `id` and added a `stripe_id`.

begedin:
@JoshSmith Added a test for assoc constraint and did some work on test formatting. Your changes look good from my side, so this should be good to merge once you review mine from yours.

begedin:
Also, still not sure why coverage dropped. The files that claim to have reduced coverage weren't touched by this PR.

begedin:
@JoshSmith I just realized that adding the `id_from_stripe` field means we need an adapter fort `StripeAccount` to, so I added it.

lib/code_corps/stripe/adapters/stripe_account.ex
```diff
@@ -0,0 +1,11 @@
+defmodule CodeCorps.Stripe.Adapters.StripeAccount do
+  import CodeCorps.MapUtils, only: [rename: 3]
+
+  def params_from_stripe(%{} = stripe_map) do
+    stripe_map |> rename("id", "id_from_stripe")
+  end
+
+  def params_to_stripe(%{} = ecto_map) do
+    ecto_map |> rename("id_from_stripe", "id")
+  end
```
joshsmith: We'll remove this.

joshsmith (commented):
Quick need for removal here.

joshsmith (approved):

Issue #367, opened 2016-10-17
Create StripePlan model


joshsmith:
For #258

Stripe's docs: https://stripe.com/docs/api#plan_object

Belongs to a `Project` on Code Corps.

ATTRIBUTES
- `stripe_id`, corresponds to `id string`
- `amount` positive integer or zero The amount in cents to be charged on the interval specified
- `created timestamp`
- `name string` Display name of the plan

dmvt:
Will this be tracking subscriptions related to CodeCorps or donation subscriptions on the connected accounts?

joshsmith:
@dmvt this is for donation subscriptions to projects.

joshsmith:
Closed by #383.

PR #383, opened 2016-10-21
Add StripePlan and StripeSubscription models

### LINKED ISSUE #361: Create StripeSubscription model
For #258 

https://stripe.com/docs/api#subscription_object

ATTRIBUTES
- `stripe_id`, corresponds to `id` `string`
- `application_fee_percent decimal` A positive decimal that represents the fee percentage of the subscription invoice amount that will be transferred to the application owner’s Stripe account each billing period.
- `canceled_at timestamp` If the subscription has been canceled, the date of that cancellation. If the subscription was canceled with cancel_at_period_end, canceled_at will still reflect the date of the initial cancellation request, not the end of the subscription period when the subscription is automatically moved to a canceled state.
- `created timestamp`
- `current_period_end timestamp` End of the current period that the subscription has been invoiced for. At the end of this period, a new invoice will be created.
- `current_period_start timestamp` Start of the current period that the subscription has been invoiced for
- `customer string` ID of the customer who owns the subscription
- `ended_at timestamp` If the subscription has ended (either because it was canceled or because the customer was switched to a subscription to a new plan), the date the subscription ended
- `stripe_plan_stripe_id` customer's Stripe plan's stripe id
- `stripe_plan_id` the id in our database
- `quantity integer` The quantity of the plan to which the customer should be subscribed. For example, if your plan is $10/user/month, and your customer has 5 users, you could pass 5 as the quantity to have the customer charged $50 (5 x $10) monthly.
- `start timestamp` Date the most recent update to this subscription started.
- `status string` Possible values are trialing, active, past_due, canceled, or unpaid. A subscription still in its trial period is trialing and moves to active when the trial period is over. When payment to renew the subscription fails, the subscription becomes past_due. After Stripe has exhausted all payment retry attempts, the subscription ends up with a status of either canceled or unpaid depending on your retry settings. Note that when a subscription has a status of unpaid, no subsequent invoices will be attempted (invoices will be created, but then immediately automatically closed. Additionally, updating customer card details will not lead to Stripe retrying the latest invoice.). After receiving updated card details from a customer, you may choose to reopen and pay their closed invoices.

begedin:
Closes #361, #367 

Since a `StripeSubscription` is related to a `StripeCustomer`, we need to merge #377 before this can be merged. The lines of code that fail due to this have been commented out for now.

Instead of going with separate fields for stripe id and actual record id, I used the approach from #377 to make the id columns into strings.

Also similarly to #377, I created an adapter which does a minor conversion between a stripe payload for a Subscription into a locally compatible payload and back.

### DIFF
--- a/lib/code_corps/stripe/adapters/stripe_plan.ex
+++ b/lib/code_corps/stripe/adapters/stripe_plan.ex
@@ -1,0 +1,17 @@
+defmodule CodeCorps.Stripe.Adapters.StripePlan do
+  @moduledoc """
+  Used for conversion between stripe api payload maps and maps
+  usable for creation of `StripePlan` records locally
+  """
+
+  import CodeCorps.MapUtils, only: [rename: 3]
+
+  @doc """
+  Converts a map received from the Stripe API into a map that can be used
+  to create a `CodeCorps.StripePlan` record
+  """
+  def params_from_stripe(%{} = stripe_map) do
+    stripe_map
+    |> rename("id", "id_from_stripe")
+  end
+end
--- a/lib/code_corps/stripe/adapters/stripe_subscription.ex
+++ b/lib/code_corps/stripe/adapters/stripe_subscription.ex
@@ -1,0 +1,19 @@
+defmodule CodeCorps.Stripe.Adapters.StripeSubscription do
+  @moduledoc """
+  Used for conversion between stripe api payload maps and maps
+  usable for creation of StripeSubscription records locally
+  """
+
+  import CodeCorps.MapUtils, only: [rename: 3]
+
+  @doc """
+  Converts a map received from the Stripe API into a map that can be used
+  to create a `CodeCorps.StripeSubscription` record
+  """
+  def params_from_stripe(%{} = stripe_map) do
+    stripe_map
+    |> rename("id", "id_from_stripe")
+    |> rename("stripe_plan", "plan_id_from_stripe")
+    |> rename("customer", "customer_id_from_stripe")
+  end
+end
--- a/priv/repo/migrations/20161021131026_create_stripe_plans_stripe_subscriptions.exs
+++ b/priv/repo/migrations/20161021131026_create_stripe_plans_stripe_subscriptions.exs
@@ -1,0 +1,44 @@
+defmodule CodeCorps.Repo.Migrations.CreateStripePlansStripeSubscriptions do
+  use Ecto.Migration
+
+  def change do
+    create table(:stripe_plans) do
+      add :amount, :integer
+      add :created, :datetime
+      add :id_from_stripe, :string, null: false
+      add :name, :string
+
+      add :project_id, references(:projects), null: false
+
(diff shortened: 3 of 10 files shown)


lib/code_corps/stripe/adapters/stripe_subscription.ex
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  @doc """
+  Coverts a `CodeCorps.StripeSubscription` record map into a map that can be used
+  with the Stripe API
+  """
+  def params_to_stripe(%{} = ecto_map) do
+    ecto_map
+    |> rename("customer_id", "customer")
+    |> rename("plan_id", "stripe_plan")
```
joshsmith: I'm not sure I see the value in converting from Ecto to Stripe. I'm not sure when we'd be doing this.

test/lib/code_corps/stripe/adapters/stripe_plan_test.exs
```diff
@@ -0,0 +1,20 @@
...
+    test "converts from stripe map to local properly" do
+      assert @stripe_map |> params_from_stripe == @local_map
+    end
+  end
+
+  describe "params_to_stripe/1" do
+    test "converts from local to stripe map properly" do
+      assert @local_map |> params_to_stripe == @stripe_map
+    end
+  end
```
joshsmith: We'll remove.

lib/code_corps/stripe/adapters/stripe_plan.ex
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  @doc """
+  Coverts a `CodeCorps.StripePlan` record map into a map that can be used
+  with the Stripe API
+  """
+  def params_to_stripe(%{} = ecto_map) do
+    ecto_map
+    |> rename("id_from_stripe", "id")
+  end
```
joshsmith: We'll remove.

joshsmith (commented):
Only thing I think is left to change prior to unblocking when #377 is merged.

priv/repo/migrations/20161021131026_create_stripe_plans_stripe_subscriptions.exs
```diff
@@ -0,0 +1,44 @@
...
+      add :ended_at, :datetime
+      add :id_from_stripe, :string, null: false
+      add :plan_id_from_stripe, :string, null: false
+      add :quantity, :integer
+      add :start, :datetime
+      add :status, :string
+
+      add :stripe_plan_id, references(:stripe_plans), null: false
+      # TODO: Dependent on stripe cutomers being merged
+      # add :stripe_customer_id, references(:stripe_customers)
```
joshsmith: Don't forget this.

joshsmith:
@begedin this is no longer blocked by #377.

joshsmith (approved):

joshsmith:
@begedin squash when you can.

PR #384, opened 2016-10-21
Add authorization test for create

joshsmith:
A couple improvements here:
- Added a new test for authorization of create which wasn't there.
- Removed useless `current_user` information where it wasn't needed.

### DIFF
--- a/test/controllers/comment_controller_test.exs
+++ b/test/controllers/comment_controller_test.exs
@@ -44,7 +44,7 @@ defmodule CodeCorps.CommentControllerTest do
   describe "create" do
     @tag :authenticated
     test "creates and renders resource when data is valid", %{conn: conn, current_user: current_user} do
-      task = insert(:task, user: current_user)
+      task = insert(:task)
 
       payload =
         build_payload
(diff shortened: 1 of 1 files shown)


begedin (approved):
Seen and fixed one or two of these cases on other endpoints already. Good catch. Looks good 👍

test/controllers/comment_controller_test.exs
```diff
@@ -87,6 +87,22 @@ defmodule CodeCorps.CommentControllerTest do
 
       assert json_response(conn, 401)
     end
+
+    @tag :authenticated
+    test "does not create resource and renders 401 when not authorized", %{conn: conn} do
```
SViccari, line 92: Perhaps I'm missing something but it looks like this tests already exists up [above](https://github.com/code-corps/code-corps-api/blob/develop/test/controllers/comment_controller_test.exs#L76)

joshsmith: That's testing authentication but not authorization, I believe. The package we're using incorrectly returns 401 for unauthorized rather than 403.

Issue #245, opened 2016-09-20
Write the API documentation for donations


joshsmith:
We should start out #10 by writing out how we expect our API to work by writing the documentation first. This should drive the other decisions that we make.

Anything in italics requires authorization.
- [x] DonationGoal - index, show, _create, edit, delete_ #381 
- [x] StripeAccount - _index, show_ #390 
- [x] StripeCard - _index, show_ #385 
- [x] StripeCustomer - _index, show_ #389 
- [x] StripePlan - index, show, _create_ #387 
- [x] StripeSubscription - _index, show, create, edit_ #391

joshsmith:
We'll want to add an endpoint for `/projects/:id/stripe_button_url`, which is behind an authenticated route for the user who has permissions for it. This will return a `url` `string` for connecting to Stripe via OAuth. See #259.

joshsmith:
We'll want a `/donation-goals` endpoint with `create`, `retrieve`, `update` and `delete`. These should only be possible to create by a project's organization's owner. See #374.

begedin:
@JoshSmith 

> We'll want to add an endpoint for /projects/:id/stripe_button_url, which is behind an authenticated route for the user who has permissions for it. This will return a url string for connecting to Stripe via OAuth. See #259.

Are we sure we don't want to just add a `stripe_url` field to the project/organization instead?

joshsmith:
@begedin I'm just worried about how long it would take to generate that JWT on every request. I think that could be a real performance issue since it involves crypto, which is why I suggested this.

begedin:
Ah. Yeah, that makes sense. I assumed there's a reason.

PR #398, opened 2016-10-26
Added Stripe Plan to API Blueprint

### LINKED ISSUE #387: Add API Blueprint documentation for StripePlan
Parent issue #245 
- [ ] Add Endpoint for StripePlan with index(get) and show(get) and create(post)
- [ ] Add Data Structure for StripePlan and StripePlans.

mackenziehicks:
# What's in this PR?

Added Endpoint for StripePlan with index(get) and show(get) and create(post)

Added Data Structure for StripePlan and StripePlans.
## References

 Fixes #387 

 Progress on: #245

### DIFF
--- a/blueprint/api.apib
+++ b/blueprint/api.apib
@@ -1286,6 +1286,77 @@ This endpoint will return one of two possible responses, depending on if the slu
 
     + Attributes (Slugged Route Organization Response)
 
+# Group Stripe Plans
+
+This endpoint is for creating and returning Stripe plans available for projects.
+
+The Stripe plan resource allows users to select a monthly donation amount for a project. Stripe plans contain the pricing information for different projects and donation levels.
+
+## Stripe Plans [/stripe-plans]
+
+### Create a Stripe plan [POST]
+
++ Parameters
+
+    + id (number) - Id of a Stripe plan.
+
++ Request
+
+    + Headers
+
+            Accept: application/vnd.api+json
+            Authorization: Bearer {token}
+
++ Response 201 (application/vnd.api+json; charset=utf-8)
+
+    + Attributes (Stripe Plan Response)
+
++ Response 401 (application/vnd.api+json; charset=utf-8)
+
+    + Attributes (JSON Web Token Invalid Response)
+
++ Response 403 (application/vnd.api+json; charset=utf-8)
+
+    + Attributes (Forbidden Response)
+
++ Response 422 (application/vnd.api+json; charset=utf-8)
+
+    + Attributes (Unprocessable Entity Response)
+
+### List all Stripe plans [GET]
+
++ Request
+
+    + Headers
+
+            Accept: application/vnd.api+json
+
++ Response 200 (application/vnd.api+json; charset=utf-8)
+
+    + Attributes (Stripe Plans Response)
+
+## Stripe Plan [/stripe-plans/{id}]
+
+### Retrieve a Stripe plan [GET]
+
++ Parameters
+
+    + id (number) - Id of a Stripe plan.
+
++ Request
+
+    + Headers
+
+            Accept: application/vnd.api+json
+
++ Response 200 (application/vnd.api+json; charset=utf-8)
+
+    + Attributes (Stripe Plan Response)
+
++ Response 404 (application/vnd.api+json; charset=utf-8)
+
+    + Attributes (Record Not Found Response)
+
 # Group Tasks
 
 ## Project Tasks [/projects/{project_id}/tasks]
@@ -1954,7 +2025,7 @@ This endpoint allows you to check whether a username is valid (by running a vali
 
 ## Comments Response (object)
 + comments
-    + data(array[Comment Response], required)
+    + data(array[Comment Response])
 + include JSON API Version
 
 ## Comment Relationship (object)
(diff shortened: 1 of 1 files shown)


blueprint/api.apib
```diff
@@ -509,6 +509,76 @@ Until the Code Corps platform is open to new organizations, only admin users can
...
+    + Attributes (Category Response)
+
++ Response 403 (application/vnd.api+json; charset=utf-8)
+
+    + Attributes (Forbidden Response)
+
++ Response 404 (application/vnd.api+json; charset=utf-8)
+
+    + Attributes (Record Not Found Response)
+
```
joshsmith: It looks like the categories got duplicated in here someone. This probably all needs removed.

blueprint/api.apib
```diff
@@ -1103,6 +1173,77 @@ This endpoint will return one of two possible responses, depending on if the slu
 
     + Attributes (Slugged Route Organization Response)
 
+# Group Stripe Plans
+
+This endpoint is for creating and returning stripe plans available for projects.
```
joshsmith: Would capitalize "Stripe" everywhere you can.

blueprint/api.apib
```diff
@@ -2174,6 +2315,30 @@ This endpoint allows you to check whether a username is valid (by running a vali
...
++ relationships
+    + project (object)
+        + data(Project Resource Identifier)
+
+## Stripe Plan Response
++ data(Stripe Plan Resource)
++ include JSON API Version
+
+## Stripe Plans Response
++ data(array[Stripe Plan Resource], required)
```
joshsmith, line 2511: Curious why you added `required` here. We're not doing it elsewhere. I'd probably opt against it.

blueprint/api.apib
```diff
@@ -2174,6 +2315,30 @@ This endpoint allows you to check whether a username is valid (by running a vali
...
+
+##Stripe Plan Resource Identifier
++ id: `1` (string, required)
++ type: `stripe-plan` (string, required)
+
+##Stripe Plan Resource (object)
++ include Stripe Plan Resource Identifier
++ attributes(Stripe Plan Attributes)
++ relationships
+    + project (object)
```
joshsmith: I don't think you need the `(object)` here on the relationship key.

blueprint/api.apib
```diff
@@ -2174,6 +2315,30 @@ This endpoint allows you to check whether a username is valid (by running a vali
 + data(Slugged Route Organization Resource)
 + include JSON API Version
 
+## Stripe Plan Attributes (object)
++ amount: 999 (number)
++ created: `2016-07-08T03:03:51.967Z` (string)
++ name: `Emerald Personal` (string)
```
joshsmith: I think this is missing `inserted-at` and `updated-at` timestamps, which actually all objects in our database will have.

joshsmith: I also think you should be writing descriptions for all the attributes, like `amount: 999 (number) - Describes the amount of a plan in cents.`

Also, for the `name` here, the example doesn't really resonate with our application. We should come up with a naming convention for plans here. I think `Monthly Donation` is sufficient.

joshsmith (changes requested):
A few questions and suggested changes here.

Overall looks great, so consider this fairly minor feedback.

joshsmith:
I would also add a better description to the pull request. You should look at the pull request templates we're about to add in #395 for examples (although in the future, it will automatically populate with that so you can edit it).

blueprint/api.apib
mackenziehicks: I think I added it because I was looking at the Tasks Response below.  I noticed now that required is on the "Tasks Response" as well as the "Comments Response". Should those be removed as well?

joshsmith: I would probably remove them. It seems a little clunky to say it's required (and maybe not accurate?). The data array could be an empty `[]`.

blueprint/api.apib
```diff
@@ -2316,9 +2246,11 @@ This endpoint allows you to check whether a username is valid (by running a vali
 + include JSON API Version
 
 ## Stripe Plan Attributes (object)
-+ amount: 999 (number)
++ amount: 999 (number) - Describes the amount of a plan in cents.
 + created: `2016-07-08T03:03:51.967Z` (string)
-+ name: `Emerald Personal` (string)
++ name: `Monthly Donation` (string) - Name of the plan, to be displayed on invoices and in the web interface.
++ `inserted-at`: `2016-07-08T03:03:51.967Z` (string)
++ `updated-at`: `2014-07-07T18:01:26.000Z` (string)
```
joshsmith, line 2493: Alpha ordering here.

joshsmith (changes requested):
Just one more change we're good!

mackenziehicks:
Ok, I updated the description, does it look correct?

joshsmith (approved):
🎉 

LGTM!

joshsmith:
Now you just need to rebase and squash your commits.

mackenziehicks:
I just realized I forgot to add  id_from_stripe to the attributes. Should it be something like this?

id_from_stripe: `monthly-donation-353`

joshsmith:
@mackenziehicks I missed that, too. Look at Stripe's docs to see the example they use in their documentation. We should mirror their docs where possible.

mackenziehicks:
Right, the stripe documentation has the example for Stripe Plan as `"id": "emerald-personal-353"`.  I am assuming I need to mirror this? Or should I be looking somewhere else?

joshsmith:
@mackenziehicks ah, I didn't realize they had that! Let me check.

joshsmith:
Weird, so they're using slugs instead. I guess `monthly-donation-1` would be appropriate here, then.

blueprint/api.apib
```diff
@@ -2248,6 +2248,7 @@ This endpoint allows you to check whether a username is valid (by running a vali
 ## Stripe Plan Attributes (object)
 + amount: 999 (number) - Describes the amount of a plan in cents.
 + created: `2016-07-08T03:03:51.967Z` (string)
++ id_from_stripe: `monthly-donation-1`
```
joshsmith: This should be dasherized, too. Like `id-from-stripe`.

blueprint/api.apib
```diff
@@ -2248,7 +2248,7 @@ This endpoint allows you to check whether a username is valid (by running a vali
 ## Stripe Plan Attributes (object)
 + amount: 999 (number) - Describes the amount of a plan in cents.
 + created: `2016-07-08T03:03:51.967Z` (string)
-+ id_from_stripe: `monthly-donation-1`
++ id-from-stripe: `monthly-donation-1`
```
joshsmith: I think this needs to be surrounded with ticks. Is this rendering for you without doing that?

blueprint/api.apib
```diff
@@ -2174,6 +2245,33 @@ This endpoint allows you to check whether a username is valid (by running a vali
 + data(Slugged Route Organization Resource)
 + include JSON API Version
 
+## Stripe Plan Attributes (object)
++ amount: 999 (number) - Describes the amount of a plan in cents.
++ created: `2016-07-08T03:03:51.967Z` (string)
++ id-from-stripe: `monthly-donation-1`
```
joshsmith: In addition to ticks surrounding this, also think we need to have this be `(string, required)`

joshsmith:
👍  okay, squash and rebase and we should be good to go.

joshsmith:
🚢 🇮🇹

joshsmith:
🍦 🐤 🌸 🍾

Issue #393, opened 2016-10-25
Refactor API Blueprint documentation


amyschools:
Referenced in comment on pull request #386 by @JoshSmith that the documentation needs to be refactored after changeover to Phoenix.

Suggestions so far:
- [ ] Check consistency of requiring both 'id' and 'type' for Resource Identifier

joshsmith:
There are absolutely a number of areas here that need cleaned up. There's some inconsistency in our object usage, places where `id` parameters are used vs not, missing `required`, etc. A solid read-through really needs done.

joshsmith:
Closed by #400.

PR #400, opened 2016-10-26
Rework API Blueprint to fix some errors, warnings, inconistencies

joshsmith:
# What's in this PR?

Fixed a bunch of tiny errors, warnings, and little inconsistencies here and there.
## References

Progress on: #393

### DIFF
--- a/blueprint/api.apib
+++ b/blueprint/api.apib
@@ -25,12 +25,12 @@ Exchange email and password for `token`.
 
 ### Exchange email and password for token [POST]
 
-+ Parameters
-
-    + username: `me@testuser.org` (string, required) - The user's email address.
-    + password: `password` (string, required)
-
 + Request
+
+    + Attributes
+
+        + username: `me@testuser.org` (string, required) - The user's email address.
+        + password: `password` (string, required)
 
     + Headers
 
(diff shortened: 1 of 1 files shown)


blueprint/api.apib
```diff
@@ -336,13 +332,13 @@ To create a donation goal the user must be a site admin or the owner of the refe
...
 ### Filter donation goals by id [GET /donation-goals{?filter[id]}]
 
-+ Parameters
++ Request
 
-    + `filter[id]`: `1,2,3` (string, required) - Comma separated string of `ids` to filter by.
+    + Attributes
 
-        Technically, the parameter is `filter: { id: '1,2,3' }`, which ends up being serialized into `filter[id]`. However, syntax constraints do not allow us to define it in the documentation this way.
+        + `filter[id]`: `1,2,3` (string, required) - Comma separated string of `ids` to filter by.
```
joshsmith, line 278: Actually I don't think this is how this should be written. I think it should still be a parameter.

begedin (approved):
Looks good. I still wish we could write the filter parameter in a better way, but we're consistent and it works. If there ever is a better way, we can change it more easily.

joshsmith:
@begedin I believe it's how it's documented in their docs.

PR #404, opened 2016-10-28
Refactor comments controller to use new helpers

joshsmith:
# What's in this PR?

Refactoring to use `CodeCorps.ApiCase` request helper functions.

Also _finally_ fixes 401 "unauthorized" to really render as 403 forbidden.

### DIFF
--- a/test/controllers/comment_controller_test.exs
+++ b/test/controllers/comment_controller_test.exs
@@ -1,13 +1,11 @@
 defmodule CodeCorps.CommentControllerTest do
-  use CodeCorps.ApiCase
+  use CodeCorps.ApiCase, resource_name: :comment
 
   alias CodeCorps.Comment
   alias CodeCorps.Repo
 
   @valid_attrs %{markdown: "I love elixir!"}
   @invalid_attrs %{markdown: ""}
-
-  defp build_payload, do: %{ "data" => %{"type" => "comment"}}
 
   describe "index" do
     test "lists all entries on index", %{conn: conn} do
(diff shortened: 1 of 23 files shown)


test/controllers/comment_controller_test.exs
```diff
@@ -1,14 +1,12 @@
 defmodule CodeCorps.CommentControllerTest do
-  use CodeCorps.ApiCase
+  use CodeCorps.ApiCase, resource_name: :comment
```
joshsmith, line 2: This is required for using the `Conn.ApiCase` request helpers.

test/controllers/comment_controller_test.exs
```diff
@@ -34,149 +32,70 @@ defmodule CodeCorps.CommentControllerTest do
...
-      payload =
-        build_payload
-        |> put_attributes(@valid_attrs)
-        |> put_relationships(current_user, task)
-
-      path = conn |> comment_path(:create)
-      conn = conn |> post(path, payload)
-
-      assert json_response(conn, 201)["data"]["id"]
+      attrs = @valid_attrs |> Map.merge(%{task: task, user: current_user})
```
joshsmith, line 51: We use `Map.merge` instead of `put_relationships`, but I would prefer to refactor `put_relationships` instead.

joshsmith (commented):
Some comments about the refactoring that happened here.

test/controllers/comment_controller_test.exs
```diff
@@ -34,149 +32,70 @@ defmodule CodeCorps.CommentControllerTest do
...
-      assert data["attributes"]["body"] == comment.body
-      assert data["attributes"]["markdown"] == comment.markdown
-      assert data["relationships"]["user"]["data"]["id"] == "#{comment.user_id}"
-      assert data["relationships"]["task"]["data"]["id"] == "#{comment.task_id}"
+      conn
+      |> request_show(comment)
+      |> json_response(200)
+      |> Map.get("data")
+      |> assert_result_id(comment.id)
     end
```
amyschools, line 40: Why you didn't check the comment.user_id and comment.task_id in the refactored test? Or is it implied in the `Map.get("data")`?

joshsmith: Because we're testing it in the view test. It doesn't really need to be tested at this level, I think.

test/controllers/comment_controller_test.exs
```diff
@@ -34,149 +32,70 @@ defmodule CodeCorps.CommentControllerTest do
...
 
     test "renders 404 when id is nonexistent", %{conn: conn} do
-      path = conn |> comment_path(:show, -1)
```
mackenziehicks, line 52: what does the '-1' mean?

joshsmith: @mackenziehicks it's a good way of saying "This `id` absolutely does not exist in the database".

test/controllers/comment_controller_test.exs
```diff
@@ -34,149 +32,70 @@ defmodule CodeCorps.CommentControllerTest do
...
-        build_payload
-        |> put_id(comment.id)
-        |> put_attributes(@valid_attrs)
-
-      path = conn |> comment_path(:update, comment)
-      conn = conn |> put(path, payload)
-
-      assert json_response(conn, 401)
+      assert conn |> request_update(@valid_attrs) |> json_response(401)
     end
```
amyschools, line 90: Do you just need to test for 401 and not 403 because they only work together?

joshsmith: `403` renders inaccurately as a `401` as unauthorized.

amyschools (approved):

PR #405, opened 2016-10-31
Refactor AnalyticsIdentify plug to use pattern matching

joshsmith:
# What's in this PR?

Changes `if/else` style to pattern matching.

### DIFF
--- a/web/plugs/analytics_identify.ex
+++ b/web/plugs/analytics_identify.ex
@@ -5,12 +5,11 @@ defmodule CodeCorps.Plug.AnalyticsIdentify do
 
   def init(opts), do: opts
 
-  def call(conn, _opts) do
-    if current_user = conn.assigns[:current_user] do
-      CodeCorps.Analytics.Segment.identify(current_user)
-      conn
-    else
-      conn
-    end
+  def call(conn, _opts), do: conn |> identify
+
+  defp identify(%{assigns: %{current_user: user}} = conn) do
+    CodeCorps.Analytics.Segment.identify(user)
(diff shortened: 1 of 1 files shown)


web/plugs/analytics_identify.ex
```diff
@@ -5,12 +5,11 @@ defmodule CodeCorps.Plug.AnalyticsIdentify do
...
-  def call(conn, _opts) do
-    if current_user = conn.assigns[:current_user] do
-      CodeCorps.Analytics.Segment.identify(current_user)
-      conn
-    else
-      conn
-    end
+  def call(conn, _opts), do: conn |> identify
+
+  def identify(%{assigns: %{current_user: user}} = conn) do
```
sbatson5: Can `identify` be private?

sbatson5 (approved):
Just 1 question. Otherwise, LGTM 👍

web/plugs/analytics_identify.ex
joshsmith: @sbatson5 I guess there's no reason why it can't be. Can you adjust your review to needs changes then so I know to change it?

sbatson5: Lol. I don't know how to do that, haha

joshsmith: Add another review that's a needs changes.

sbatson5 (dismissed):
Can we make the `identify` method private for both cases?

joshsmith:
@sbatson5 ready for another review.

Issue #406, opened 2016-10-31
Refactor Preview Controller Test to use new helpers 


mackenziehicks:
# Problem

Refactoring to use CodeCorps.ApiCase request helper functions.

PR #407, opened 2016-10-31
Refactor Preview Controller Test with Helpers

mackenziehicks:
# What's in this PR?

Refactored the preview controller test to use CodeCorps.ApiCase request helper functions.

Added 401/403 Response to comment endpoint for create. 

## References
Fixes: #406

### DIFF
--- a/blueprint/api.apib
+++ b/blueprint/api.apib
@@ -169,6 +169,14 @@ You can list all comments or retrieve individual comments, as well as create and
 
     + Attributes (Comment Response)
 
++ Response 401 (application/vnd.api+json; charset=utf-8)
+
+    + Attributes (JSON Web Token Invalid Response)
+
++ Response 403 (application/vnd.api+json; charset=utf-8)
+
+    + Attributes (Forbidden Response)
+
 + Response 422 (application/vnd.api+json; charset=utf-8)
 
     + Attributes (Unprocessable Entity Response)
@@ -622,6 +630,10 @@ Preview resources allow a user to create a temporary HTML preview for any kind o
 + Response 401 (application/vnd.api+json; charset=utf-8)
 
     + Attributes (JSON Web Token Invalid Response)
+
++ Response 403 (application/vnd.api+json; charset=utf-8)
+
+    + Attributes (Forbidden Response)
 
 + Response 422 (application/vnd.api+json; charset=utf-8)
 
--- a/test/controllers/preview_controller_test.exs
+++ b/test/controllers/preview_controller_test.exs
@@ -1,53 +1,24 @@
 defmodule CodeCorps.PreviewControllerTest do
-  use CodeCorps.ApiCase
+  use CodeCorps.ApiCase, resource_name: :preview
 
   alias CodeCorps.Preview
 
-  defp build_payload do
-    %{"data" => %{"type" => "preview", "attributes" => %{markdown: "A **strong** element"}}}
-  end
+  @valid_attrs %{markdown: "A **strong** element"}
 
   describe "create" do
     @tag :authenticated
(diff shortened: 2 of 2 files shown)


test/controllers/preview_controller_test.exs
```diff
@@ -44,9 +25,7 @@ defmodule CodeCorps.PreviewControllerTest do
     end
 
     test "does not create resource, and responds with 401 when unauthenticated", %{conn: conn} do
-      payload = build_payload
-      path = conn |> preview_path(:create)
-      assert conn |> post(path, payload) |> json_response(401)
+        assert conn |> request_create(@valid_attrs) |> json_response(401)
```
joshsmith: Watch indentation here.

test/controllers/preview_controller_test.exs
```diff
@@ -44,9 +25,7 @@ defmodule CodeCorps.PreviewControllerTest do
     end
```
joshsmith, line 17: Is there a better way to refactor these lines, or should they be removed? What are we testing here?

blueprint/api.apib
```diff
@@ -169,6 +169,14 @@ You can list all comments or retrieve individual comments, as well as create and
 
     + Attributes (Comment Response)
 
++ Response 401 (application/vnd.api+json; charset=utf-8)
+
+    + Attributes (JSON Web Token Invalid Response)
+
++ Response 403 (application/vnd.api+json; charset=utf-8)
+
+    + Attributes (Forbidden Response)
```
joshsmith, line 178: Good call on adding these; are they both tested in our controller test, though?

mackenziehicks: Yeah they are. Should we also add 403 to the Preview create responses? Or is it unnecessary in this case?

test/controllers/preview_controller_test.exs
mackenziehicks: Isn't it testing that we get a 201 response when creating a preview?  This test was in the comments controller as well.

blueprint/api.apib
joshsmith: You're sure `403` is tested?

test/controllers/preview_controller_test.exs
joshsmith: Yes, but there are several other lines here in your test, too.

blueprint/api.apib
mackenziehicks: It is tested in the Comment Controller test, but it's not in the Preview Controller Test. We weren't sure if it should be added to both the API blueprint and the test for the Preview Controller.

joshsmith: If it's in the API Blueprint it should probably be tested.

Both tests and documentation are your contracts that verify you're doing what you said you were going to do.

mackenziehicks: It's not in the API Blueprint.  Should it be?

joshsmith: Ah, sorry. I missed that you were updating _comments_ here, not previews. One moment.

joshsmith: So `create` does a `load_and_authorize_changeset`, right? What does that tell you about the responses you need to document and test?

amyschools: Since it's load_and_authorize it seems like it would need 403 as well as 401 since you need to be authorized as well as authenticated...?

joshsmith: That sounds right to me.

What would you change then?

joshsmith (approved):
Looks great! Rebase and squash.

joshsmith:
🙌 @amyschools @mackenziehicks

PR #410, opened 2016-11-01
Refactor Project Skills Controller Tests with helpers

### LINKED ISSUE #408: Refactor Project Skill Controller Tests with helpers
# Problem

Adding helpers from api_case.ex to the tests for Project Skill Controller.

amyschools:
# What's in this PR?

Added helpers from api_case.ex to refactor controller tests. 

## References
Fixes #408 

Progress on #413

### DIFF
--- a/test/controllers/project_skill_controller_test.exs
+++ b/test/controllers/project_skill_controller_test.exs
@@ -1,42 +1,25 @@
 defmodule CodeCorps.ProjectSkillControllerTest do
-  use CodeCorps.ApiCase
-
-  alias CodeCorps.ProjectSkill
-  alias CodeCorps.Repo
-
-  defp build_payload, do: %{ "data" => %{"type" => "project-skill", "attributes" => %{}}}
+  use CodeCorps.ApiCase, resource_name: :project_skill
 
   describe "index" do
     test "lists all entries on index", %{conn: conn} do
-      conn = get conn, project_skill_path(conn, :index)
-      assert json_response(conn, 200)["data"] == []
+      [project_skill_1, project_skill_2] = insert_pair(:project_skill)
+
+      conn
+      |> request_index
+      |> json_response(200)
+      |> assert_ids_from_response([project_skill_1.id, project_skill_2.id])
     end
 
     test "filters resources on index", %{conn: conn} do
-      elixir = insert(:skill, title: "Elixir")
-      phoenix = insert(:skill, title: "Phoenix")
-      rails = insert(:skill, title: "Rails")
+      [project_skill_1, project_skill_2 | _] = insert_list(3, :project_skill)
 
-      project = insert(:project)
-      project_skill_1 = insert(:project_skill, project: project, skill: elixir)
-      project_skill_2 = insert(:project_skill, project: project, skill: phoenix)
-      insert(:project_skill, project: project, skill: rails)
-      json =
-        conn
-        |> get("project-skills/?filter[id]=#{project_skill_1.id},#{project_skill_2.id}")
-        |> json_response(200)
+      path = "project-skills/?filter[id]=#{project_skill_1.id},#{project_skill_2.id}"
 
-      [first_result, second_result | _rest] = json |> Map.get("data")
-
-      first_result
-      |> assert_result_id(project_skill_1.id)
-      |> assert_jsonapi_relationship("project", project.id)
-      |> assert_jsonapi_relationship("skill", elixir.id)
-
-      second_result
-      |> assert_result_id(project_skill_2.id)
-      |> assert_jsonapi_relationship("project", project.id)
-      |> assert_jsonapi_relationship("skill", phoenix.id)
+      conn
+      |> get(path)
+      |> json_response(200)
+      |> assert_ids_from_response([project_skill_1.id, project_skill_2.id])
     end
   end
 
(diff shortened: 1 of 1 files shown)


test/controllers/project_skill_controller_test.exs
```diff
@@ -45,20 +38,16 @@ defmodule CodeCorps.ProjectSkillControllerTest do
...
       |> assert_result_id(project_skill.id)
-      |> assert_jsonapi_relationship("project", project.id)
-      |> assert_jsonapi_relationship("skill", skill.id)
-      |> assert_attributes(%{})
     end
 
-    test "does not show resource and instead throw error when id is nonexistent", %{conn: conn} do
-      path = conn |> project_skill_path(:show, -1)
-      assert conn |> get(path) |> json_response(:not_found)
+    test "does not show resource and instead renders 404 when id is nonexistent", %{conn: conn} do
```
mackenziehicks: I think on some of the others this is just "renders 404 when id is nonexistent"

test/controllers/project_skill_controller_test.exs
```diff
@@ -71,75 +60,49 @@ defmodule CodeCorps.ProjectSkillControllerTest do
 
       insert(:organization_membership, role: "admin", member: current_user, organization: organization)
```
mackenziehicks: could the `role: "admin"` be eliminated by changing the `@tag :authenticated` to  `@tag authenticated: :admin` ? @JoshSmith, I'm not clear about this.

test/controllers/project_skill_controller_test.exs
```diff
@@ -1,14 +1,16 @@
 defmodule CodeCorps.ProjectSkillControllerTest do
-  use CodeCorps.ApiCase
+  use CodeCorps.ApiCase, resource_name: :project_skill
 
   alias CodeCorps.ProjectSkill
   alias CodeCorps.Repo
 
-  defp build_payload, do: %{ "data" => %{"type" => "project-skill", "attributes" => %{}}}
+  @valid_attrs %{}
```
mackenziehicks: maybe this should be called `@attrs` since there is not an `@invalid_attrs` ?

test/controllers/project_skill_controller_test.exs
```diff
@@ -71,75 +60,49 @@ defmodule CodeCorps.ProjectSkillControllerTest do
...
 
     test "does not delete resource and renders 401 when unauthenticated", %{conn: conn} do
-      path = conn |> project_skill_path(:delete, "id not important")
-      assert conn |> delete(path) |> json_response(401)
+      assert conn |> request_delete |> json_response(401)
     end
 
     @tag :authenticated
-    test "does not create resource and renders 403 when not authorized", %{conn: conn} do
-      project_skill = insert(:project_skill)
```
mackenziehicks, line 134: does this need to stay? or is it unnecessary? Was wondering if it was needed for test above too? @JoshSmith

joshsmith: It's unnecessary in this one (since `request_delete` now handles the `insert` on its own).

test/controllers/project_skill_controller_test.exs
mackenziehicks: actually kind of confused on what this actually does with the Map.merge since its empty?

test/controllers/project_skill_controller_test.exs
joshsmith: I do worry about the `@tag authenticated: :admin`, though. I think this is too magical and needs some explanation. Been worried about this previously but hadn't mentioned to @begedin.

What we're testing is that it at least hits the policy; the policy test tests the variations that can happen here (site admin, project owner, project admin, etc). So just having a single test here is fine.

joshsmith: I don't think the `insert` is even required on line 89 in that case. The site admin can do whatever they want.

test/controllers/project_skill_controller_test.exs
joshsmith: I would even shorten universally to just `renders 404`.

test/controllers/project_skill_controller_test.exs
joshsmith: No, in the latter the `admin` is the "site admin".  In the former the admin is the admin role on the organization.

I think we need to open an issue to change `@tag authenticated: :admin` to `@tag authenticated: :site_admin` for more explicitness.

amyschools: That's what I thought, I think more explicitness would be helpful though.

test/controllers/project_skill_controller_test.exs
joshsmith: I think the contrapositive here is again instructive: is `@valid_attrs` useful _at all_?

Seems like you'd be better off assembling the `attrs` as a map, without a `Map.merge` operation, inside the tests where you need those relationships.

mackenziehicks: Ok, we actually tried that and noticed the test would pass.  But also noticed in several other tests that `@valid_attrs` and `@invalid_attrs` were set equal to `#{}` so wanted to check before changing it. In some cases, these weren't referenced anywhere else in the tests besides where they were "assigned".  Are these tests that haven't been refactored? Or partially refactored?

joshsmith: It's possible. The churn is so high here that if something has a code smell, it's probably a valid smell.

test/controllers/project_skill_controller_test.exs
```diff
@@ -71,75 +60,49 @@ defmodule CodeCorps.ProjectSkillControllerTest do
...
 
     @tag authenticated: :admin
-    test "does not create resource and renders errors when data is invalid", %{conn: conn} do
-      payload = build_payload
-      path = conn |> project_skill_path(:create)
-      json = conn |> post(path, payload) |> json_response(422)
-
-      assert json["errors"] != %{}
+    test "renders 422 when data is invalid", %{conn: conn} do
+      assert conn |> request_create(@attrs) |> json_response(422)
```
joshsmith: This looks like a case where we want `@invalid_attrs`, which is empty (no relationships).

test/controllers/project_skill_controller_test.exs
```diff
@@ -71,75 +60,49 @@ defmodule CodeCorps.ProjectSkillControllerTest do
...
     end
 
-    test "does not create resource and renders 401 when unauthenticated", %{conn: conn} do
-      path = conn |> project_skill_path(:create)
-      assert conn |> post(path) |> json_response(401)
+    test "renders 401 when unauthenticated", %{conn: conn} do
+      assert conn |> request_create(@attrs) |> json_response(401)
```
joshsmith: Do you need to pass in empty attrs here?

test/controllers/project_skill_controller_test.exs
```diff
@@ -71,75 +60,49 @@ defmodule CodeCorps.ProjectSkillControllerTest do
...
-      project = insert(:project, organization: organization)
-      skill = insert(:skill)
 
       insert(:organization_membership, role: "contributor", member: current_user, organization: organization)
 
-      payload = build_payload |> put_relationships(project, skill)
-
-      path = conn |> project_skill_path(:create)
-      assert conn |> post(path, payload) |> json_response(403)
+      assert conn |> request_create(@attrs) |> json_response(403)
```
joshsmith: Do you need to pass in empty attrs here?

Is the organization/organization membership above doing anything?

test/controllers/project_skill_controller_test.exs
```diff
@@ -71,75 +60,49 @@ defmodule CodeCorps.ProjectSkillControllerTest do
...
     end
   end
 
   describe "delete" do
     @tag authenticated: :admin
     test "deletes chosen resource", %{conn: conn} do
       project_skill = insert(:project_skill)
-      path = conn |> project_skill_path(:delete, project_skill)
-      assert conn |> delete(path) |> response(204)
+      assert conn |> request_delete(project_skill) |> response(204)
```
joshsmith: Do we need to specify the `project_skill` here?

amyschools:
@JoshSmith I can't get this test to pass that you changed during our screencast: 

```
  describe "create" do
    @tag authenticated: :admin
    test "creates and renders resource when data is valid", %{conn: conn} do
      project_skill = build(:project_skill)
      assert conn |> request_create(project_skill) |> json_response(201)
    end
```

the error says "Protocol Enumerable not implemented for %CodeCorps.ProjectSkill..."
Am I missing something?

joshsmith:
Do it as an `insert` instead and see if that works.

amyschools:
@JoshSmith this works, I think it's basically what I had before

```
  describe "create" do
    @tag authenticated: :admin
    test "creates and renders resource when data is valid", %{conn: conn} do
      project = insert(:project)
      skill = insert(:skill)
      attrs = %{project: project, skill: skill}
      assert conn |> request_create(attrs) |> json_response(201)
    end
```

test/controllers/project_skill_controller_test.exs
```diff
@@ -1,42 +1,25 @@
...
-      second_result
-      |> assert_result_id(project_skill_2.id)
-      |> assert_jsonapi_relationship("project", project.id)
-      |> assert_jsonapi_relationship("skill", phoenix.id)
+      path = "project-skills/?filter[id]=#{project_skill_1.id},#{project_skill_2.id}"
+
+      conn
+      |> get(path)
+      |> json_response(200)
+      |> assert_ids_from_response([project_skill_1.id, project_skill_2.id])
```
joshsmith, line 22: Is there no `request_index` variant for filters?

test/controllers/role_skill_controller_test.exs
```diff
@@ -28,22 +29,21 @@ defmodule CodeCorps.RoleSkillControllerTest do
...
       second_result
-      |> assert_result_id(role_skill_2.id)
-      |> assert_jsonapi_relationship("role", role.id)
-      |> assert_jsonapi_relationship("skill", phoenix.id)
+       |> assert_result_id(role_skill_2.id)
+       |> assert_jsonapi_relationship("role", role.id)
+       |> assert_jsonapi_relationship("skill", phoenix.id)
 
-      assert second_result["attributes"] == %{}
+       assert second_result["attributes"] == %{}
```
joshsmith: It looks like spacing in this whole file was affected somehow.

joshsmith (changes requested):

test/controllers/project_skill_controller_test.exs
begedin: I don't think there is at the moment, but it would not be a bad idea to write one.

test/controllers/project_skill_controller_test.exs
```diff
@@ -1,42 +1,25 @@
...
+     |> request_index
+     |> json_response(200)
+     |> assert_ids_from_response([project_skill_1.id, project_skill_2.id])
     end
 
     test "filters resources on index", %{conn: conn} do
-      elixir = insert(:skill, title: "Elixir")
-      phoenix = insert(:skill, title: "Phoenix")
-      rails = insert(:skill, title: "Rails")
+      [project_skill_1, project_skill_2] = insert_pair(:project_skill)
```
begedin: You figured out the missing third record, so I'll just link https://github.com/code-corps/code-corps-api/pull/425/files#r86388037 as a pointer on how to write this in one line.

begedin (dismissed):
Good for me, after a minor change.

test/controllers/role_skill_controller_test.exs
joshsmith: This still needs to be fixed to be mergeable.

joshsmith (changes requested):

test/controllers/role_skill_controller_test.exs
amyschools: Was looking at the wrong file. I think I got it now

joshsmith (approved):

joshsmith:
LGTM.

Issue #415, opened 2016-11-01
Refactor User Category Controller Test 


mackenziehicks:
# Problem

Refactor by adding helpers from api_case.ex to the User Category Controller Test 

## References

Progress on: # 413

Issue #413, opened 2016-11-01
Refactor Controller Tests with helpers


mackenziehicks:
# Problem

Refactor by adding helpers from api_case.ex to the tests for controllers.

See #404 as an example. 

## Subtasks
- [x]  User Role Controller Test #409 
- [x]  Project Skill Controller Test #408 
- [x]  User Category Controller Test #415 
- [x]  Role Skill Controller Test #414 
- [x]  Project Category Controller Test #418 
- [x] Organization Membership Controller Test #419 
- [x] User Skill Controller Test #421 
- [x]  Category #427 
- [x] Donation Goal #428 
- [x] Organization #430 
- [x] Preview #406  
- [x] Role #433
- [x] Skill #436 
- [x] Slugged Route #439 
- [x] Project #437 
- [x] Task #442 
- [x] User #443 
- [ ] Token
- [ ] Stripe Auth

PR #411, opened 2016-11-01
Refactor User Role Controller Test with Helpers

mackenziehicks:
# What's in this PR?

Added helpers from api_case.ex to the tests for User Role Controller.

## References
Fixes: #409

PR #417, opened 2016-11-01
Refactor User Category Controller Test

mackenziehicks:
# What's in this PR?

Refactored by adding helpers from api_case.ex to the User Category Controller Test

## References
Fixes: #415

Progress on: #413

### DIFF
--- a/test/controllers/user_category_controller_test.exs
+++ b/test/controllers/user_category_controller_test.exs
@@ -1,60 +1,40 @@
 defmodule CodeCorps.UserCategoryControllerTest do
-  use CodeCorps.ApiCase
-
-  alias CodeCorps.Repo
-  alias CodeCorps.UserCategory
-
-  @attrs %{}
-
-  defp build_payload, do: %{ "data" => %{"type" => "user-category", "attributes" => %{}}}
+  use CodeCorps.ApiCase, resource_name: :user_category
 
   describe "index" do
     test "lists all entries on index", %{conn: conn} do
-      path = conn |> user_category_path(:index)
-      json = conn |> get(path) |> json_response(200)
+      [user_category_1, user_category_2] = insert_pair(:user_category)
 
-      assert json["data"] == []
+      conn
+       |> request_index
+       |> json_response(200)
+       |> assert_ids_from_response([user_category_1.id, user_category_2.id])
     end
 
     test "filters resources on index", %{conn: conn} do
-      society = insert(:category, name: "Society")
-      technology = insert(:category, name: "Technology")
-      government = insert(:category, name: "Government")
-
-      user = insert(:user)
-      user_category_1 = insert(:user_category, user: user, category: society)
-      user_category_2 = insert(:user_category, user: user, category: technology)
-      insert(:user_category, user: user, category: government)
+      [user_category_1, user_category_2 | _] = insert_list(3, :user_category)
 
       path = "user-categories/?filter[id]=#{user_category_1.id},#{user_category_2.id}"
-      json = conn |> get(path) |> json_response(200)
 
-      data = json["data"]
-      assert length(data) == 2
-
-      [first_result, second_result | _] = data
-      assert first_result["id"] == "#{user_category_1.id}"
-      assert second_result["id"] == "#{user_category_2.id}"
+      conn
+      |> get(path)
+      |> json_response(200)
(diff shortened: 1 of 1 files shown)


test/controllers/user_category_controller_test.exs
```diff
@@ -63,65 +50,44 @@ defmodule CodeCorps.UserCategoryControllerTest do
...
+    test "renders 401 when unauthenticated", %{conn: conn} do
+      assert conn |> request_create(%{}) |> json_response(401)
     end
 
     @tag :authenticated
-    test "does not create resource and renders 403 when not authorized", %{conn: conn} do
-      path = conn |> user_category_path(:create)
-      assert conn |> post(path, build_payload) |> json_response(403)
+    test "renders 403 when not authorized", %{conn: conn} do
+      assert conn |> request_create(@invalid_attrs) |> json_response(403)
```
amyschools: Do you need invalid_attrs here or just an empty map?

test/controllers/user_category_controller_test.exs
```diff
@@ -63,65 +50,44 @@ defmodule CodeCorps.UserCategoryControllerTest do
...
   end
 
   describe "delete" do
     @tag authenticated: :admin
     test "deletes resource", %{conn: conn} do
       user_category = insert(:user_category)
-      path = conn |> user_category_path(:delete, user_category)
-      assert conn |> delete(path) |> response(204)
+      assert conn |> request_delete(user_category) |> response(204)
     end
```
amyschools, line 69: Don't forget to take out line 75 as discussed in #411

sbatson5 (approved):
This LGTM 👍 . These tests are so much more readable now!

test/controllers/user_category_controller_test.exs
```diff
@@ -1,60 +1,40 @@
...
     test "filters resources on index", %{conn: conn} do
-      society = insert(:category, name: "Society")
-      technology = insert(:category, name: "Technology")
-      government = insert(:category, name: "Government")
-
-      user = insert(:user)
-      user_category_1 = insert(:user_category, user: user, category: society)
-      user_category_2 = insert(:user_category, user: user, category: technology)
-      insert(:user_category, user: user, category: government)
+      [user_category_1, user_category_2] = insert_pair(:user_category)
```
begedin: What we want is a third user category. Then, we fetch by the two ids and ensure that only those two records were retrieved, but not the third one.

Consider what would happen if we remove filtering by a list of ids from our controller, so that the controller would simply return all user categories. Since we only have these two, the test would still pass, even though the feature it's testing no longer works.

begedin: A relatively elegant way to do it would be to just replace the line with 

``` Elixir
[user_category_1, user_category_2 | _] = insert_list(3, :user_category)
```

begedin (changes requested):
Just a minor change. Then it looks great.

test/controllers/user_category_controller_test.exs
amyschools: @begedin this is a cool solution! thanks!

begedin (approved):
👍 Once tests pass, feel free to merge

Don't forget to squash.

begedin:
And don't worry about coveralls here. Based on the report, the drop in coverage was not really affected by changes here.

test/controllers/user_category_controller_test.exs
```diff
@@ -63,65 +43,43 @@ defmodule CodeCorps.UserCategoryControllerTest do
...
-      assert json["errors"] != %{}
+    test "renders 422 when data is invalid", %{conn: conn} do
+      assert conn |> request_create |> json_response(422)
     end
 
-    test "does not create resource and renders 401 when unauthenticated", %{conn: conn} do
-      path = conn |> user_category_path(:create)
-      assert conn |> post(path) |> json_response(401)
+    test "renders 401 when unauthenticated", %{conn: conn} do
+      assert conn |> request_create(%{}) |> json_response(401)
```
joshsmith: The `403` does not have the `%{}` so maybe remove here?

mackenziehicks: oh yeah, must have missed this one

joshsmith (approved):

joshsmith:
Rebase and squash!

PR #420, opened 2016-11-02
Refactoring Project Category Controller Tests

### LINKED ISSUE #418: Refactor Project Category controller tests
# Problem
Add helpers from api_case.ex to Project Category Controller Tests

## References 

Progress on: #413

amyschools:
# What's in this PR?

Adding api_case.ex helpers.

## References
Closes #418 

Progress on: #413

### DIFF
--- a/test/controllers/project_category_controller_test.exs
+++ b/test/controllers/project_category_controller_test.exs
@@ -1,47 +1,25 @@
 defmodule CodeCorps.ProjectCategoryControllerTest do
-  use CodeCorps.ApiCase
-
-  alias CodeCorps.Category
-  alias CodeCorps.Project
-  alias CodeCorps.ProjectCategory
-  alias CodeCorps.Repo
-
-  @attrs %{}
-
-  defp build_payload, do: %{ "data" => %{"type" => "project-category", "attributes" => %{}}}
+  use CodeCorps.ApiCase, resource_name: :project_category
 
   describe "index" do
     test "lists all entries on index", %{conn: conn} do
-      conn = get conn, project_category_path(conn, :index)
-      assert json_response(conn, 200)["data"] == []
+      [project_category_1, project_category_2] = insert_pair(:project_category)
+
+      conn
+      |> request_index
+      |> json_response(200)
+      |> assert_ids_from_response([project_category_1.id, project_category_2.id])
     end
 
     test "filters resources on index", %{conn: conn} do
-      arts = insert(:category, name: "Arts")
-      society = insert(:category, name: "Society")
-      technology = insert(:category, name: "Technology")
+      [project_category_1, project_category_2 | _] = insert_list(3, :project_category)
 
-      project = insert(:project)
-      project_category_1 = insert(:project_category, project: project, category: arts)
-      project_category_2 = insert(:project_category, project: project, category: society)
-      insert(:project_category, project: project, category: technology)
+      path = "project-categories/?filter[id]=#{project_category_1.id},#{project_category_2.id}"
 
-      response =
-        conn
-        |> get("project-categories/?filter[id]=#{project_category_1.id},#{project_category_2.id}")
-        |> json_response(200)
-
-      [first_result, second_result] = response |> Map.get("data")
-
-      first_result
-      |> assert_result_id(project_category_1.id)
-      |> assert_jsonapi_relationship("project", project.id)
-      |> assert_jsonapi_relationship("category", arts.id)
-
-      second_result
-      |> assert_result_id(project_category_2.id)
-      |> assert_jsonapi_relationship("project", project.id)
-      |> assert_jsonapi_relationship("category", society.id)
+      conn
+      |> get(path)
+      |> json_response(200)
+      |> assert_ids_from_response([project_category_1.id, project_category_2.id])
     end
   end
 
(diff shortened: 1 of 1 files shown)


test/controllers/project_category_controller_test.exs
```diff
@@ -1,47 +1,25 @@
...
-      response =
-        conn
-        |> get("project-categories/?filter[id]=#{project_category_1.id},#{project_category_2.id}")
-        |> json_response(200)
-
-      [first_result, second_result] = response |> Map.get("data")
-
-      first_result
-      |> assert_result_id(project_category_1.id)
-      |> assert_jsonapi_relationship("project", project.id)
```
sbatson5, line 38: Do we no longer want to test the relationships are valid?

test/controllers/project_category_controller_test.exs
```diff
@@ -52,106 +30,62 @@ defmodule CodeCorps.ProjectCategoryControllerTest do
...
 
     @tag authenticated: :admin
-    test "does not create resource and renders 422 when data is invalid", %{conn: conn} do
-      payload = build_payload()
-
-      path = conn |> project_category_path(:create)
-      data = conn |> post(path, payload) |> json_response(422)
-      assert data["errors"] != %{}
+    test "renders 422 when data is invalid", %{conn: conn} do
+      invalid_attrs = %{}
```
sbatson5, line 57: Do we need to assign `invalid_attrs`?  On the two tests below we simply call `request_create(%{})`.  Thoughts on removing this or possibly declaring `@invalid_attrs` up top and reusing it in the tests below as well?

sbatson5 (commented):
Just a couple of questions, but this looks great! 👍 :shipit:

test/controllers/project_category_controller_test.exs
amyschools: I think since they are already tested in the model test, we decided to take them out of the controller tests.

test/controllers/project_category_controller_test.exs
amyschools: In the other controller tests we were declaring @invalid_attrs at the top and using it in the 422 test, but since it's the only place we end up using it @JoshSmith said this might be a little cleaner.

I think this difference with the `request_create` for the 401 and 403 tests we don't even hit the attributes before it halts with an error, so an empty attribute map is kind of just placeholding instead of being invalid.

test/controllers/project_category_controller_test.exs
amyschools: Do you agree with that?

sbatson5: That makes sense to me. Thanks for clarifying!

test/controllers/project_category_controller_test.exs
sbatson5: Gotcha. Do we need to even pass anything in the tests below, then? Can we just do `conn |> request_create |> json_response(401)` ?

amyschools: @mackenziehicks and I were asking all the same questions yesterday! In the api_case helper for `request_create` you have to pass in attributes, unlike for request_delete where you have an option to just use the conn.

```
      def request_create(conn, attrs) do
        path = conn |> path_for(:create)
        payload = json_payload(factory_name, attrs)
        conn |> post(path, payload)
      end
```

```
      def request_delete(conn), do: conn |> request_delete(default_record)
      def request_delete(conn, :not_found), do: conn |> request_delete(-1)
      def request_delete(conn, resource_or_id) do
        path = conn |> path_for(:delete, resource_or_id)
        conn |> delete(path)
      end
    end
  end
```

sbatson5: Ah, thanks for explaining!

What are your thoughts on adding a default to `request_create`?

`def request_create(conn, attrs \\ %{})`?

joshsmith: Ah I'm dumb and hadn't considered this. Yes, this makes sense.

mackenziehicks: @JoshSmith  I think we should add it for update too. 

Right now the test looks like this:
`assert conn |> request_update({@valid_attrs}) |> json_response(401)`
but we haven't actually added anything to update, so do we need to pass in `@valid_attrs`?

joshsmith: Yeah, I'd add it for `update`, too.

test/controllers/project_category_controller_test.exs
```diff
@@ -1,47 +1,25 @@
...
 
     test "filters resources on index", %{conn: conn} do
-      arts = insert(:category, name: "Arts")
-      society = insert(:category, name: "Society")
       technology = insert(:category, name: "Technology")
-
       project = insert(:project)
-      project_category_1 = insert(:project_category, project: project, category: arts)
-      project_category_2 = insert(:project_category, project: project, category: society)
+      [project_category_1, project_category_2] = insert_pair(:project_category)
```
begedin: Same as [comment](https://github.com/code-corps/code-corps-api/pull/425/files#r86388037), while this is good, it could be rewritten in one line as 

``` Elixir
[project_category_1, project_category_2 | _] = insert_list(3, : project_category_1)
```

We don't really need to pre-insert the category, or the project. The factory should do that part for us.

begedin (changes requested):
Minor change and then good for me.

begedin (approved):

PR #434, opened 2016-11-03
Refactor Organization Controller Tests

### LINKED ISSUE #430: Refactor Organization Controller Tests
# Problem

Add api_case.ex helpers to controller tests.

## References

Progress on: #413

amyschools:
# What's in this PR?

Added api_case.ex test helpers.

## References
Closes #430 

Progress on: #413

### DIFF
--- a/test/controllers/organization_controller_test.exs
+++ b/test/controllers/organization_controller_test.exs
@@ -1,34 +1,28 @@
 defmodule CodeCorps.OrganizationControllerTest do
-  use CodeCorps.ApiCase
-
-  alias CodeCorps.Organization
-  alias CodeCorps.Repo
-  alias CodeCorps.SluggedRoute
+  use CodeCorps.ApiCase, resource_name: :organization
 
   @valid_attrs %{description: "Build a better future.", name: "Code Corps"}
   @invalid_attrs %{name: ""}
 
-  defp build_payload, do: %{ "data" => %{"type" => "organization"}}
-
   describe "index" do
     test "lists all entries on index", %{conn: conn} do
-      path = conn |> organization_path(:index)
-      conn = conn |> get(path)
+      [organization_1, organization_2] = insert_pair(:organization)
 
-      assert json_response(conn, 200)["data"] == []
+      conn
+      |> request_index
+      |> json_response(200)
+      |> assert_ids_from_response([organization_1.id, organization_2.id])
     end
 
     test "filters resources on index", %{conn: conn} do
-      first_org = insert(:organization, name: "Org A")
-      second_org = insert(:organization, name: "Org B")
-      insert(:organization, name: "Org C")
+      [organization_1, organization_2 | _] = insert_list(3, :organization)
 
-      path = "organizations/?filter[id]=#{first_org.id},#{second_org.id}"
+      path = "organizations/?filter[id]=#{organization_1.id},#{organization_2.id}"
 
       conn
       |> get(path)
       |> json_response(200)
-      |> assert_ids_from_response([first_org.id, second_org.id])
+      |> assert_ids_from_response([organization_1.id, organization_2.id])
     end
   end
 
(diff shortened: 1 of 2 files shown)


test/controllers/organization_controller_test.exs
```diff
@@ -1,176 +1,86 @@
...
-        |> put_id(organization)
-        |> put_attributes(@invalid_attrs)
-
-      path = conn |> organization_path(:update, organization)
-      conn = conn |> put(path, payload)
-
-      assert json_response(conn, 403)
+    test "renders 403 when not authorized", %{conn: conn} do
+      assert conn |> request_create |> json_response(403)
     end
```
amyschools, line 85: @JoshSmith what should I do with the test below since I don't have the aws env keys? The build_payload on line 186 is causing the tests to fail.

joshsmith: Just noting that we talked about this after this comment, so think change is pending.

mackenziehicks (commented):
👍

test/support/api_case.ex
```diff
@@ -110,16 +110,17 @@ defmodule CodeCorps.ApiCase do
         conn |> post(path, payload)
       end
 
-      def request_update(conn, :not_found), do: conn |> request_update(-1, %{})
-      def request_update(conn, attrs \\ %{}), do: conn |> request_update(default_record, attrs)
+      def request_update(conn), do: request_update(conn, %{})
```
sbatson5, line 113: Rather than trying to cover the varying arity, what are your thoughts on keeping the default on line 115? Having `request_update/1` call `request_update/2` seems like the same idea as a default value.

`def request_update(conn, attrs // %{}), do: request_update(conn, default_record, attrs)`

test/support/api_case.ex
```diff
@@ -110,16 +110,17 @@ defmodule CodeCorps.ApiCase do
...
+      def request_update(conn, attrs), do: request_update(conn, default_record, attrs)
       def request_update(conn, resource_or_id, attrs) do
         payload = json_payload(factory_name, attrs)
         path = conn |> path_for(:update, resource_or_id)
         conn |> put(path, payload)
       end
 
-      def request_delete(conn), do: conn |> request_delete(default_record)
-      def request_delete(conn, :not_found), do: conn |> request_delete(-1)
+      def request_delete(conn), do: request_delete(conn, default_record)
```
sbatson5, line 122: 👏 I like that we aren't piping this anymore.  I find not piping more readable when there is only one method call.

test/controllers/organization_controller_test.exs
```diff
@@ -1,202 +1,97 @@
...
-      assert data["id"] == "#{organization.id}"
-      assert data["type"] == "organization"
-      assert data["attributes"]["name"] == organization.name
-      assert data["attributes"]["description"] == organization.description
-      assert data["attributes"]["slug"] == organization.slug
+      conn
+      |> request_show(organization)
+      |> json_response(200)
+      |> Map.get("data")
+      |> assert_result_id(organization.id)
```
sbatson5, line 37: 🤔 When we call `result_id`, we seem to have to retrieve the data hash first, unlike `assert_ids_from_response`, which extracts it for us.  What are you thoughts on updating the `assert_result_id` helper to accept either the data list _or_ the full json response?

sbatson5 (commented):
Just a couple of questions.  This looks really great! +40, -144 is awesome!

test/controllers/organization_controller_test.exs
joshsmith: Agreed on this.

test/support/api_case.ex
joshsmith: @sbatson5 I believe this was what we had and were receiving a warning for it because of line 114.

test/support/api_case.ex
joshsmith: Agreed, noticed that the other day so we made the change.

test/controllers/organization_controller_test.exs
joshsmith: @amyschools can you open an issue for this?

joshsmith (approved):
Approved but a comment here about opening an issue. Also not sure if @sbatson5 agrees with my thinking on his other note.

test/controllers/organization_controller_test.exs
amyschools: Done! #445

PR #448, opened 2016-11-08
Refactor Task Controller Tests with helpers

### LINKED ISSUE #442: Refactor Task Controller tests with helpers
# Problem

Add api_case.ex helpers to Task controller tests.

## References

Progress on: #413

amyschools:
# What's in this PR?

Add api_case.ex helpers to Task controller tests.

Stuck on a fix to make two tests pass: update: create and update: renders 422 error. 

## References
Closes #442 

Progress on: #413

### DIFF
--- a/test/controllers/task_controller_test.exs
+++ b/test/controllers/task_controller_test.exs
@@ -1,8 +1,5 @@
 defmodule CodeCorps.TaskControllerTest do
-  use CodeCorps.ApiCase
-
-  alias CodeCorps.Task
-  alias CodeCorps.Repo
+  use CodeCorps.ApiCase, resource_name: :task
 
   @valid_attrs %{
     title: "Test task",
@@ -12,17 +9,19 @@ defmodule CodeCorps.TaskControllerTest do
   }
 
   @invalid_attrs %{
+    title: nil,
     task_type: "issue",
     status: "nonexistent"
   }
 
-  defp build_payload, do: %{ "data" => %{"type" => "task"}}
-
   describe "index" do
     test "lists all entries", %{conn: conn} do
-      path = conn |> task_path(:index)
-      json = conn |> get(path) |> json_response(200)
-      assert json["data"] == []
+      [task_1, task_2] = insert_pair(:task)
+
+      conn
+      |> request_index
+      |> json_response(200)
+      |> assert_ids_from_response([task_1.id, task_2.id])
     end
 
     test "lists all entries newest first", %{conn: conn} do
@@ -112,18 +111,12 @@ defmodule CodeCorps.TaskControllerTest do
   describe "show" do
     test "shows chosen resource", %{conn: conn} do
       task = insert(:task)
-      path = conn |> task_path(:show, task)
 
-      data = conn |> get(path) |> json_response(200) |> Map.get("data")
-
-      assert data["id"] == "#{task.id}"
-      assert data["type"] == "task"
-      assert data["attributes"]["body"] == task.body
-      assert data["attributes"]["markdown"] == task.markdown
-      assert data["attributes"]["number"] == task.number
-      assert data["attributes"]["task-type"] == task.task_type
-      assert data["attributes"]["status"] == task.status
-      assert data["attributes"]["title"] == task.title
+      conn
+      |> request_show(task)
+      |> json_response(200)
+      |> Map.get("data")
+      |> assert_result_id(task.id)
     end
 
     test "shows task by number for project", %{conn: conn} do
@@ -136,9 +129,8 @@ defmodule CodeCorps.TaskControllerTest do
       assert data["type"] == "task"
     end
 
-    test "does not show resource and instead throw error when id is nonexistent", %{conn: conn} do
-      path = conn |> task_path(:show, -1)
-      assert conn |> get(path) |> json_response(:not_found)
+    test "renders 404 when id is nonexistent", %{conn: conn} do
+      assert conn |> request_show(:not_found) |> json_response(404)
     end
   end
 
@@ -146,16 +138,8 @@ defmodule CodeCorps.TaskControllerTest do
     @tag :authenticated
     test "creates and renders resource when data is valid", %{conn: conn, current_user: current_user} do
       project = insert(:project)
-
-      payload =
-        build_payload
-        |> put_attributes(@valid_attrs)
-        |> put_relationships(current_user, project)
-
-      path = conn |> task_path(:create)
-      json = conn |> post(path, payload) |> json_response(201)
-
-      assert json["data"]["id"]
+      attrs = @valid_attrs |> Map.merge(%{project: project, user: current_user})
+      json = conn |> request_create(attrs) |> json_response(201)
 
       # ensure record is reloaded from database before serialized, since number is added
       # on database level upon insert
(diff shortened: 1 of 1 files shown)


test/controllers/task_controller_test.exs
```diff
@@ -16,13 +13,14 @@ defmodule CodeCorps.TaskControllerTest do
...
-  defp build_payload, do: %{ "data" => %{"type" => "task"}}
-
   describe "index" do
     test "lists all entries", %{conn: conn} do
-      path = conn |> task_path(:index)
-      json = conn |> get(path) |> json_response(200)
-      assert json["data"] == []
+      [task_1, task_2] = insert_pair(:task)
+
+    conn
```
sbatson5: Looks like the indentation is off here.

test/controllers/task_controller_test.exs
```diff
@@ -147,91 +138,44 @@ defmodule CodeCorps.TaskControllerTest do
...
-        build_payload
-        |> put_attributes(@valid_attrs)
-        |> put_relationships(current_user, project)
-
-      path = conn |> task_path(:create)
-      json = conn |> post(path, payload) |> json_response(201)
-
-      assert json["data"]["id"]
+      attrs = @valid_attrs |> Map.merge(%{project: project, user: current_user})
+      assert conn |> request_create(attrs) |> json_response(201)]
```
sbatson5: Looks like there is a trailing bracket here.

sbatson5 (commented):
This is looking good 👍

begedin:
@amyschools What difficulty are you having with the tests? Would you mind fixing the extra bracket on line 142? Can't really see which tests are failing when the app doesn't compile.

amyschools:
@begedin it should compile now. I'm getting a 403 error for three of the tests:   

`1) test create renders 422 when data is invalid (CodeCorps.TaskControllerTest)`  
`2) test update renders 422 when data is invalid (CodeCorps.TaskControllerTest)`  
`3) test create creates and renders resource when data is valid (CodeCorps.TaskControllerTest)` 

and a new error saying key :task is not recognized that I thought I had solved yesterday on 

`4) test update updates and renders chosen resource when data is valid (CodeCorps.TaskControllerTest)`

begedin:
@amyschools Looks like you didn't push up, or at least, I'm not seeing the changes. I'll try to reply best I can.

For the update test, we have the following: 

``` Elixir
@tag authenticated: :admin # an admin user is performing this test
test "updates and renders chosen resource when data is valid", %{conn: conn, current_user: current_user} do
  task = insert(:task, user: current_user)

  # you're not sending the task to `request_update`, so it doesn't do anything
  assert conn |> request_update(@valid_attrs) |> json_response(200)
end

@tag :authenticated # notice the different tag. a regular user is performing
test "renders 422 when data is invalid", %{conn: conn, current_user: current_user} do
  assert conn |> request_update(@invalid_attrs) |> json_response(422)
end
```

For the first test, it's tagged to authenticate as admin.  

You create a task, but then create the version of the `request_update` function that doesn't accept a task and instead creates its own resource. The tests pass, since you're authenticated as site admin, but the correct approach would be either to 

### Option A

Remove the `task = insert(:task, user: current_user)`line (and also remove the current user from the setup hash in the test signature), 

### Option B

Keep the task line, tag as `:authenticated`(so not `authenticated: :admin`) and then call

``` Elixir
assert conn |> request_update(@valid_attrs) |> json_response(200)
```

My vote would be option B, since as a side-effect, it's testing a more complex case of authorization.

For the second test, as you can see, you're not authenticating as admin, which is why it's failing. If you go with option B, then the solution here would be to basically copy everything from the first test and just replace `@valid_attrs` with `@invalid_attrs`. That should do the trick, so you would end up with:

``` Elixir
@tag :authenticated
test "renders 422 when data is invalid", %{conn: conn, current_user: current_user} do
  task = insert(:task, user: current_user)
  assert conn |> request_update(task, @invalid_attrs) |> json_response(422)
end
```

For create, the problem is similar, but not quite the same

```
assert conn |> request_create(@invalid_attrs) |> json_response(422)
```

`@invalid_attrs` doesn't have a user_id or a project id assigned. The authorization runs, tries to see if the user is the author of the task and figures out it not (philosophical, really 😄 ) and concludes that you are not authorized.

In order to get it working, you mimic the valid request approach:

``` Elixir
project = insert(:project)
attrs = @valid_attrs |> Map.merge(%{project: project, user: current_user})
assert conn |> request_create(attrs) |> json_response(201)]
```

The difference being, you replace `@valid_attrs` with `@invalid_attrs`.

``` Elixir
project = insert(:project)
attrs = @invalid_attrs |> Map.merge(%{project: project, user: current_user})
assert conn |> request_create(attrs) |> json_response(422)
```

It's a bit complicated, but basically, in all our cases, we run authorization before we validate the record. That means that, to test validation error responses, you want to set up your request so authorization passes, but validation fails after. On some controllers, that isn't even possible, but in this one, it is, albeit it gets a bit complicated.

It might make many things simpler, but I'm not sure.

For your create and update valid request tests, it seems to me like they should work, from the changes here. However, it may be you altered something, so I'll take another look when you push your changes up.

@JoshSmith Writing all this, maybe we should at one point look into our process being something like:

```
get_record_if_applicable
initialize_changeset_for_record_and_attributes
respond_with_422_if_changeset_invalid
authorize_if_changeset_valid
respond_with_403_if_unauthorized
respond_with_ok_if_authorized
```

amyschools:
@begedin Thank you so much!! That explanation was super helpful. All the tests are passing now so maybe one last check if you have time?

test/controllers/task_controller_test.exs
```diff
@@ -12,17 +9,19 @@ defmodule CodeCorps.TaskControllerTest do
...
     test "lists all entries", %{conn: conn} do
-      path = conn |> task_path(:index)
-      json = conn |> get(path) |> json_response(200)
-      assert json["data"] == []
+    [task_1, task_2] = insert_pair(:task)
+
+    conn
+    |> request_index
+    |> json_response(200)
+    |> assert_ids_from_response([task_1.id, task_2.id])
```
begedin: Indentation here is bad. Which editor are you using, and does it have an `.editorconfig` plugin? Should prevent this from happening. 

Feel free to mention it on slack. Someone can guide you through configuring an editor when we get the chance.

test/controllers/task_controller_test.exs
```diff
@@ -136,102 +129,55 @@ defmodule CodeCorps.TaskControllerTest do
...
-        |> put_id(task.id)
-        |> put_attributes(@valid_attrs)
-
-      path = conn |> task_path(:update, task)
-      json = conn |> put(path, payload) |> json_response(200)
-
-      assert json["data"]["id"]
-      assert Repo.get_by(Task, @valid_attrs)
+    task = insert(:task, user: current_user)
+    assert conn |> request_update(task, @valid_attrs) |> json_response(200)
```
begedin: More indentation problems here.

begedin (dismissed):
Just some indentation problems that need fixing. Looks good otherwise.

test/controllers/task_controller_test.exs
amyschools: just installed the editorconfig package for atom!

joshsmith:
@amyschools this had a review with changes that need made.

amyschools:
pushed to the wrong branch! should be good now.

The 'close pull request' button is very close to the comment button...

joshsmith (approved):

joshsmith:
@amyschools awesome, rebase, squash, and merge.

Issue #447, opened 2016-11-07
Write policies for Stripe resources


joshsmith:
# Problem

We should write policies for all of our Stripe-related resources: accounts, cards, customers, plans, and subscriptions.

## Subtasks

1.  Look at existing policies to learn how they're written
2. Create issues for each of the policies we need to create

- [x] StripeConnectAccount #451 
- [ ] StripePlatformCard #454 
- [x] StripePlatformCustomer #453 
- [ ] StripeConnectPlan
- [ ] StripeConnectSubscription #457 

For each policy:

- [ ] Look at the blueprint to determine which actions you need to authorize on the resource
- [ ] Write out how you expect the policy to work inside its issue
- [ ] Write the policy tests
- [ ] Write the policy

## References

Progress on: #10

joshsmith:
Closing in favor of actual issues.

PR #451, opened 2016-11-08
Write Policy for Stripe Account

### LINKED ISSUE #450: Write Policy for Stripe Account
# Problem

Write policy, policy tests, and add Stripe Account to ability.ex. 

A User will need Organization admin role to see the Index and Show. 

## References

Progress on: #447

amyschools:
# What's in this PR?

Added Stripe Account policy to ability.ex, stripe_account_policy.ex, factory, and tests to stripe_account_policy_test.

Currently on hold until the index: filter question has been resolved. See comments for more information.

## References
Fixes #450 

Progress on: #447

### DIFF
--- a/test/policies/stripe_account_policy_test.exs
+++ b/test/policies/stripe_account_policy_test.exs
@@ -1,0 +1,64 @@
+defmodule CodeCorps.StripeAccountPolicyTest do
+  use CodeCorps.PolicyCase
+
+  import CodeCorps.StripeAccount
+  import CodeCorps.StripeAccountPolicy, only: [show?: 2]
+
+  alias CodeCorps.StripeAccount
+
+  describe "show?" do
+    test "returns true when user is an admin" do
+      user = build(:user, admin: true)
+      stripe_account = insert(:stripe_account)
+
+      assert show?(user, stripe_account)
+    end
+
+    test "returns true when user is owner of organization" do
+      user = insert(:user)
+      organization = insert(:organization)
+      insert(:organization_membership, role: "owner", member: user, organization: organization)
+
+      stripe_account = insert(:stripe_account, organization: organization)
+
+      assert show?(user, stripe_account)
+    end
+
+    test "returns false when user is admin of organization" do
+      user = insert(:user)
+      organization = insert(:organization)
+      insert(:organization_membership, role: "admin", member: user)
+
+      stripe_account = insert(:stripe_account, organization: organization)
+
+      refute show?(user, stripe_account)
+    end
+
+    test "returns false when user is not member of organization" do
+      user = insert(:user)
+      stripe_account = insert(:stripe_account)
+
+      refute show?(user, stripe_account)
+    end
+
+    test "returns false when user is pending member of organization" do
+      user = insert(:user)
+      organization = insert(:organization)
+      insert(:organization_membership, role: "pending", member: user)
+
+      stripe_account = insert(:stripe_account, organization: organization)
+
+      refute show?(user, stripe_account)
+    end
+
+    test "returns false when user is contributor of organization" do
+      user = insert(:user)
+      organization = insert(:organization)
+      insert(:organization_membership, role: "contributor", member: user)
+
+      stripe_account = insert(:stripe_account, organization: organization)
+
+      refute show?(user, stripe_account)
+    end
+  end
+end
--- a/web/models/abilities.ex
+++ b/web/models/abilities.ex
@@ -12,6 +12,7 @@ defmodule Canary.Abilities do
   alias CodeCorps.Role
   alias CodeCorps.RoleSkill
   alias CodeCorps.Skill
+  alias CodeCorps.StripeAccount
   alias CodeCorps.StripeCustomer
   alias CodeCorps.User
   alias CodeCorps.UserCategory
@@ -91,6 +92,8 @@ defmodule Canary.Abilities do
     def can?(%User{} = user, :delete, %RoleSkill{}), do: RoleSkillPolicy.delete?(user)
 
     def can?(%User{} = user, :create, Skill), do: SkillPolicy.create?(user)
+
+    def can?(%User{} = user, :show, %StripeAccount{}=stripe_account), do: StripeAccountPolicy.show?(user, stripe_account)
 
     def can?(%User{} = user, :create, %Changeset{data: %StripeCustomer{}} = changeset), do: StripeCustomerPolicy.create?(user, changeset)
     def can?(%User{} = user, :show, %StripeCustomer{} = stripe_customer), do: StripeCustomerPolicy.show?(user, stripe_customer)
--- a/web/policies/stripe_account_policy.ex
+++ b/web/policies/stripe_account_policy.ex
@@ -1,0 +1,13 @@
+defmodule CodeCorps.StripeAccountPolicy do
+  import CodeCorps.Helpers.Policy,
+    only: [get_membership: 2, get_role: 1, owner?: 1]
+
+  alias CodeCorps.StripeAccount
+  alias CodeCorps.User
+  alias CodeCorps.OrganizationMembership
+  alias Ecto.Changeset
+
+  def show?(%User{admin: true}, %StripeAccount{}), do: true
+  def show?(%User{} = user, %StripeAccount{} = stripe_account), do: stripe_account |> get_membership(user) |> get_role |> owner?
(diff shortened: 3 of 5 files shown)


test/policies/stripe_account_policy_test.exs
```diff
@@ -0,0 +1,146 @@
+defmodule CodeCorps.StripeAccountPolicyTest do
+  use CodeCorps.PolicyCase
+
+  import CodeCorps.StripeAccountPolicy, only: [index?: 2, show?: 2]
+  import CodeCorps.StripeAccount, only: [create_changeset: 2]
```
joshsmith: Don't think you need to import this since we're not using that changeset.

test/policies/stripe_account_policy_test.exs
```diff
@@ -0,0 +1,146 @@
...
+      refute show?(user, organization)
+    end
+
+    test "returns true when user is admin of organization" do
+      user = insert(:user)
+      organization = insert(:organization)
+
+      insert(:organization_membership, role: "admin", member: user, organization: organization)
+
+      assert show?(user, organization)
```
joshsmith: I think we just want owner at the lowest level of privilege here.

I would also consider grouping your trues and your falses together.

test/policies/stripe_account_policy_test.exs
```diff
@@ -0,0 +1,146 @@
...
+  #
+  #   test "returns false when user is owner of organization" do
+  #     user = insert(:user)
+  #     organization = insert(:organization)
+  #
+  #     insert(:organization_membership, role: "owner", member: user, organization: organization)
+  #
+  #     assert stripe_auth?(user, organization)
+  #   end
+  # end
```
joshsmith: This isn't needed here.

web/policies/stripe_account_policy.ex
```diff
@@ -0,0 +1,14 @@
+defmodule CodeCorps.StripeAccountPolicy do
+  import CodeCorps.Helpers.Policy,
+    only: [get_membership: 2, get_role: 1, admin_or_higher?: 1, owner?: 1]
```
joshsmith: This is probably like this in other policies, but would alpha order the fns here.

web/policies/stripe_account_policy.ex
```diff
@@ -0,0 +1,14 @@
...
+
+  alias CodeCorps.StripeAccount
+  alias CodeCorps.User
+  alias Ecto.Changeset
+
+  def index?(%User{admin: true}, %StripeAccount{}), do: true
+  def index?(%User{} = user, %StripeAccount{} = stripe_account), do: stripe_account |> get_membership(user) |> get_role |> admin_or_higher?
```
joshsmith: For here, I'm not sure yet until we decide what we're doing on `filter` and I get some weigh-in with @begedin. This is likely going to be more complex than that.

web/policies/stripe_account_policy.ex
```diff
@@ -0,0 +1,14 @@
...
+
+  def show?(%User{admin: true}, %StripeAccount{}), do: true
+  def show?(%User{} = user, %StripeAccount{} = stripe_account), do: stripe_account |> get_membership(user) |> get_role |> admin_or_higher?
```
joshsmith: I would do this as `owner` instead, which means you may need a `false` case (I don't remember if there's a convenience function that returns a `true` or `false` for this inside of the policy helpers.)

joshsmith (commented):
A few comments here.

web/policies/stripe_account_policy.ex
begedin: I'll contact Alan, but it seems like the convention here is to use a scope.

Bodyguard offers [scopes](https://github.com/schrockwell/bodyguard#policy-scopes), while, sadly, our current canary offers no such thing.

Since we're using `ja_resource`, perhaps a `def records, do: records_current_user_is_authorized_for` would be the correct approach? If we ever switch to bodyguard, maybe later on it would make sense to move that code into a `scope`.

Looking at the API blueprint, it doesn't look an `organization_id` parameter is required. Should we make it a requirement?

We would probably have something like a records callback matching a request without an org id, which returns a zero records, as well as one with an org id, which returns zero records if current user is not org owner, and all records for current org if user is org owner. At that point, we can have an `index?` that accepts a current user and an organization or keyword list with an organization id.

Really, even without contacting Alan, it looks to me like scopes would be the way to go and this is as close as we can get to a scope, without switching from canary to a bodyguard. Really, it's a scope in anything other than the name.

begedin:

amyschools:
@begedin or anyone else who reviews this do you know why I would be getting this error on the test below? I've been trying adding in an organization, changing insert to build but I don't think I am on the right track...

```
    test "returns true when user is owner of organization" do
      user = insert(:user)
      stripe_account = insert(:stripe_account)

      insert(:organization_membership, role: "owner", member: user)

      assert show?(user, stripe_account)
    end
```

```
  1) test show? returns true when user is owner of organization (CodeCorps.StripeAccountPolicyTest)
     test/policies/stripe_account_policy_test.exs:17
     Expected truthy, got false
     code: show?(user, stripe_account)
     stacktrace:
       test/policies/stripe_account_policy_test.exs:23: (test)
```

begedin:
@amyschools 

The code you have creates a user and an owner level membership for an unspecified organization. It also creates a stripe account for a completely different unspecified organization. What you should be doing is to create a common organization for both of these records:

``` Elixir
user = insert(:user)
insert(:organization_membership, role: "owner", member: user, organization: organization)

stripe_account = insert(:stripe_account, organization: organization)

assert show?(user, stripe_account)
```

At that point, the test should pass.

web/models/abilities.ex
```diff
@@ -90,6 +91,8 @@ defmodule Canary.Abilities do
 
     def can?(%User{} = user, :create, Skill), do: SkillPolicy.create?(user)
 
+    def can?(%User{} = user, :show, StripeAccount), do: StripeAccountPolicy.show?(user)
```
begedin: We need a `can?` method here which will call `StripeAccountPolicy.show?(user, stripe_account)` instead of this one.

```
def can?(%User{} = user, :show, %StripeAccount{}=stripe_account), do: StripeAccountPolicy.show?(user, stripe_account)
```

This tells me that we ought to be testing the whole ability model. I had the same bug creep up on me in my own PR.

test/policies/stripe_account_policy_test.exs
```diff
@@ -0,0 +1,61 @@
...
+    end
+
+    test "returns false when user is admin of organization" do
+      user = insert(:user)
+      stripe_account = insert(:stripe_account)
+
+      insert(:organization_membership, role: "admin", member: user)
+
+      refute show?(user, stripe_account)
+    end
```
begedin, line 35: In the same way you'fe fixed the "returns true when user is owner of organization" test, you should fix all the other tests. 

In all of them, a membership is being created for the user in an organization that is unrelated to the stripe account organization. Really, while the tests say "returns false when user is X of organization", what they're all testing right now is "returns false when when user is not a member of organization".

begedin (dismissed):

web/models/abilities.ex
```diff
@@ -91,7 +91,7 @@ defmodule Canary.Abilities do
 
     def can?(%User{} = user, :create, Skill), do: SkillPolicy.create?(user)
 
-    def can?(%User{} = user, :show, StripeAccount), do: StripeAccountPolicy.show?(user)
+    def can?(%User{} = user, :show, %StripeAccount{}=stripe_account), do: StripeAccountPolicy.show?(user, stripe_account)
```
joshsmith, line 96: There should be spacing around the `=` in `%StripeAccount{}=stripe_account`

test/policies/stripe_account_policy_test.exs
```diff
@@ -42,19 +43,21 @@ defmodule CodeCorps.StripeAccountPolicyTest do
...
     end
 
     test "returns false when user is contributor of organization" do
       user = insert(:user)
-      stripe_account = insert(:stripe_account)
-
+      organization = insert(:organization)
       insert(:organization_membership, role: "contributor", member: user)
 
+      stripe_account = insert(:stripe_account, organization: organization)
```
joshsmith, line 59: Think we can remove the spacing between this and `insert` on all tests.

joshsmith (approved):

Issue #454, opened 2016-11-10
Write Policy for StripePlatformCard


mackenziehicks:
# Problem

Write policy, policy tests, and add Stripe Card to ability.ex.

A User will need to be authenticated and authorized and only be able to see/create their own card(s) for create and show.  

## References

Progress on: #447

PR #455, opened 2016-11-10
Write policies and endpoints for StripeCustomer

### LINKED ISSUE #453: Write Policy for Stripe Customer
# Problem

This resource identifies a customer created on the [Stripe API](https://stripe.com/docs/api#customers).

Endpoints for this resource are `create` and  `show` for an individual Customer. A user must be authorized for both and they can only view or create their own record.

## Subtasks
- [ ] Stripe Customer factory - #455
- [ ] Stripe Customer policy for `show?` - #455
- [ ] Stripe Customer Policy tests for `show?` - #455
- [ ] Stripe Customer ability for `show?` - #455
- [ ] Stripe Customer policy for `create?`
- [ ] Stripe Customer Policy tests for `create?`
- [ ] Stripe Customer ability for `create?`

## References

Progress on: #447

amyschools:
# What's in this PR?

Added Stripe Customer policy for `index(filter_by_id)` and `show` endpoints to ability.ex, stripe_customer_policy.ex, factory, and tests to stripe_customer_policy_test.

## References
Progress on: #453 
Progress on: #447

### DIFF
--- a/lib/code_corps/stripe/adapters/stripe_customer.ex
+++ b/lib/code_corps/stripe/adapters/stripe_customer.ex
@@ -1,7 +1,29 @@
 defmodule CodeCorps.Stripe.Adapters.StripeCustomer do
-  import CodeCorps.MapUtils, only: [rename: 3]
+  import CodeCorps.MapUtils, only: [rename: 3, keys_to_string: 1]
 
-  def params_from_stripe(%{} = stripe_map) do
-    stripe_map |> rename("id", "id_from_stripe")
+  def to_params(%Stripe.Customer{} = customer) do
+    customer
+    |> Map.from_struct
+    |> Map.take([:created, :currency, :delinquent, :id, :email])
+    |> rename(:id, :id_from_stripe)
+    |> keys_to_string
+  end
+
+  @non_stripe_attribute_keys ["user_id"]
+
+  def add_non_stripe_attributes(%{} = params, %{} = attributes) do
+    attributes
+    |> get_non_stripe_attributes
+    |> add_to(params)
+  end
+
+  defp get_non_stripe_attributes(%{} = attributes) do
+    attributes
+    |> Map.take(@non_stripe_attribute_keys)
+  end
+
+  defp add_to(%{} = attributes, %{} = params) do
+    params
+    |> Map.merge(attributes)
   end
 end
--- a/lib/code_corps/stripe/in_memory.ex
+++ b/lib/code_corps/stripe/in_memory.ex
@@ -1,0 +1,24 @@
+defmodule CodeCorps.Stripe.InMemory do
+  defmodule Customer do
+    def create(map) do
+      {:ok, do_create(map)}
+    end
+
+    defp do_create(_) do
+      {:ok, created} = DateTime.from_unix(1479472835)
+
+      %Stripe.Customer{
+        id: "cus_9aMOFmqy1esIRE",
+        account_balance: 0,
+        created: created,
+        currency: "usd",
+        default_source: nil,
+        delinquent: false,
+        description: nil,
+        email: "mail@test.com",
+        livemode: false,
+        metadata: %{}
+      }
+    end
+  end
+end
--- a/test/policies/stripe_customer_policy_test.exs
+++ b/test/policies/stripe_customer_policy_test.exs
@@ -1,0 +1,28 @@
+defmodule CodeCorps.StripeCustomerPolicyTest do
+  use CodeCorps.PolicyCase
+
+  import CodeCorps.StripeCustomerPolicy, only: [show?: 2]
+
+  describe "show?" do
+    test "returns true when user is an admin" do
+      user = build(:user, admin: true)
+      stripe_customer = build(:stripe_customer)
+
+      assert show?(user, stripe_customer)
+    end
+
+    test "returns true when user is viewing their own information" do
+      user = insert(:user)
+      stripe_customer = insert(:stripe_customer, user: user)
+
+      assert show?(user, stripe_customer)
+    end
+
+    test "returns false when user id is not the StripeCustomer's user_id" do
+      [user, another_user] = insert_pair(:user)
+      stripe_customer = insert(:stripe_customer, user: user)
+
+      refute show?(another_user, stripe_customer)
+    end
+  end
+end
--- a/test/views/stripe_customer_view_test.exs
+++ b/test/views/stripe_customer_view_test.exs
@@ -1,0 +1,72 @@
+defmodule CodeCorps.StripeCustomerViewTest do
+  use CodeCorps.ConnCase, async: true
+
+  import Phoenix.View, only: [render: 3]
+
+  test "renders all attributes and relationships properly" do
+    user = insert(:user)
+    stripe_customer = insert(:stripe_customer, id_from_stripe: "some_id", email: "email", user: user)
+
+    rendered_json =  render(CodeCorps.StripeCustomerView, "show.json-api", data: stripe_customer)
+
+    expected_json = %{
+      "data" => %{
+        "attributes" => %{
+          "email" => "",
+          "created" => stripe_customer.created,
+          "currency" => stripe_customer.currency,
+          "delinquent" => stripe_customer.delinquent,
+          "id-from-stripe" => "",
+          "inserted-at" => stripe_customer.inserted_at,
+          "updated-at" => stripe_customer.updated_at
+        },
+        "id" => stripe_customer.id |> Integer.to_string,
+        "relationships" => %{
+          "user" => %{
+            "data" => %{"id" => user.id |> Integer.to_string, "type" => "user"}
+          }
+        },
+        "type" => "stripe-customer",
+      },
+      "jsonapi" => %{
+        "version" => "1.0"
+      }
+    }
+
+    assert rendered_json == expected_json
+  end
+
+  test "renders email and id_from_stripe when user is the authenticated user" do
+    user = insert(:user)
+    stripe_customer = insert(:stripe_customer, id_from_stripe: "some_id", email: "email", user: user)
+
+    conn = Phoenix.ConnTest.build_conn |> assign(:current_user, user)
+    rendered_json = render(CodeCorps.StripeCustomerView, "show.json-api", data: stripe_customer, conn: conn)
+    assert rendered_json["data"]["attributes"]["email"] == stripe_customer.email
+    assert rendered_json["data"]["attributes"]["id-from-stripe"] == stripe_customer.id_from_stripe
+  end
+
+  test "renders email and id_from_stripe for only the authenticated user when rendering list" do
+    stripe_customers = insert_list(4, :stripe_customer)
+    auth_customer = stripe_customers |> List.last
+
+    conn = Phoenix.ConnTest.build_conn |> assign(:current_user, auth_customer.user)
+    rendered_json = render(CodeCorps.StripeCustomerView, "show.json-api", data: stripe_customers, conn: conn)
+
+    emails =
+      rendered_json["data"]
+      |> Enum.map(&Map.get(&1, "attributes"))
+      |> Enum.map(&Map.get(&1, "email"))
+      |> Enum.filter(fn(email) -> email != "" end)
+
+    assert emails == [auth_customer.email]
+
+    stripe_ids =
+      rendered_json["data"]
+      |> Enum.map(&Map.get(&1, "attributes"))
+      |> Enum.map(&Map.get(&1, "id-from-stripe"))
+      |> Enum.filter(fn(id_from_stripe) -> id_from_stripe != "" end)
+
+    assert stripe_ids == [auth_customer.id_from_stripe]
+  end
+end
--- a/test/views/user_view_test.exs
+++ b/test/views/user_view_test.exs
@@ -6,6 +6,8 @@ defmodule CodeCorps.UserViewTest do
   test "renders all attributes and relationships properly" do
     user = insert(:user)
     organization_membership = insert(:organization_membership, member: user)
+    slugged_route = insert(:slugged_route, user: user)
+    stripe_customer = insert(:stripe_customer, user: user)
     user_category = insert(:user_category, user: user)
     user_role = insert(:user_role, user: user)
     user_skill = insert(:user_skill, user: user)
@@ -38,7 +40,10 @@ defmodule CodeCorps.UserViewTest do
             ]
           },
           "slugged-route" => %{
-            "data" => nil
+            "data" => %{"id" => slugged_route.id |> Integer.to_string, "type" => "slugged-route"}
+          },
+          "stripe-customer" => %{
+            "data" => %{"id" => stripe_customer.id |> Integer.to_string, "type" => "stripe-customer"}
           },
           "user-categories" => %{
             "data" => [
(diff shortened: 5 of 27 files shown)


test/policies/stripe_customer_policy_test.exs
```diff
@@ -0,0 +1,36 @@
...
+  describe "index?" do
+    test "returns true when user is an admin" do
+
+    end
+
+    test "returns true when user is viewing their own information" do
+
+    end
+
+    test "returns false when user id is not the StripeCustomer id" do
```
joshsmith: I think this is more when the user id is not the Stripe customer's `user_id`, right?

test/policies/stripe_customer_policy_test.exs
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  describe "show?" do
+    test "returns true when user is viewing their own information" do
+
+    end
+
+    test "returns true when user is an admin" do
+
+    end
```
joshsmith: Would put `admin` first like in the test above. Think about it as reducing privileges as you go.

begedin:

For this specific one, while it's on a good track already, the #456 PR can serve as an example, since the policy is similar.

amyschools:
@begedin similar question to #451 - I'm getting the  `Expected truthy, got false` error for this test:
`1) test show? returns true when user is viewing their own information (CodeCorps.StripeCustomerPolicyTest)`
The formatting is exactly the same as on #456 so I'm not sure what's wrong with it.

begedin:
@amyschools In this one, the tests seem to be OK, but your policy is wrong.

``` elixir
def show?(%User{admin: true}, %StripeCustomer{}), do: true
def show?(%User{} = user, %User{} = current_user, %StripeCustomer{} = stripe_customer), do: stripe_customer.id == current_user.id
def show?(%User{}, %StripeCustomer{}), do: false
```

The second one wont ever be used. Nothing is calling it.

What you need instead of that line is something like:

``` Elixir
def show?(%User{id: current_user_id}, %StripeCustomer{user_id: user_id}), do: current_user_id == user_id
```

begedin (changes requested):
This looks good, but it only handles the `show?` part of the issue #453 

I say it's good to merge, but upon merge, you should update the description of #453 so it shows that the `create?` part still needs to be done.

EDIT: Already did the update of description of #453, just make sure you check the respective boxes in the description upon merge.

EDIT: Was going to approve, but accidentally hit "request changes" instead. Corrected.

begedin (approved):

begedin:
@amyschools This needs a minor fix I missed last time. Since I need the endpoints to continue work on stripe, I'll take over and repurpose this PR. No need to work on it further, but good job up to this point.

begedin:
@JoshSmith This is still a work in progress, but the actual creation of stripe customer is done. What is left to handle is:
- [x] Right now, the client does not need to send any token. We simpl create a StripeCustomer and store it with a user_id reference, so the stripe request should not "validation-fail". It might have to request a token and it might have to fail, though. Could use your thoughts on that.
- [x] If the stripe request fails for some other reason, this is unhandled
- [x] We are expected to have a stripe api key in config/dev. It can be a test key, but I'm not sure how we should add it. We might need something like "dot_env", or use the "secret file" approach phoenix apps usually use

Once all of this is done, we can merge and move on to #454 

All of these changes should go together with code-corps/code-corps-ember#689

joshsmith:
@begedin did not get around to this today, but a couple notes:

Environment variables are already handled in `develop` in `config.exs`, like so:

``` elixir
config :stripity_stripe,
  api_key: System.get_env("STRIPE_SECRET_KEY"),
  connect_client_id: System.get_env("STRIPE_PLATFORM_CLIENT_ID")
```

Tests should probably not mock `stripity-stripe`. We should consider either writing a Bypass-oriented `stripity-stripe-testing` library that can be used here, or we should simply take a similar approach to our Segment API code as outlined in José's post here: http://blog.plataformatec.com.br/2015/10/mocks-and-explicit-contracts/

joshsmith:
I doubt we would send a token for creating a `StripeCustomer`. Can't see the utility in that. I think we probably want to eagerly create the customer the first time someone goes to donate (lands on the donation page?), or maybe even just after signing up? However it's done, it should be idempotent. And we do need to handle failure cases (API being down?)

begedin:
Update:
- Creation process works now. Had to update the stripe adapter for StripeCustomer so it properly converts from a `%Stripe.Customer{}` struct into a `%Map{}`.
- Without us realising, our switch to `JaResource` has made it so `ChangesetView` is no longer used for validation errors. Instead, the internal `errors.json-api` renderer is used, provided by `JaSerializer`. This works, but the errors no longer have a status field, so our client-side error formatter is no longer working properly. I made some changes here and made the required changes client-side to get it to a useful state.
- `stripity-stripe` wasn't handling 400 errors correctly, so I made a PR to fix that. Until that's merged, this PR will fail, unless we use switch our `mix.exs` to use the commit in the PR. I did that locally, but didn't want to push to this PR yet. The change is:

``` Elixir
{:stripity_stripe, "~> 2.0.0-alpha.1", git: "https://github.com/code-corps/stripity-stripe", ref: "cd1515989953da46f817de72549be5e431f2ae7d"}
```

The last thing left to do here is 
- [x] Write tests for `stripe_customer_controller`. This should be mergeable at that point. We probably need to implement something to mock stripe, similar to segment.

begedin:
Additional note/todo:
- [x] stripity-stripe should be returning a datetime for the `created` field. It's returning an integer right now.

joshsmith:
@begedin this is now returning a `DateTime` with `2.0.0-alpha.2`.

lib/code_corps/stripe/in_memory.ex
```diff
@@ -0,0 +1,24 @@
...
+        created: created,
+        currency: "usd",
+        default_source: nil,
+        delinquent: false,
+        description: nil,
+        email: "mail@test.com",
+        livemode: false,
+        metadata: %{}
+      }
+    end
```
joshsmith, line 22: I'm not sure this has gone far enough. In http://blog.plataformatec.com.br/2015/10/mocks-and-explicit-contracts/ it looks like the recommendation is to use behaviours and callbacks. That is, we seem to be lacking explicit contracts right now.

The boundary of testing the HTTP interactions should probably happen at the level of `stripity-stripe` and not here, so I don't think that will have to be added. But even in adjusting the small cases we have here, I ended up with the behavior getting out of sync because the contract is not explicit.

joshsmith: Further, now that I look at the effort we're putting into this, I'm not convinced that this shouldn't be happening in `stripity-stripe` or a related testing library, lest we reproduce the whole surface area of that library in our mocked API.

lib/code_corps/stripe/adapters/stripe_customer.ex
```diff
@@ -1,7 +1,11 @@
 defmodule CodeCorps.Stripe.Adapters.StripeCustomer do
-  import CodeCorps.MapUtils, only: [rename: 3]
+  import CodeCorps.MapUtils, only: [rename: 3, keys_to_string: 1]
 
-  def params_from_stripe(%{} = stripe_map) do
-    stripe_map |> rename("id", "id_from_stripe")
+  def params_from_stripe(%Stripe.Customer{} = customer) do
```
joshsmith: I might call this `to_params` instead.

test/views/stripe_customer_view_test.exs
```diff
@@ -0,0 +1,32 @@
...
+    user = insert(:user)
+    stripe_customer = insert(:stripe_customer, id_from_stripe: "some_id", email: "email", user: user)
+
+    rendered_json =  render(CodeCorps.StripeCustomerView, "show.json-api", data: stripe_customer)
+
+    expected_json = %{
+      "data" => %{
+        "attributes" => %{
+          "email" => stripe_customer.email,
+        },
```
joshsmith, line 22: This appears to be missing attributes (and relationships) that are in the API Blueprint but not here.

test/views/user_view_test.exs
```diff
@@ -38,7 +40,10 @@ defmodule CodeCorps.UserViewTest do
             ]
           },
           "slugged-route" => %{
-            "data" => nil
+            "data" => %{"id" => slugged_route.id |> Integer.to_string, "type" => "slugged-route"}
+          },
+          "stripe-customer" => %{
+            "data" => %{"id" => stripe_customer.id |> Integer.to_string, "type" => "stripe-customer"}
```
joshsmith, line 46: Do we want to be rendering this relationship out regardless of authorization?

web/controllers/stripe_customer_controller.ex
```diff
@@ -0,0 +1,35 @@
...
+  def handle_create(conn, attributes) do
+    attributes
+    |> CodeCorps.StripeService.create_customer
+    |> handle_stripe_response(attributes, conn)
+  end
+
+  defp handle_stripe_response({:ok, stripe_response}, attributes, conn) do
+    stripe_response
+    |> CodeCorps.Stripe.Adapters.StripeCustomer.params_from_stripe
+    |> Map.merge(attributes)
```
joshsmith: How does this merging work? Are there attributes that are not sent to Stripe or transformed somehow that need merged?

web/policies/stripe_customer_policy.ex
```diff
@@ -0,0 +1,12 @@
+defmodule CodeCorps.StripeCustomerPolicy do
+  alias CodeCorps.StripeCustomer
+  alias CodeCorps.User
+  alias Ecto.Changeset
+
+  def create?(%User{id: current_user_id}, %Changeset{changes: %{user_id: user_id}}), do: current_user_id == user_id
+  def create?(%User{}, %Changeset{}), do: false
```
joshsmith, line 7: Been wondering if we should just use `(_, _)` to handle the `false` case.

web/router.ex
```diff
@@ -84,10 +81,19 @@ defmodule CodeCorps.Router do
...
     resources "/user-skills", UserSkillController, only: [:create, :delete]
   end
+
+  scope "/", CodeCorps, host: "api." do
+    pipe_through [:api, :bearer_auth, :current_user]
+
+    get "/:slug", SluggedRouteController, :show
+    get "/:slug/projects", ProjectController, :index
+    get "/:slug/:project_slug", ProjectController, :show
+  end
```
joshsmith, line 93: Why don't we reorder authenticated routes above this so we're not separating the same pipeline?

web/views/stripe_customer_view.ex
```diff
@@ -0,0 +1,9 @@
+defmodule CodeCorps.StripeCustomerView do
+  use CodeCorps.PreloadHelpers, default_preloads: [:user]
+  use CodeCorps.Web, :view
+  use JaSerializer.PhoenixView
+
+  attributes [:email]
```
joshsmith: Mentioned above but missing attributes defined in API docs. `email` and `id_from_stripe` are hidden when the user is not the current user.

lib/code_corps/stripe/in_memory.ex
begedin: I agree that it's not great, but I couldn't think of a better way. We would not be recreating the whole of `stripity-stripe`, but we would basically be recreating the parts that we use.

I'm open to suggestions, but until the library itself offers some testing tools, I don't know what else to do. Maybe https://github.com/parroty/exvcr ?

web/controllers/stripe_customer_controller.ex
begedin: The client also sends the `user_id`

test/views/user_view_test.exs
begedin: It's just our own db id, not an actual stripe id, so it should not be important. I wouldn't want to confuse ember data with hiding potential relationships. Our actual stripe_customer view can hide the `id_from_stripe` based on authorization, if necessary.

web/controllers/stripe_customer_controller.ex
begedin: Made this more explicit by moving it into an adapter function called `add_non_stripe_attributes`

web/policies/stripe_customer_policy.ex
begedin: Considering how our `ability.ex` is written, it can never be anything other than a `%User{}` and a `%Changeset{}`.

web/router.ex
begedin: Didn't consider it, but fairly obvious in hindsight. Done!

web/views/stripe_customer_view.ex
begedin: I've added all the attributes listed in the API docs.

I dod not add the subscriptions, because there is at the moment no relationship between the stripe_customer and the stripe_subscription. There is a relationship between `user` and `stripe_subscription` in that a `stripe_subscription` belongs to a `user` and a `stripe_plan`. A `stripe_customer` also belongs to a `user`. 

Should I update the API docs, or should I update our relationships? I'm pretty sure we mentioned this and decided to go with the relationship being on the `user`, not the `stripe_customer`.

Also, regarding the attributes, should we really render all those attributes? What do we get from it, really? Unless I'm missing something, we don't actually use the data client-side. We could even potentially go as far as completely hiding the `stripe_customer` from the client and just adding some sort of `connected_to_stripe` flag on the user record. I'm really not sure what the best approach is here, really.

lib/code_corps/stripe/adapters/stripe_customer.ex
```diff
@@ -1,11 +1,29 @@
...
   import CodeCorps.MapUtils, only: [rename: 3, keys_to_string: 1]
 
-  def params_from_stripe(%Stripe.Customer{} = customer) do
+  def to_params(%Stripe.Customer{} = customer) do
     customer
     |> Map.from_struct
     |> Map.take([:created, :currency, :delinquent, :id, :email])
     |> rename(:id, :id_from_stripe)
     |> keys_to_string
   end
```
joshsmith, line 10: What about making `to_params` take `customer` and `attributes` so we don't have to do two separate adapter operations, then this function will combine the `add_non_stripe_attributes` in?

web/views/stripe_customer_view.ex
```diff
@@ -3,7 +3,29 @@ defmodule CodeCorps.StripeCustomerView do
...
+  @doc """
+  Returns the id_from_stripe or an empty string, depending on the stripe_customer record
+  being rendered is the authenticated user's record, or some other user's.
+
+  Users can only see their own stripe ids. Everyone else's are private.
+  """
+  def id_from_stripe(stripe_customer, %Plug.Conn{assigns: %{current_user: current_user}}) do
+    if stripe_customer.user_id == current_user.id, do: stripe_customer.id_from_stripe, else: ""
+  end
+  def id_from_stripe(_stripe_customer, _conn), do: ""
```
joshsmith, line 30: I don't think this is what I meant. I meant hiding the relationship from the user view if they're not authed.

test/views/user_view_test.exs
joshsmith: I can see the logic in that, but do we actually risk confusing `ember-data` by doing so? If we're manually refreshing the user on authentication (we are), then it shouldn't matter.

Even the existence of those relationships leaking to other people seems problematic to me. Will think on this and ask around some.

joshsmith (approved):

Issue #446, opened 2016-11-07
Add controllers for Stripe data


joshsmith:
# Problem

We've documented a number of endpoints in our API blueprint that need to be turned into actual API endpoints with tests.

## Subtasks
- [ ] Create tasks for all the individual endpoints that need created
- [ ] Write the tests first
- [ ] Make the tests pass with working code

## References

Progress on: #10

joshsmith:
Closing in favor of actual controller issues.

PR #456, opened 2016-11-10
Write policy and controller actions for StripePlatformCard

### LINKED ISSUE #454: Write Policy for StripePlatformCard

### LINKED ISSUE #470: Change StripeCard to StripePlatformCard
# Problem

We need to change the `StripeCard` to a `StripePlatformCard` that has many `StripeConnectCard`s.

mackenziehicks:
# What's in this PR?

Added Stripe Card policy to stripe_card_policy.ex, factory, and empty tests to stripe_account_card_test.

## References
Fixes #454 
Closes #470 

Progress on: #447 
Progress on #446

This can serve as an example for our other issues in #446 and #447

## Need to implement the following behavior

* [x] controller receives action, params are `stripe_token` and `user_id`
* [x] controller checks authorization, which is based on user_id and current user (changeset based)
* [ ] stripe service is called to create the actual record on stripe
* [ ] the service request returns the record data
* [ ] we store the record data in our database


joshsmith (commented):
👍  all looks good.

begedin:
@JoshSmith One of us is gonna have to take over here, since this is a whole new thing we're doing here (`index?` and `show?` policies).

# Specs

For this specific case, the only thing that makes sense is
- both endpoints require authentication
- index endpoint sends all records for current user if no filter
- index endpoint allows coalesced ids. We need to fetch records and see if the user has access to each individual record. If any of them are not accessible by user, return a 403, otherwise, return a 200. 
- show endpoint looks fine, we just need tests for it

# Implementation
- for now, we should implement the index endpoint using the `ja_resource` records callback, calling authorize manually for each record in the coalesced id case

I talked to Alan about it, but he mostly reiterated our thoughts on it. There is no good way to do this. My proposal is as following.

# Notes

`stripe-card` is the simpler one. `stripe-subscriptions` are going to be more difficult.

I talked to Alan on api design, but he mainly reiterated my thoughts on it. There is no "correct" way to do this. We need to consider what our endpoint is fetching and decide how it should behave based on that.

joshsmith:
@begedin is this ready to review now?

begedin:
As discussed, I removed `index` and wrote polices for index, show, delete.

Also wrote controller specs and endpoints. Only thing not working is the create endpoint. Success and validation tests are there and both fail. It's a pretty big thing that isn't working, but it's a step in that direction.

We might need to modify the behavior of the create policy so it doesn't accept a changeset and instead just goes with a hash, which should be relatively simple.

Also, there is a problem with our api case helpers, or more precicely, our json_payload strategy. It relies on the views we defined, so if the view doesn't render some attribute, the payload strategy won't render it either, even though it should, in the case of `create`. 

We might need to build the payload for the create helper manually. I have an idea how to fix the payload strategy, but I'll have to do it at a later time.

joshsmith:
@begedin I have now renamed `StripeCard` to `StripePlatformCard` and added a `StripeConnectCard`.

Feel free to apply the last three items on the checklist in the description to `StripePlatformCard`, and then we can review/merge in favor of doing #476 next.

test/controllers/stripe_platform_card_controller_test.exs
```diff
@@ -30,26 +30,42 @@ defmodule CodeCorps.StripePlatformCardControllerTest do
...
+      insert(:stripe_platform_customer, user: current_user)
       valid_attrs = %{stripe_token: "tok_test123456", user: current_user}
-      assert conn |> request_create(valid_attrs) |> json_response(201)
-    end
-
-    @tag :authenticated
-    test "renders 422 when data is invalid", %{conn: conn, current_user: current_user} do
-      invalid_attrs = %{stripe_token: nil, user: current_user}
-      assert conn |> request_create(invalid_attrs) |> json_response(422)
+      assert conn |> make_create_request(valid_attrs) |> json_response(201)
```
joshsmith, line 59: I'm thinking we should do some checking here like we do for the Analytics identify. Was the Stripe API hit? Not sure where the right boundary for tests happens to be but not sure we're hitting it right now.

joshsmith (approved):

PR #458, opened 2016-11-11
Added a note to installation steps for windows users

rileytaylor:
# What's in this PR?

Simple fix, this adds a note to INSTALLING.md for windows users to work with docker through powershell.

### DIFF
--- a/docs/INSTALLING.md
+++ b/docs/INSTALLING.md
@@ -8,6 +8,8 @@ Here are some direct links if you're on [Mac OS X](https://docs.docker.com/docke
 
 Follow those download instructions. Once you can run the `docker` command, you can safely move on.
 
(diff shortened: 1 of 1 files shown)


joshsmith:
I would make this a recommend and not a must, unless Docker's docs say otherwise.

joshsmith:
@rileytaylor ping.

rileytaylor:
@JoshSmith from what I can find in documentation there is nothing saying you have to, though it is generally what Docker uses for samples. I do know it was at a minimum highly recommended (that's where I'm recalling the requirement) at BUILD. I'm going to make it recommended and put a note that if there are odd issues to run PS as Admin.

joshsmith (approved):

joshsmith:
LGTM!

PR #460, opened 2016-11-11
Add StripeConnectAccount controller and tests

### LINKED ISSUE #459: Add controller for StripeConnectAccount
# Problem

Adding  Stripe Account controller and tests.

Account actions needed: `create`, `show`

## References

Progress on: #446

mackenziehicks:
# What's in this PR?

Adding Stripe Account controller and tests.  Added factory, router path, and controller and tests for index and show.

- [x] Rename StripeAccount to StripeConnectAccount everywhere
- [x] Create the account in the Stripe API
- [x] Store the data in our database

## References
Fixes #459 

Progress on: #446


mackenziehicks:
@JoshSmith the test for index is giving a 404 response.  We aren't sure what we are missing, can you help?

joshsmith:
We need to reorder the endpoints in the router. Authenticated ones need to come first, otherwise slugged routes glob up everything for `index` and`show` both, so it looks like a `404` when it's not.

begedin:

mackenziehicks:
@begedin I removed the index endpoints.  I wasn't sure how to write more tests before the policy and ability changes got merged (#451)

joshsmith:
This needs updated now to reflect the changes in `stripity-stripe`. That's why the test is now failing. 

https://hexdocs.pm/stripity_stripe/2.0.0-alpha.2/Stripe.Connect.OAuth.html#authorize_url/1

joshsmith:
We might want to rename this to `StripeConnectAccount`, in keeping with the `StripeConnect` renaming going on throughout.

joshsmith:
@begedin do you mind reworking this to make the tests pass, fixing conflicts? I can rename the model if you like, since I'm kind of an expert in it now. But I want tests passing first.

begedin:
@JoshSmith  The checkboxes are filled, but this is not done yet. 

There is a virtual field called `account` that needs renaming to `access_code`, both here and on ember. We may want to consider using `adapterOptions` instead of a fake attribute on ember.

Also, the process on the API is 
* receive access code
* use access code to get token credentials
* use account id from token credentials to get account
* store account into db

I'm not sure if we should store credentials somewhere. I'm thinking possibly, or even probably. The credentials involve a publishable key, an access token and a refresh token.

joshsmith:
We should not store account credentials. We'll use the account id to connect in the opts list of the library calls that sets the Stripe account header.

joshsmith:
We still need to:

~- Fix Connect.OAuth faked API~
~- Create controller test needs fixed~

joshsmith (approved):

PR #487, opened 2016-11-24
Add StripeSubscription controller

joshsmith:
# What's in this PR?

WIP.


begedin:
# NOTES 

__updating as I go along__

* Had trouble with creating a card token. Had to reconnect to stripe using a brand new stripe account to get it working. Might be a problem with using the same stripe account to get an access code (using stripe connect button) which also owns the API keys. Wasted some time on that :(
* Wrote a couple of test methods to quickly destroy/create data to work with. They/re in `lib/stripe/stripe_connect_subscription.ex`, at the bottom.
* Since we only have one card (and will have one "default" card in the future), there is no need to specify it on create. It can be inferred from the user. This decouples things a bit.

* Tried working on webhooks, but was unable to get to a point where I can receive them. In big part, it's setup issues (couldn't completely figure it out), but another major part is that, using stripity stripe to create a subscription apparently works, but with two major issues:
  * The hash I receive back is a `StripeSubscription` with ID set to, for example, `sub_12345`. It contains, embedded, a stripe plan with the exact same ID set. This is, I guess, a stripity_stripe bug, but I was unable to confirm yet.
  * Attempting to retrieve the subscription with the specified ID, connected account or not, returns a "no such subscription" response.
* As for my setup, I tried using the ultrahooks gem, to send them to localhost, combined with `stripe_eventex` (for which the setup is still missing a route in the router), but due to the above issue, I was unable to trigger an event to attempt to catch it.

* EDIT: Got subscription to update, attempting to catch webhooks now. Also found a "send test webhook" feature on dashboard

PR #491, opened 2016-11-25
479 set current donation goal

### LINKED ISSUE #479: Set current donation goal appropriately
# Problem

We need to set the proper donation goal to `current` when:

- the donation goals are updated
- the cached total donated column changes on the project

We'll need to find the donation goal that is most closely above the current amount and set that to the current goal. That is, the goal can shift depending on the other goals and the current monthly raised.

The other goals will need to be set to `current: false` in that case.

begedin:
# What's in this PR?

Adds a donation goals manager module, for inserting and updating donation goals into the database, while also updating the current goal.

Also exposes an `update_current_goal/1` which can be used to perform the update, without actually inserting/updating a specific record. This can be used when a subscription is altered in some way.

`create/1` and `update/2` will return one of 3 outcomes:
* `{:ok, record_inserted_or_updated}`
* `{:error, changeset_with_validation_errors}`
* `{:error, :unhandled}` - should not happen, triggers if something happens that we are not yet handling

`update_current_goal/1` returns
* `{:ok, current_goal_record}` - if a goal has been set to current
* `{:ok, nil}` - if no goal has been set to current - this can only happen if either there are no goals defined yet, or there is no plan for the project

Determining the current amount donated is currently done by aggregating through subscriptions and adding them up. Once we implement that part, this method chain can be replaced with something simpler. There is a `# TODO` comment in place to remind us of that.

## References
Fixes #479

### DIFF
--- a/lib/code_corps/donation_goals_manager.ex
+++ b/lib/code_corps/donation_goals_manager.ex
@@ -1,0 +1,115 @@
+defmodule CodeCorps.DonationGoalsManager do
+  @moduledoc """
+  Handles CRUD operations for donation goals.
+
+  When operations happen on `CodeCorps.DonationGoal`, we need to set
+  the current donation goal on `CodeCorps.Project`.
+
+  The current donation goal should be the smallest value that is
+  greater than the project's current total donations, _or_ falls back to
+  the largest donation goal.
+  """
+
+  import Ecto.Query
+
+  alias CodeCorps.{DonationGoal, Project, Repo}
+  alias Ecto.Multi
+
+  def create(attributes) do
+    changeset = %DonationGoal{} |> DonationGoal.create_changeset(attributes)
+
+    multi = Multi.new
+    |> Multi.insert(:donation_goal, changeset)
+    |> Multi.run(:update_related_goals, &update_related_goals/1)
+
+    case Repo.transaction(multi) do
+      {:ok, %{donation_goal: donation_goal, update_related_goals: _}} ->
+        {:ok, donation_goal}
+      {:error, :donation_goal, %Ecto.Changeset{} = changeset, %{}} ->
+        {:error, changeset}
+      {:error, _failed_operation, _failed_value, _changes_so_far} ->
+        {:error, :unhandled}
+    end
+  end
+
+  def update(%DonationGoal{} = donation_goal, attributes) do
+    changeset = donation_goal |> DonationGoal.create_changeset(attributes)
+
+    multi = Multi.new
+    |> Multi.update(:donation_goal, changeset)
+    |> Multi.run(:update_related_goals, &update_related_goals/1)
+
+    case Repo.transaction(multi) do
+      {:ok, %{donation_goal: donation_goal, update_related_goals: _}} ->
+        {:ok, Repo.get(DonationGoal, donation_goal.id)}
+      {:error, :donation_goal, %Ecto.Changeset{} = changeset, %{}} ->
+        {:error, changeset}
+      {:error, _failed_operation, _failed_value, _changes_so_far} ->
+        {:error, :unhandled}
+    end
+  end
+
+  def set_current_goal_for_project(%Project{} = project) do
+    project
+    |> find_current_goal
+    |> set_to_current(project)
+  end
+
+  defp update_related_goals(%{donation_goal: %DonationGoal{project_id: project_id}}) do
+    Project
+    |> Repo.get(project_id)
+    |> set_current_goal_for_project
+  end
+
+  defp find_current_goal(%Project{} = project) do
+    amount_donated = get_amount_donated(project)
+    case find_lowest_not_yet_reached(project, amount_donated) do
+      nil -> find_largest_goal(project)
+      %DonationGoal{} = donation_goal -> donation_goal
+    end
+  end
+
+  defp get_amount_donated(%Project{id: project_id}) do
+    # TODO: This should be simplified by having
+    # subscriptions relate to projects instead of plans
+    # and by caching the total amount on the project itself
+
+    CodeCorps.StripeConnectPlan
+    |> Repo.get_by(project_id: project_id)
+    |> aggregate_donations
+    |> default_to_zero
+  end
+
+  defp aggregate_donations(nil), do: 0
+  defp aggregate_donations(%CodeCorps.StripeConnectPlan{id: plan_id}) do
+    CodeCorps.StripeConnectSubscription
+    |> where([s], s.stripe_connect_plan_id == ^plan_id)
+    |> Repo.aggregate(:sum, :quantity)
+  end
+
+  defp default_to_zero(nil), do: 0
+  defp default_to_zero(value), do: value
+
+  defp find_lowest_not_yet_reached(%Project{id: project_id}, amount_donated) do
+    DonationGoal
+    |> where([d], d.project_id == ^project_id and d.amount > ^amount_donated)
+    |> order_by(asc: :amount)
+    |> limit(1)
+    |> Repo.one
+  end
+
+  defp find_largest_goal(%Project{id: project_id}) do
+    DonationGoal
+    |> where([d], d.project_id == ^project_id)
+    |> order_by(desc: :amount)
+    |> limit(1)
+    |> Repo.one
+  end
+
+  defp set_to_current(%DonationGoal{} = donation_goal, %Project{} = project) do
+    attrs = %{current_donation_goal_id: donation_goal.id}
+    project
+    |> Project.set_current_donation_goal_changeset(attrs)
+    |> Repo.update
+  end
+end
--- a/test/controllers/donation_goal_controller_test.exs
+++ b/test/controllers/donation_goal_controller_test.exs
@@ -1,7 +1,7 @@
 defmodule CodeCorps.DonationGoalControllerTest do
   use CodeCorps.ApiCase, resource_name: :donation_goal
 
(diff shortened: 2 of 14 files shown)


web/models/project.ex
```diff
@@ -22,9 +22,11 @@ defmodule CodeCorps.Project do
     belongs_to :organization, CodeCorps.Organization
     has_one :stripe_connect_plan, CodeCorps.StripeConnectPlan
 
+
     has_many :donation_goals, CodeCorps.DonationGoal
     has_many :project_categories, CodeCorps.ProjectCategory
     has_many :project_skills, CodeCorps.ProjectSkill
+    has_many :stripe_connect_subscriptions, through: [:stripe_connect_plan, :stripe_connect_subscriptions]
```
begedin: This can eventually not be a `through` if we make it so subscriptions belong to project directly, instead of plan

joshsmith: Despite our conversation, it still may make sense to add subscriptions to projects. Not completely necessary, but could be fine.

test/controllers/donation_goal_controller_test.exs
```diff
@@ -1,7 +1,7 @@
 defmodule CodeCorps.DonationGoalControllerTest do
   use CodeCorps.ApiCase, resource_name: :donation_goal
 
-  @valid_attrs %{amount: 200, current: false, description: "A description"}
+  @valid_attrs %{amount: 200, description: "A description"}
   @invalid_attrs %{description: nil}
```
joshsmith, line 5: Should we add `current` to the invalid attrs?

test/lib/code_corps/donation_goals_manager_test.exs
```diff
@@ -0,0 +1,171 @@
...
+
+      # total donated is 0,
+      # only goal inserted is the first goal
+      assert_current_goal_id(first_goal.id)
+
+      {:ok, second_goal} = DonationGoalsManager.create(%{amount: 20, description: "Test", project_id: project.id})
+
+      # total donated is still 0
+      # first goal larger than 0 is the very first goal
+      assert_current_goal_id(first_goal.id)
```
joshsmith, line 50: I think maybe reversing amounts on these would be good, so `20` then `10`, so we know `10` becomes the goal.

test/lib/code_corps/donation_goals_manager_test.exs
```diff
@@ -0,0 +1,171 @@
...
+  alias CodeCorps.DonationGoalsManager
+
+  defp assert_current_goal_id(goal_id) do
+    current_goal = DonationGoal |> Repo.get_by(current: true)
+    assert current_goal.id == goal_id
+  end
+
+  defp donate(plan, amount) do
+    insert(:stripe_connect_subscription, quantity: amount, stripe_connect_plan: plan)
+  end
```
joshsmith, line 20: This is nice!

web/models/project.ex
```diff
@@ -22,9 +22,11 @@ defmodule CodeCorps.Project do
     belongs_to :organization, CodeCorps.Organization
     has_one :stripe_connect_plan, CodeCorps.StripeConnectPlan
 
+
```
joshsmith, line 26: Random extra line here.

joshsmith (commented):
This is really nice. I'll have to do a more in-depth look later, but like the direction you took with it. Me attempting this at 12:30AM was not good. 😂

lib/code_corps/donation_goals_manager.ex
```diff
@@ -0,0 +1,114 @@
...
+  end
+
+  def update_current_goal(%Project{} = project) do
+    unset_all_goals_from_current(project)
+    set_new_current_goal(project)
+  end
+
+  defp transactional_update_current_goal(%{donation_goal: %DonationGoal{project_id: project_id}}) do
+    Project
+    |> Repo.get(project_id)
```
venkatd, line 60: I think this would be more readable as `Repo.get(Project, project_id)`

lib/code_corps/donation_goals_manager.ex
```diff
@@ -0,0 +1,114 @@
...
+        {:error, :unhandled}
+    end
+  end
+
+  def update(donation_goal, attributes) do
+    changeset = donation_goal |> DonationGoal.create_changeset(attributes)
+
+    multi = Multi.new
+    |> Multi.update(:donation_goal, changeset)
+    |> Multi.run(:update_current_goal, &transactional_update_current_goal/1)
```
venkatd: I've never used `Multi.run` so can't comment on this technique.

lib/code_corps/donation_goals_manager.ex
```diff
@@ -0,0 +1,114 @@
...
+    |> fallback_to_largest_reached(project, amount_donated)
+  end
+
+  defp get_amount_donated(%Project{id: project_id}) do
+    # TODO: This should be simplified by having
+    # subscriptions relate to projects instead of plans
+    # and by caching the total amount on the project itself
+
+    CodeCorps.StripeConnectPlan
+    |> Repo.get_by(project_id: project_id)
```
venkatd, line 78: Same thing here. I find it difficult to read when you are piping the model into here.

test/controllers/donation_goal_controller_test.exs
begedin: Nah. `current` is not set by the user, period. It should not be considered invalid, it should just be ignored.

lib/code_corps/donation_goals_manager.ex
begedin: I think it warrants a discussion, but we've been using this approach in a lot of other places in the project, so I wouldn't want to change styles for this module alone.

joshsmith: I think prefer this approach.

lib/code_corps/donation_goals_manager.ex
```diff
@@ -0,0 +1,114 @@
...
+  def create(attributes) do
+    changeset = %DonationGoal{} |> DonationGoal.create_changeset(attributes)
+
+    multi = Multi.new
+    |> Multi.insert(:donation_goal, changeset)
+    |> Multi.run(:update_current_goal, &transactional_update_current_goal/1)
+
+    case Repo.transaction(multi) do
+      {:ok, %{donation_goal: donation_goal, update_current_goal: _}} ->
+        {:ok, Repo.get(DonationGoal, donation_goal.id)}
```
joshsmith: Not sure if we need to `Repo.get` here since the goal is returned from the transaction.

lib/code_corps/donation_goals_manager.ex
```diff
@@ -0,0 +1,114 @@
...
+    |> aggregate_donations
+    |> default_to_zero
+  end
+
+  defp aggregate_donations(nil), do: 0
+  defp aggregate_donations(%CodeCorps.StripeConnectPlan{id: plan_id}) do
+    CodeCorps.StripeConnectSubscription
+    |> where([s], s.stripe_connect_plan_id == ^plan_id)
+    |> Repo.aggregate(:sum, :quantity)
+  end
```
joshsmith, line 88: We'll replace with `total_donated` from `Project` as soon as #492 is merged.

lib/code_corps/donation_goals_manager.ex
DavidAntaramian: Was asking Josh about this since I wasn't sure if I missed something. `Multi.insert` should be returning a `%DonationGoal{}` under the `:donation_goal` key.

lib/code_corps/donation_goals_manager.ex
DavidAntaramian: @venkatd It's an appropriate solution

lib/code_corps/donation_goals_manager.ex
```diff
@@ -0,0 +1,114 @@
...
+        {:error, :unhandled}
+    end
+  end
+
+  def update_current_goal(%Project{} = project) do
+    unset_all_goals_from_current(project)
+    set_new_current_goal(project)
+  end
+
+  defp transactional_update_current_goal(%{donation_goal: %DonationGoal{project_id: project_id}}) do
```
DavidAntaramian: 1. The name of this is misleading. It implies that it will wrap its behavior in a transaction, which it doesn't. Instead, it expects to be used _inside_ a transaction.
2. I don't believe that you need a `Project` record at this point, therefore the function itself seems like its doing extra work.

DavidAntaramian:
I discussed this with @JoshSmith. In my opinion, the benefits you get from a circular reference system may outweigh its evils. In essence, you currently have a system which makes it easy to get into an inconsistent state (i.e., that multiple `DonationGoal` records for the same project can be set to `current: true`). If you were to have the `Project` record identify the `current_donation_goal_id`, it would alleviate that possibility of inconsistency.

It does, of course, add a circular reference into the schematic design. Since `projects.current_donation_goal_id` should be nullable, it is possible to set the relationship to `ON DELETE SET NULL` so that when a project deletion is triggered, it will trigger deletion of all the donation goals, which will first nullify appropriate `current_donation_goal_id` fields.

The other option to achieve this without a circular reference requires creating a join table specifically for current donation goals. Because the `projects_current_donation_goals` table is dependent on `projects` and `donation_goals`, and `donation_goals` is dependent upon `projects`, but `projects` is not dependent on either, the dependency graph is acyclic. As long as the table is set to unique for the `project_id`, every project can only ever have a single current donation goal.

joshsmith (approved):

PR #497, opened 2016-11-28
Update Stripe webhooks

### LINKED ISSUE #370: Add webhook handling for Stripe events
- [ ] We can do a plug maybe for deseralizing events of any type
- [ ] Controllers will handle creating/updating the right things based on the events we're getting (cards, etc)
- [ ] There needs to be two webhooks on Stripe, an account webhook and a connect webhook

I think we want to handle the following events:
- [ ] `account.updated` describes an account (on the connect webhook)

> Occurs whenever an account status or property has changed.

We will want to update the `StripeConnectAccount` with the new information.

- [ ] `customer.subscription.updated` describes a subscription (on the connect webhook)

> Occurs whenever a subscription changes. Examples would include switching from one plan to another, or switching status from trial to active.

We will want to update the `StripeConnectSubscription` with the new information, as well as update the total monthly amount that the project is currently raising.

joshsmith:
# What's in this PR?

This implements most of what's left in the Stripe webhooks.

I could not get the failing test to work for the "customer.subscription.updated" event so I removed it. Apparently `nil` chaining is being added to `Repo.preload` that would fix this issue, so going to hold off.

## References
Fixes #370 when finished

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -46,7 +46,10 @@ config :guardian, Guardian,
   secret_key: "e62fb6e2746f6b1bf8b5b735ba816c2eae1d5d76e64f18f3fc647e308b0c159e"
 
 config :code_corps, :analytics, CodeCorps.Analytics.InMemoryAPI
+
+# Configures stripe for dev mode
 config :code_corps, :stripe, Stripe
+config :code_corps, :stripe_env, :dev
 
 config :sentry,
   environment_name: Mix.env || :dev
--- a/lib/code_corps/stripe_service/events/account_updated.ex
+++ b/lib/code_corps/stripe_service/events/account_updated.ex
@@ -1,148 +1,25 @@
 defmodule CodeCorps.StripeService.Events.AccountUpdated do
-  alias CodeCorps.StripeService.Adapters
+  alias CodeCorps.StripeService.Adapters.StripeConnectAccountAdapter
   alias CodeCorps.StripeConnectAccount
   alias CodeCorps.Repo
 
   @api Application.get_env(:code_corps, :stripe)
 
-  # def handle(%{"data" => %{"object" => %{"livemode" => false}}}), do: {:ok, :ignored_not_live}
-  def handle(%{"data" => data}) do
-    stripe_account = data |> retrieve_account
-    local_account = data |> load_account
-
-    local_account |> update(stripe_account)
-  end
-
-  defp retrieve_account(%{"object" => %{"id" => id_from_stripe}}) do
-    # hardcoded for testing
-    id_from_stripe = "acct_19JlxsFTVVt7Lv80"
-
-    {:ok, stripe_account} = @api.Account.retrieve(id_from_stripe)
-    stripe_account
-  end
-
-  defp load_account(%{"object" => %{"id" => id_from_stripe}}) do
-     # hardcoded for testing
-    id_from_stripe = "acct_19JlxsFTVVt7Lv80"
-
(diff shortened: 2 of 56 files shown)


joshsmith:
The subscription updated event code is more complex than I would like and there is some code being duplicated in other places (like aggregating and updating project monthly totals, which could be moved out into a separate module specific to managing that state).

joshsmith:
It's also becoming clear that as our mocks are really tightly coupled to the code itself. We need to figure this out soon-ish.

joshsmith:
I have not yet tested this against Stripe to see if the webhooks are working. Will use http://www.ultrahook.com/ when I get a chance but need a break from this.

lib/code_corps/stripe_service/events/customer_subscription_updated.ex
```diff
@@ -58,80 +64,19 @@ defmodule CodeCorps.StripeService.Events.CustomerSubscriptionUpdated do
...
+  defp default_to_zero(value), do: value
+
   defp update_project_totals(%Project{id: project_id} = project) do
     total_monthly_donated =
       StripeConnectSubscription
-      |> where([s], s.status=="active")
+      |> where([s], s.status == "active")
       |> Repo.aggregate(:sum, :quantity)
 
+    total_monthly_donated = default_to_zero(total_monthly_donated)
```
begedin: Could just add to the above chain.

config/dev.exs
```diff
@@ -46,7 +46,10 @@ config :guardian, Guardian,
   secret_key: "e62fb6e2746f6b1bf8b5b735ba816c2eae1d5d76e64f18f3fc647e308b0c159e"
 
 config :code_corps, :analytics, CodeCorps.Analytics.InMemoryAPI
+
+# Configures stripe for dev mode
 config :code_corps, :stripe, Stripe
+config :code_corps, :stripe_env, :dev
```
begedin, line 52: `Mix.env` is not available in production, so if we want to differentiate between livemode events on different environments, we need to set an explicit config value. 

Right now, in the code below, the events are setup so that production will ignore `livemode: false` events and other environments will ignore `livemode: true` events.

lib/code_corps/donation_goals_manager.ex
```diff
@@ -71,7 +71,6 @@ defmodule CodeCorps.DonationGoalsManager do
   Updates sibling goals for a `CodeCorps.DonationGoal` by:
 
   - Finding this goal's project
-  - Calculating the total donations for the project
```
begedin, line 74: We don't calculate the totals here anymore.

lib/code_corps/stripe_service/events/account_updated.ex
```diff
@@ -5,144 +5,21 @@ defmodule CodeCorps.StripeService.Events.AccountUpdated do
...
+         %StripeConnectAccount{} = local_account <-
+           Repo.get_by(StripeConnectAccount, id_from_stripe: id_from_stripe),
+         {:ok, params} <-
+           stripe_account |> Adapters.StripeConnectAccount.to_params(%{})
+    do
+      local_account
+      |> StripeConnectAccount.webhook_update_changeset(params)
+      |> Repo.update
+    else
+      {:error, %Stripe.APIErrorResponse{}} -> {:error, :stripe_error}
```
begedin, line 20: It could conceivable fail with this as well, so just a `nil` case is not enough. I also added a catch-all case, just in case (no pun intended).

PR #501, opened 2016-11-28
Move user and customer updates into a transaction, managed by a module

### LINKED ISSUE #489: Pass parameters to Stripe customer
# Problem

Right now we're not passing _any_ parameters to Stripe to create the Stripe customer.

We should, at the very least, pass an email and ensure it's kept up-to-date so we can search Stripe for the user's email.

begedin:
# What's in this PR?

Adds a `UserManager` module, which, for now, is placed under lib/code_corps/user_manager.ex, but ought to probably moved into a new namespace, together with `DonationGoals` manager.

This module, for now, exposes an `update` function, which receives a user and a set of attributes.

The `update` function behaves in the following way:

* If the update does not change the user's email, then it only updates the user
* If the update does change the user email, and there is a platform customer associated with the user, it updates the platform customer as well, both on stripe and locally.

To update the platform customer, I've also implemented `StripeService.StripePlatformCustomer.update/2`, which handles updating of both the local and the API record.

The scope of the PR was unclear to me, so I implemented the whole process. Based on the issue title alone, the assumption would be we just need to pass in the parameters during customer creation. However, the issue description also mentions keeping it up to date so "we can search by user email".

## Questions:

* Right now, it seems like the email is the only field that makes sense to update. A platform customer also has a description field, so I guess we could be setting something there as well.
* Initially, I was going to perform connect customer updates as well, but we aren't storing anything locally that could conceivably be updated. The email might need to be kept up to date on stripe, but we are not storing it locally.

## References
Fixes #489

### DIFF
--- a/lib/code_corps/services/user_service.ex
+++ b/lib/code_corps/services/user_service.ex
@@ -1,0 +1,100 @@
+defmodule CodeCorps.Services.UserService do
+  @moduledoc """
+  Handles CRUD operations for users.
+
+  When operations happen on `CodeCorps.User`, we need to make sure changes
+  are propagated to related records, ex., `CodeCorps.StripePlatformCustomer` and
+  `CodeCorps.StripeConnectCustomer`
+
+  """
+
+  alias CodeCorps.{Repo, StripeConnectCustomer, StripePlatformCustomer, User}
+  alias CodeCorps.StripeService.{StripeConnectCustomerService, StripePlatformCustomerService}
+  alias Ecto.Changeset
+  alias Ecto.Multi
+
+  @doc """
+  Updates a `CodeCorps.User` record and, if necessary, associated
+  `CodeCorps.StripePlatformCustomer` and `CodeCorps.StripeConnectCustomer` records.
+
+  These related records inherit the email field from the user,
+  so they need to be kept in sync, both locally, and on the Stripe platform.
+
+  Returns one of
+  * `{:ok, %CodeCorps.User{}, nil, nil}`
+  * `{:ok, %CodeCorps.User{}, %CodeCorps.StripePlatformCustomer{}, nil}`
+  * `{:ok, %CodeCorps.User{}, %CodeCorps.StripePlatformCustomer{}, %CodeCorps.StripeConnectCustomer{}}`
+  * `{:error, %Ecto.Changeset{}}`
+  * `{:error, :unhandled}`
+
+  """
+  def update(%User{} = user, attributes) do
+    changeset = user |> User.update_changeset(attributes)
+    do_update(changeset)
+  end
+
+  defp do_update(%Changeset{changes: %{email: _email}} = changeset) do
+    multi =
+      Multi.new
+      |> Multi.update(:update_user, changeset)
+      |> Multi.run(:update_platform_customer, &update_platform_customer/1)
+      |> Multi.run(:update_connect_customers, &update_connect_customers/1)
+
+    case Repo.transaction(multi) do
+      {:ok, %{
+        update_user: user,
+        update_platform_customer: update_platform_customer_result,
+        update_connect_customers: update_connect_customers_results
+      }} ->
+        {:ok, user, update_platform_customer_result, update_connect_customers_results}
+      {:error, :update_user, %Ecto.Changeset{} = changeset, %{}} ->
+        {:error, changeset}
+      {:error, _failed_operation, _failed_value, _changes_so_far} ->
+        {:error, :unhandled}
+    end
+  end
+
+  defp do_update(%Changeset{} = changeset) do
+    with {:ok, user} <- Repo.update(changeset) do
+      {:ok, user, nil, nil}
+    else
+      {:error, changeset} -> {:error, changeset}
+      _ -> {:error, :unhandled}
+    end
+  end
+
+  defp update_platform_customer(%{update_user: %User{id: user_id, email: email}}) do
+    StripePlatformCustomer
+    |> Repo.get_by(user_id: user_id)
+    |> do_update_platform_customer(%{email: email})
+  end
+
+  defp do_update_platform_customer(nil, _), do: {:ok, nil}
+  defp do_update_platform_customer(%StripePlatformCustomer{} = stripe_platform_customer, attributes) do
+    {:ok, %StripePlatformCustomer{} = platform_customer, _} =
+      StripePlatformCustomerService.update(stripe_platform_customer, attributes)
+
+    {:ok, platform_customer}
+  end
+
+  defp update_connect_customers(%{update_platform_customer: nil}), do: {:ok, nil}
+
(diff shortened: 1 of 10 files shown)


joshsmith:
We discussed that we're going to need to change this to push out connect customer email changes, but that we won't store the email on our server.

joshsmith:
Marked as needs work but still needs code review, too.

joshsmith:
I am marking as low priority at the moment and removing the needs code review for the time being.

There is no mechanism right now for a user to update their email through Code Corps, and that's the only change going to Stripe.

Once we allow users to change their email – or once we have time to circle back – we can reprioritize this.

Let's not let it get out of date.

We still need the change mentioned above.

joshsmith:
I think we actually need to finish this up. Was wrong.

begedin:
I made some updates to this, so it ought to be ready for review.

When a user updates, if the updates contain an e-mail, 
* any associated platform customer record will also update
* any associated connect customer records will also update

`UserService.update(user, attributes)` will return one of the following:
* `{:ok, %User{}, :nothing_to_update, :nothing_to_update}`
* `{:ok, %User{}, %StripePlatformCustomer{}, :nothing_to_update }`
* `{:ok, %User{}, %StripePlatformCustomer{}, [{:ok, %StripeCustomer}, ...]`
* `{:error, changeset}`
* `{:error, :unhandled}`

Considering the amount of potential overlap with #508, I would prefer to merge this before touching #508 further.

lib/code_corps/services/user_service.ex
```diff
@@ -0,0 +1,99 @@
...
+        update_platform_customer: update_platform_customer_result,
+        update_connect_customers: update_connect_customers_results
+      }} ->
+        {:ok, user, update_platform_customer_result, update_connect_customers_results}
+      {:error, :update_user, %Ecto.Changeset{} = changeset, %{}} ->
+        {:error, changeset}
+      {:error, _failed_operation, _failed_value, _changes_so_far} ->
+        {:error, :unhandled}
+      other ->
+        IO.inspect(other, pretty: true)
```
joshsmith: This needs removed. Is there an `other` case that needs handled?

lib/code_corps/services/user_service.ex
```diff
@@ -0,0 +1,99 @@
...
+    end
+  end
+
+  defp do_update(%Changeset{} = changeset) do
+    with {:ok, user} <- Repo.update(changeset) do
+      {:ok, user, :nothing_to_update, :nothing_to_update}
```
joshsmith: Why not return `nil` and `nil` here?

joshsmith: `:nothing_to_update` seems overly specific here whereas `nil` is just as descriptive, IMO.

web/controllers/user_controller.ex
```diff
@@ -14,27 +15,24 @@ defmodule CodeCorps.UserController do
...
-  end
+  defp login(user, conn), do: Plug.Conn.assign(conn, :current_user, user)
 
-  def handle_update(conn, model, attributes) do
-    model
-    |> User.update_changeset(attributes)
-    |> Repo.update
-    |> CodeCorps.Analytics.Segment.track(:updated_profile, conn)
+  def handle_update(conn, record, attributes) do
+    with {:ok, user, _platform_customer_updates, _connect_customer_updates} <- UserService.update(record, attributes)
```
joshsmith: I would just go unnamed on the nil returns here, straight `_`. See no harm in that and easier to read.

web/models/stripe_connect_customer.ex
```diff
@@ -19,4 +19,9 @@ defmodule CodeCorps.StripeConnectCustomer do
     |> unique_constraint(:id_from_stripe)
     |> unique_constraint(:stripe_connect_account_id, name: :index_projects_on_user_id_role_id)
   end
+
+  def update_changeset(struct, params) do
+    struct
+    |> cast(params, [:email])
```
joshsmith: Why no `validate_required` here? as there is with platform?

test/lib/code_corps/stripe_service/stripe_platform_customer_service_test.exs
```diff
@@ -0,0 +1,21 @@
...
+
+  alias CodeCorps.StripeService.StripePlatformCustomerService
+
+  describe "update/2" do
+    test "performs update" do
+      customer = insert(:stripe_platform_customer)
+      {:ok, customer} = StripePlatformCustomerService.update(customer, %{email: "mail@mail.com"})
+      assert customer.email == "mail@mail.com"
+
+       # TODO: Figure out testing if stripe API request was made
```
joshsmith: Is this TODO leftover or intentional?

lib/code_corps/services/user_service.ex
```diff
@@ -0,0 +1,99 @@
+defmodule CodeCorps.Services.UserService do
+  @moduledoc """
+  Handles CRUD operations for users.
+
+  When operations happen on `CodeCorps.User`, we need to make sure changes
+  are propagated to related records, ex., `CodeCorps.StripePlatformCustomer` or
```
joshsmith: Should be `and`, right?

lib/code_corps/services/user_service.ex
```diff
@@ -0,0 +1,99 @@
...
+  `CodeCorps.StripeConnectCustomer`
+
+  """
+
+  alias CodeCorps.{Repo, StripeConnectCustomer, StripePlatformCustomer, User}
+  alias CodeCorps.StripeService.{StripeConnectCustomerService,StripePlatformCustomerService}
```
joshsmith: Space after comma.

web/controllers/user_controller.ex
joshsmith: It feels like the `UserService.update` should probably happen in an asynchronous task rather than sync here.

joshsmith (commented):
Some comments.

joshsmith:
@begedin feel free to merge even if there are unresolved questions here, but if so (like the async task), open in another issue and reference this.

web/controllers/user_controller.ex
begedin: We can either do it in a transaction, meaning we have to wait for sync, or we ditch the transaction, update the user, respond with that and do the propagation async. I don't think we can do both.

test/lib/code_corps/stripe_service/stripe_platform_customer_service_test.exs
begedin: It's intentional. I'm not sure how to check if a stripe API update was made. Maybe the function should return both the updated local record and the updated stripe record?

I'm starting to think we should avoid "side effects" in our functions as much as we can. I added a response containing updated platform and connect customer records to the user service for basically the same reason.

web/models/stripe_connect_customer.ex
begedin: That changeset should go away completely, actually. I added it by mistake, before realising there is nothing to update with the local record.

PR #508, opened 2016-11-29
Handle customer.source.updated, propagate platform card updates to connect card

### LINKED ISSUE #476: StripePlatformCard updates StripeConnectCards on update
# Problem

- [x] When a `StripePlatformCard` is created, it should create `StripeConnectCards` on all `StripeAccount`s for which the `User` currently has a `StripeConnectSubscription`.
- [ ] When a `StripePlatformCard` is updated, it should update the `StripeConnectCards` in its has_many.

begedin:
# What's in this PR?

This PR handles the `customer.source.updated` platform webhook.
In a single transaction, it fetches the platform card info from stripe via ID, updates the platform card record locally, then finds and updates associated connect cards, API only. It should technically update locally to, but we do not store any updatable fields for connect cards locally, so there is nothing to update.

## References
Fixes #476


JoshSmith:
I think we actually need to finish this up. Was wrong.

begedin:
This is good to review.

joshsmith (approved):

PR #513, opened 2016-11-30
Handle customer.subscription.deleted webhook

### LINKED ISSUE #509: Respond to customer.subscription.deleted event
# Problem

We're not responding to this one.

begedin:
# What's in this PR?

This PR adds handling of the `customer.subscription.deleted` event.

Since the actual subscription stripe record is not deleted, instead simply removed from the customer account, and it's status set to cancelled, the solution was to simply handle it identically to `customer.subscription.updated`.

This raises a question, though.

**Should our API, when requesting a user's subscriptions, only return those that are active, or all of them?**

If the answer is **all of them**, then we need to modify our client logic so that the donate UI is hidden/redirected from only if there is an **active** subscription, not just any subscription as it is right now.

If the answer is **only those that are active**, then we need to modify our user/project views.

## Implementation

The logical place to put the update code was `CodeCorps.StripeService.StripeConnectSubscriptionService`. 

However, I found the code there difficult to understand and manage, so adding additional code there would make it even worse. 

In an attempt to remedy this, I refactored the module:

I rewrote a few methods

I added validator modules - `ProjectSubscribable` and `UserCanSubscribe`. 

These both have a `validate/1` method, which receives a `Project` or `User` record respectively. It either returns back an `{:ok, record}` or an `{:error, :project_not_ready}` / `{:error, :user_not_ready}` if the two records do not have the necessary stripe relationships set up.  

The idea is that these validators are a simple blackbox that determines if a user or a project is able to participate in the donation system. It allows us to have less checks in the subscription creation process reduces the required size of method signatures there, making the code more readable.

Once that was in, I cleaned up further and then moved the logic from `Events.CustomerSubscriptionUpdated` into `StripeConnectSubscriptionService.update_from_stripe/2`. This same method is then used by `Events.CustomerSubscriptionDeleted`.

## References
Fixes #509

Progress on: #

### DIFF
--- a/lib/code_corps/stripe_service/stripe_connect_subscription.ex
+++ b/lib/code_corps/stripe_service/stripe_connect_subscription.ex
@@ -2,10 +2,10 @@ defmodule CodeCorps.StripeService.StripeConnectSubscriptionService do
   import Ecto.Query
 
   alias CodeCorps.{
-    Project, Repo, StripeConnectCard, StripeConnectCustomer,
+    Project, Repo, StripeConnectCustomer, StripeConnectAccount,
     StripeConnectPlan, StripeConnectSubscription, User
   }
-  alias CodeCorps.Services.ProjectService
+  alias CodeCorps.Services.{DonationGoalsService, ProjectService}
   alias CodeCorps.StripeService.{StripeConnectCardService, StripeConnectCustomerService}
   alias CodeCorps.StripeService.Adapters.StripeConnectSubscriptionAdapter
   alias CodeCorps.StripeService.Validators.{ProjectSubscribable, UserCanSubscribe}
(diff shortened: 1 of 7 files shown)


lib/code_corps/stripe_service/stripe_connect_subscription.ex
```diff
@@ -2,89 +2,79 @@ defmodule CodeCorps.StripeService.StripeConnectSubscriptionService do
...
-      {:ok, stripe_connect_subscription}
+      {:ok, %StripeConnectSubscription{} = subscription} = do_find_or_create(project, user, attributes)
+
+      ProjectService.update_project_totals(project)
+      DonationGoalsService.update_project_goals(project)
+
+      {:ok, subscription}
     else
+      {:error, :project_not_ready} -> {:error, :project_not_ready}
+      {:error, :user_not_ready} -> {:error, :user_not_ready}
```
sbatson5: It seems that all these cases return what we are matching against. What are your thoughts on just matching on `nil` and then returning everything else.

```
nil -> {:error, :not_found}
error_tuple -> error_tuple
```

lib/code_corps/stripe_service/stripe_connect_subscription.ex
```diff
@@ -112,7 +106,10 @@ defmodule CodeCorps.StripeService.StripeConnectSubscriptionService do
     |> Repo.insert
   end
 
-  defp to_create_attributes(%StripeConnectCard{} = card, %StripeConnectCustomer{} = customer, %StripeConnectPlan{} = plan, quantity) do
+  defp to_create_attributes(
+    %StripeConnectCard{} = card, %StripeConnectCustomer{} = customer,
+    %StripeConnectPlan{} = plan, %{"quantity" => quantity}
```
sbatson5: Do we need to pattern match here for `StripeConnectCard` and `StripeConnectPlan` etc? This is the only method we have by this name, so the pattern matching seems unnecessary when we could simply do:
`defp to_create_attributes(card, customer, plan, %{"quantity" => quantity)`

lib/code_corps/stripe_service/stripe_connect_subscription.ex
```diff
@@ -125,4 +122,25 @@ defmodule CodeCorps.StripeService.StripeConnectSubscriptionService do
   defp to_insert_attributes(attrs, %StripeConnectPlan{id: stripe_connect_plan_id}) do
     attrs |> Map.merge(%{"stripe_connect_plan_id" => stripe_connect_plan_id})
   end
+
+  defp retrieve_connect_account(connect_customer_id) do
+    customer = StripeConnectCustomer
+    |> Repo.get_by(id_from_stripe: connect_customer_id)
+    |> Repo.preload(:stripe_connect_account)
```
sbatson5: Can we indent here?

lib/code_corps/stripe_service/stripe_connect_subscription.ex
```diff
@@ -125,4 +122,25 @@ defmodule CodeCorps.StripeService.StripeConnectSubscriptionService do
...
+
+    {:ok, customer.stripe_connect_account}
+  end
+
+  defp load_subscription(id_from_stripe) do
+    subscription = StripeConnectSubscription
+    |> Repo.get_by(id_from_stripe: id_from_stripe)
```
sbatson5: Same as above. I believe the style is to indent when we are assigning

test/controllers/stripe_connect_events_controller_test.exs
```diff
@@ -54,9 +54,7 @@ defmodule CodeCorps.StripeConnectEventsControllerTest do
 
       assert response(conn, 200)
 
-      updated_account =
-        StripeConnectAccount
-        |> Repo.get_by(id_from_stripe: stripe_id)
+      updated_account = StripeConnectAccount |> Repo.get_by(id_from_stripe: stripe_id)
```
sbatson5: Thoughts on not piping here? `updated_account = Repo.get_by(StripeConnectAccount, id_from_stripe: stripe_id)`

test/controllers/stripe_connect_events_controller_test.exs
```diff
@@ -94,9 +92,38 @@ defmodule CodeCorps.StripeConnectEventsControllerTest do
...
+
+      plan = insert(:stripe_connect_plan, project: project)
+
+      insert(:stripe_connect_subscription,
+        id_from_stripe: stripe_id,
+        stripe_connect_plan: plan)
+
+      conn = post conn, stripe_connect_events_path(conn, :create), event
+
+      assert response(conn, 200)
```
sbatson5: Thoughts on piping this block?

`assert conn |> post(stripe_connect_events_path(conn, :create), event) |> response(200)`

sbatson5 (commented):
Just a couple of questions and comments. This is looking good 👍

joshsmith (approved):

joshsmith:
@begedin feel free to merge when you're comfortable making any changes you'd like in response to code review comments. Don't await another review.

lib/code_corps/stripe_service/stripe_connect_subscription.ex
begedin: I'd like to keep it explicit because I'm expecting we might be doing custom handling/errors at some point. I can move that into a comment above the line, though.

lib/code_corps/stripe_service/stripe_connect_subscription.ex
begedin: Good catch. The validate call at the start of the process actually does all the pattern matching required.

Issue #511, opened 2016-11-30
Check if the account is ready to start taking payments before a plan can be created


joshsmith:
# Problem

We want to be able to check that an account is actually ready to start taking payments before a plan can be created.

There is some flag on Stripe accounts that lets you know whether Stripe has actually approved them to begin taking payments.

We should modify the UI to actually prevent the user from creating any sort of donation goal or anything until this is resolved, and have a refresh button or something with some helper text.

https://projects.invisionapp.com/share/RT8PS9FK2#/screens/191325319

I think "verification" is the wrong word there and we actually mean "approve." We should also let them know if Stripe denied their account for some reason.

We'll need to update this through webhooks. Not sure which event.

begedin:
I looked into this and if I'm understanding it correctly, the field we're looking for is `charges_enabled`, as well as, possibly, `transfers_enabled` (though we should probably look at that separately.

From https://stripe.com/docs/api#account_object-charges_enabled
>`charges_enabled, boolean`

> Whether or not the account can create live charges

From https://stripe.com/docs/api#account_object-transfers_enabled


>`transfers_enabled, boolean`

>Whether or not Stripe will send automatic transfers for this account. 
This is only false when Stripe is waiting for additional 
information from the account holder.

And also https://stripe.com/docs/api#account_object-verification

>verification, MANAGED ACCOUNTS ONLY, hash

>The state of the account’s information requests, including what information is needed and by when it must be provided.

The way I understand it, if an account can charge a user, `charges_enabled` will be set to true. If it can also transfer funds from stripe to their external account, `transfers_enabled` will be set to true. If either is `false`, the `verification` hash will be populated with additional information for managed accounts only. 

We are rendering both `charges_enabled` and `transfers_enabled` in our view, so that part is not a  problem. The "refresh" button would simply trigger a `stripeConnectAccount.reload()`.

As for the correct webhook, the only one I can think of is `account.updated`, which we are already handling, so that should be fine as well.

Considering that, it ought to be enough to, on the client, add the button and prevent making a plan until `charges_enabled` is `true`. I think `transfers_enabled` could be false initially, but we may want to give the user a warning if it is false, so that they are aware they need to do something to be able to transfer funds from stripe to their own account.

begedin:
The thing that did need changing on the API side of things was to additionally check the `charges_enabled` flag before creating a plan, which I did in #514

begedin:
The API part for this is done, closing. 

[...]

PR #514, opened 2016-11-30
Add verification before enabling donations on project

begedin:
# What's in this PR?

Handles the API part of verifying that a project can enable donations. Done by adding a `Validator` module which ensures all project fields and relationships have the correct values. Similar in how it was implemented in #513 

## References
Progress on: #511

### DIFF
--- a/lib/code_corps/stripe_service/stripe_connect_plan.ex
+++ b/lib/code_corps/stripe_service/stripe_connect_plan.ex
@@ -1,28 +1,22 @@
 defmodule CodeCorps.StripeService.StripeConnectPlanService do
-  alias CodeCorps.Organization
-  alias CodeCorps.Project
-  alias CodeCorps.Repo
+  alias CodeCorps.{Project, Repo, StripeConnectPlan}
   alias CodeCorps.StripeService.Adapters.StripeConnectPlanAdapter
-  alias CodeCorps.StripeConnectAccount
-  alias CodeCorps.StripeConnectPlan
+  alias CodeCorps.StripeService.Validators.ProjectCanEnableDonations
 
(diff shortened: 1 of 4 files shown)


lib/code_corps/stripe_service/stripe_connect_plan.ex
```diff
@@ -1,28 +1,23 @@
 defmodule CodeCorps.StripeService.StripeConnectPlanService do
-  alias CodeCorps.Organization
-  alias CodeCorps.Project
+  alias CodeCorps.{Project, StripeConnectPlan}
```
sbatson5: Can we also alias `Repo` here?

lib/code_corps/stripe_service/validators/project_can_enable_donations.ex
```diff
@@ -0,0 +1,32 @@
...
+  If the project has these relationships set up, it returns `{:ok, project}`
+
+  In any other case, it returns {:error, :project_not_ready}
+  """
+  def validate(%Project{} = project), do: do_validate(project)
+
+  @invalid {:error, :project_not_ready}
+
+  defp do_validate(%Project{
+    donation_goals: [_h | _t],
```
sbatson5, line 28: Do we need to specify head and tail here?

sbatson5 (commented):
This is looking good to me 👍

lib/code_corps/stripe_service/validators/project_can_enable_donations.ex
joshsmith: @sbatson5 no, this just validates that it's non-empty.

joshsmith (approved):

PR #541, opened 2016-12-07
Add full name attribute to user view

### LINKED ISSUE #145: Add full_name to user and its view
The front-end expected a `full_name` which is not currently present on the model. This is a virtual composite of `first_name` and `last_name` for easier rendering by the client.

rossta:
# What's in this PR?

Adds a `name` attribute expected by the client to be user's first and last name.

## References
Fixes #145

@JoshSmith I believe this is what you were going for based on your comments in #145. Let me know if you had something else in mind.

### DIFF
--- a/web/views/user_view.ex
+++ b/web/views/user_view.ex
@@ -8,7 +8,7 @@ defmodule CodeCorps.UserView do
   use JaSerializer.PhoenixView
 
   attributes [
-    :biography, :email, :first_name, :last_name,
+    :biography, :email, :first_name, :last_name, :name,
     :photo_large_url, :photo_thumb_url, :state, :state_transition, :twitter,
     :username, :website, :inserted_at, :updated_at
   ]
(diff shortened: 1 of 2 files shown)


web/views/user_view.ex
```diff
@@ -41,4 +41,13 @@ defmodule CodeCorps.UserView do
     if user.id == current_user.id, do: user.email, else: ""
   end
   def email(_user, _conn), do: ""
+
+  @doc """
+  Returns the user's full name when both first and last name are present.
+  Returns only user's first or last name when one is missing, or an empty
+  string when both are missing
```
joshsmith: I think we may want to simply return `null` if both are missing. That seems more appropriate to me here. What do you think?

rossta: Yes, that makes sense since it looks like most of the optional db string columns, including `first_name` and `last_name`, default to null in the db.

rossta: Updated

joshsmith: @rossta the docs just need updated to say `nil` instead of `empty string` now.

rossta: Ah, good catch, thanks.  Updated.

joshsmith (approved):

joshsmith:
Glad to have this implemented. 👍

PR #544, opened 2016-12-07
Store events to avoid duplication. Move event handling to async tasks.

### LINKED ISSUE #512: Rework handling of Stripe webhooks
# Problem

From the stripe documentation: https://stripe.com/docs/webhooks#responding-to-a-webhook

>To acknowledge receipt of a webhook, your endpoint should return a 2xx HTTP status code. Any other information returned in the request headers or request body is ignored. All response codes outside this range, including 3xx codes, will indicate to Stripe that you did not receive the webhook. This does mean that a URL redirection or a "Not Modified" response will be treated as a failure.

>If a webhook is not successfully received for any reason, Stripe will continue trying to send the webhook once an hour for up to 3 days. Webhooks cannot be manually retried after this time, though you can query for the event to reconcile your data with any missed events.

Right now, if everything goes OK, or we outright ignore an event, we send a 200, so the webhook won't be sent anymore.

If something goes wrong, we send a 400, so according to documentation, the webhook will be resent every 3 hours for 3 days. 

That sounds like OK behavior for me, but I'm not completely sure.

## Subtasks
- [ ] Figure out if the current behavior is acceptable. Change if it it is not.
- [ ] Possibly start sending a 204 (No Content) instead of 200, since the content is ignored by stripe anyway.

## References

* https://stripe.com/docs/webhooks#responding-to-a-webhook

### LINKED ISSUE #547: Handle events that do not match the environment (livemode) with a 400 response
# Problem

Stripe webhook events can be livemode or non-livemode. 

Right now, the production environment handles livemode events only, while the other environments handle non-livemode events only.

However, even if the event is ignored, our API responds to the webhook request with a 204, which Stripes inteprets as "handled, do not send again", since it's in the 2xx range.

I think it's possible we should be sending some other code here, so Stripe keeps resending the event until it hits the correct  environment. Then again, considering each of our environments has a separate set of stripe accounts, I'm not sure. 

I think we should disuss this.

begedin:
# What's in this PR?

This PR adds a `StripeEvent` model and associated table.

Then it hooks into the current webhook handling process.

As a webhook request is received, a `StripeEvent` record is stored, with fields

```
* `id_from_stripe` - the id of the Stripe event
* `status` - the state of the event. Automatically set to `"unprocessed"`
```

Once the request is processed, the created event is set to either `"processed"` or `"errored"` depending on the processing result.

## Details

* We first attempt to find the event in the database, using `:id_from_stripe`. The event record is only created when there is no such event stored already.
* If the event record is found, then it only get's processed if the status is `"unprocessed"` or `"errored"`. An already processed event is ignored.
* The endpoint returns a 204 if the event was ignored due to being processed already, or due to the system not supporting that specific event type.
* In all other cases, the endpoint returns a 200

## References
Fixes #512
Fixes #547

### DIFF
--- a/lib/code_corps/stripe_service/stripe_platform_customer.ex
+++ b/lib/code_corps/stripe_service/stripe_platform_customer.ex
@@ -56,7 +56,7 @@ defmodule CodeCorps.StripeService.StripePlatformCustomerService do
   - `{:error, :unhandled}` -if something unexpected went wrong
   """
   def update_from_stripe(id_from_stripe) do
-    with customer                                            <- Repo.get_by(StripePlatformCustomer, id_from_stripe: id_from_stripe),
+    with %StripePlatformCustomer{} = customer                <- Repo.get_by(StripePlatformCustomer, id_from_stripe: id_from_stripe),
          {:ok, %Stripe.Customer{} = stripe_customer}         <- @api.Customer.retrieve(id_from_stripe),
          {:ok, params}                                       <- StripePlatformCustomerAdapter.to_params(stripe_customer, %{}),
          {:ok, %StripePlatformCustomer{} = platform_customer, connect_customer_updates} <- perform_update(customer, params)
--- a/lib/code_corps.ex
+++ b/lib/code_corps.ex
@@ -18,7 +18,7 @@ defmodule CodeCorps do
       supervisor(CodeCorps.Endpoint, []),
       # Start your own worker by calling: CodeCorps.Worker.start_link(arg1, arg2, arg3)
       # worker(CodeCorps.Worker, [arg1, arg2, arg3]),
-
+      supervisor(Task.Supervisor, [[name: :webhook_processor, restart: :transient]]),
       worker(Segment, [Application.get_env(:segment, :write_key)])
     ]
 
--- a/lib/code_corps/stripe_service/stripe_connect_customer.ex
+++ b/lib/code_corps/stripe_service/stripe_connect_customer.ex
@@ -19,7 +19,7 @@ defmodule CodeCorps.StripeService.StripeConnectCustomerService do
     end
   end
 
-  def update(%StripeConnectCustomer{id_from_stripe: id_from_stripe, stripe_connect_account: connect_account} = connect_customer, attributes) do
+  def update(%StripeConnectCustomer{id_from_stripe: id_from_stripe, stripe_connect_account: connect_account}, attributes) do
     @api.Customer.update(id_from_stripe, attributes, connect_account: connect_account.id_from_stripe)
   end
 
--- a/lib/code_corps/stripe_service/stripe_platform_card.ex
+++ b/lib/code_corps/stripe_service/stripe_platform_card.ex
@@ -81,7 +81,7 @@ defmodule CodeCorps.StripeService.StripePlatformCardService do
     stripe_platform_card |> Map.take([:exp_month, :exp_year, :name])
   end
 
-  defp do_update_connect_cards(stripe_platform_card, attributes) when attributes == %{}, do: []
+  defp do_update_connect_cards(_stripe_platform_card, attributes) when attributes == %{}, do: []
   defp do_update_connect_cards(stripe_platform_card, attributes) do
     stripe_platform_card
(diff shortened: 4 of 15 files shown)


lib/code_corps/stripe_service/stripe_platform_customer.ex
```diff
@@ -56,7 +56,7 @@ defmodule CodeCorps.StripeService.StripePlatformCustomerService do
   - `{:error, :unhandled}` -if something unexpected went wrong
   """
   def update_from_stripe(id_from_stripe) do
-    with customer                                            <- Repo.get_by(StripePlatformCustomer, id_from_stripe: id_from_stripe),
+    with %StripePlatformCustomer{} = customer                <- Repo.get_by(StripePlatformCustomer, id_from_stripe: id_from_stripe),
```
begedin, line 59: `get_by` might return a `nil`. We weren't restricting the type.

joshsmith:
We should still consider doing the following here:

- Add `processing` status, respond with 200 immediately, and kick off asynchronous task that updates later
- Add another field for the event type (for easily looking at events in the database, in logs, etc)
- Add `unhandled` status for unhandled events instead of a 204, since Stripe will keep trying to hit us for all these "unhandled" events

joshsmith:
@begedin did not get a chance to implement my comment, but think all of that should still be implemented.

begedin:
> Add `processing` status, respond with 200 immediately, and kick off asynchronous task that updates later

Working on this one

> Add another field for the event type (for easily looking at events in the database, in logs, etc)

Done

> Add `unhandled` status for unhandled events instead of a 204, since Stripe will keep trying to hit us for all these "unhandled" events

We now store the event as "unhandled". Returning a 204, however, is not a problem. It's just for us, so we can easily test the response and see that the server differentiates. According to stripe documentation, any 2xx response will be considered as "event is done, do not send it again".

That being said, we only return a 204 in the case of "ignored by environment", because we, for example, got a livemode event on staging or vice-versa. We do not return it in case of "unhandled", since in that case, we should just consume the event and (for now) not do anything about it. In this case, we return a 200, same as "errored" or "processed"

From https://stripe.com/docs/webhooks#responding-to-a-webhook

>To acknowledge receipt of a webhook, **your endpoint should return a 2xx HTTP status code**. Any other information returned in the request headers or request body is ignored. All response codes outside this range, including 3xx codes, will indicate to Stripe that you did not receive the webhook.

begedin:
I have now added async processing of webhooks. 

In order to achieve this, we use a `Task.Supervisor`, referenced `:webhook_processor`

Testing was giving me trouble with this one, but I managed to get around it by explicitly waiting for the child processes to exit, using a `&Process.monitor/1` followed by an `assert_received`. It's not perfect, but I don't have a better idea. I could use an experts thoughts on this one. Should I maybe "inject" our supervisor and stub it out with a test module, similar to how we do it with our services?

begedin:
The process is now as follows:

* Controller will call `WebhookProcessor.process_async(json, handlermodule)`
* `process_async` will use a supervisor to handle the event. It will immediately return either `{:ok, pid}` or `{:ok, :ignored_by_environment}`. For the first result, the reply will be 200, for the second, 204. Both are considered "handled, do not send again" by stripe.
* The supervisor will run the task. Event will be handled according to our handler setup
  * Event record is initially created with status `processing`
  * If event already exists and its status is already `processing`, the task just returns with `nil`
  * If event already exists and its status is something else, we try to process it again
  * If processing is completed successfully, event will be set to `processed`
  * If there is some sort of error, event will be set to `errored`
  * If we do not handle the type of event, it will be set to `unhandled`

begedin:
Something we should consider adding but is probably out of scope for this PR:

* [ ] Right now, we respond with 2xx for all events, so they do not get resent. Hower, some events are currently unhandled because we do not support the type yet. Some simply error out. Both of these should probably be either periodically restarted or handled in some other way. We Should figure out what to do with them and make an issue prior to merging this one.

joshsmith:
@begedin if they error out or are unhandled I think that means human intervention time, not machine. We haven't written code yet to effectively handle the edge case, but we can go in and process those events since we're keeping track.

I think were in a good state to merge now.

joshsmith (approved):

PR #552, opened 2016-12-09
Code spike for switch to managed accounts

begedin:
# What's in this PR?

This is a code spike to make the switch to managed stripe accounts

What I got

1. connect account create endpoint recieves arguments from client
2. calls `StripeService.StripeManagedConnectAccountService.create`
3. Adapter parses arguments and creates a param hash to create account on stripe API
4. Record is created using stripity-stripe (the latest commit from https://github.com/code-corps/stripity_stripe/pull/153) and then inserted into the db locally.

NOTES:

Haven't added/changed any database fields yet. Not sure which ones will be needed. I use those we have, others are virtual for now.

## References
* Paired with https://github.com/code-corps/code-corps-ember/pull/849


web/controllers/stripe_connect_account_controller.ex
```diff
@@ -3,15 +3,15 @@ defmodule CodeCorps.StripeConnectAccountController do
   use JaResource
 
   alias CodeCorps.StripeConnectAccount
-  alias CodeCorps.StripeService.StripeConnectAccountService
+  alias CodeCorps.StripeService.StripeManagedConnectAccountService
```
joshsmith: Would just keep this `StripeConnectAccountService` personally. `managed` is just a param.

joshsmith:
Hadn't had a chance to comment on this, but may want to consider reducing scope by pulling off some of the individual tasks in the milestone now that they're around.

lib/code_corps/stripe_service/adapters/stripe_connect_account.ex
```diff
@@ -18,6 +19,102 @@ defmodule CodeCorps.StripeService.Adapters.StripeConnectAccountAdapter do
...
+
+    legal_entity = %{
+      business_name: business_name,
+      business_tax_id: business_ein,
+      dob: %{day: dob_day, month: dob_month, year: dob_year},
+      first_name: first_name,
+      last_name: last_name,
+      ssn_last_4: ssn_last4,
+      type: recipient_type,
+      # TODO: Decide on setting legal_entity.address or legal_entity.personal_address
```
joshsmith: I think this is mentioned in that Ember issue.

lib/code_corps/stripe_service/stripe_managed_connect_account.ex
```diff
@@ -0,0 +1,17 @@
...
+    with {:ok, stripe_params} <- StripeConnectAccountAdapter.to_stripe_params(attributes),
+         {:ok, %Stripe.Account{} = stripe_account} <- @api.Account.create(stripe_params),
+         {:ok, params} <- StripeConnectAccountAdapter.to_managed_params(stripe_account, attributes)
+    do
+      %CodeCorps.StripeConnectAccount{}
+      |> CodeCorps.StripeConnectAccount.managed_create_changeset(params)
+      |> CodeCorps.Repo.insert
+    end
+  end
+end
```
joshsmith: I rewrote some of this with pseudocode in #575. I would consider taking the pattern matching of that function. Would also just use `create_changeset/2` rather than a special one for `managed`, per comment above.

web/models/stripe_connect_account.ex
```diff
@@ -6,28 +6,67 @@ defmodule CodeCorps.StripeConnectAccount do
...
 
+  @managed_create_fields [
+    :address1, :address2, :business_ein, :business_name, :business_type,
+    :city, :country, :dob_day, :dob_month, :dob_year, :email,
+    :first_name, :last_name, :recipient_type, :ssn_last4,
+    :state, :zip, :organization_id,
+    :managed, :id_from_stripe, :charges_enabled, :transfers_enabled
+  ]
+
+  def managed_create_changeset(struct, params) do
```
joshsmith, line 63: Don't think we want distinction between `create` and `managed_create`. Also, `create` fields will be just `country` and  `managed`, where the latter is always `true`.

web/controllers/stripe_connect_account_controller.ex
begedin: It was just a temporary thing so I can work on a new one while still having the old one for reference.

joshsmith:
I'm going to close this since I think it's pretty out of date.

Issue #564, opened 2016-12-10
Add StripeConnectAccountService create test


joshsmith:
# Problem

We need to test the `StripeConnectAccountService`'s `create` to create an account with `managed: true` and the country code only.

PR #575, opened 2016-12-12
Remove StripeAuth everywhere from the API

### LINKED ISSUE #556: Remove StripeAuth everywhere from the API
# Problem

We need to remove the `StripeAuth` model and everything related to it.

We are creating Stripe managed accounts and no longer need OAuth to be able to connect to Stripe.

## Subtasks
- [x] Remove the StripeAuth model and tests (do not have a table to drop since it's virtual)
- [x] Remove from `web/models/abilities.ex`
- [x] Remove controller and tests
- [x] Remove from router
- [x] Remove view and tests
- [x] Remove reference to Stripe Auth from the API blueprint
- [x] Remove `CodeCorps.StripeTesting.Connect.OAuth`
- [x] Remove references to `Connect.OAuth` from `StripeConnectAccountService`
- [x] Remove anywhere that mentions the `stripe_redirect_uri` (including mention in USAGE.md)

joshsmith:
# What's in this PR?

Removes StripeAuth everywhere since we're lo longer doing OAuth connection for standalone accounts.

## References
Fixes #556


joshsmith:
This is a little hacky on the implementation of `StripeConnectAccountService.create/1`, but I figured that we could always work to replace this in #564 when the behavior actually changes, and depending on upstream changes from `stripity_stripe` this will change anyway. This decision here at least unblocks that without much detriment at present.

begedin:
I so love red PRs :)

begedin (approved):

PR #576, opened 2016-12-14
Add stripe external account model and tests

### LINKED ISSUE #571: Create StripeExternalAccount model and tests
# Problem

We need to create a `StripeExternalAccount` model and its tests.

You can only create an external account.

It must have these fields: `id_from_stripe`, `account_id_from_stripe`, `account_holder_name`, `account_holder_type`, `bank_name`, `country`, `currency`, `default_for_currency`, `fingerprint`, `last4`, `routing_number`, `status`

Stripe's docs: https://stripe.com/docs/api#update_account-external_account

joshsmith:
# What's in this PR?

@paulsullivanjr's work adding an external account model, migration and test.

## References
Fixes #571

### DIFF
--- a/test/models/stripe_external_account_test.exs
+++ b/test/models/stripe_external_account_test.exs
@@ -1,0 +1,18 @@
+defmodule CodeCorps.StripeExternalAccountTest do
+  use CodeCorps.ModelCase
+
+  alias CodeCorps.StripeExternalAccount
+
+  @valid_attrs %{account_id_from_stripe: "some content", id_from_stripe: "some content"}
+  @invalid_attrs %{}
+
+  test "changeset with valid attributes" do
+    changeset = StripeExternalAccount.changeset(%StripeExternalAccount{}, @valid_attrs)
+    assert changeset.valid?
+  end
+
+  test "changeset with invalid attributes" do
+    changeset = StripeExternalAccount.changeset(%StripeExternalAccount{}, @invalid_attrs)
+    refute changeset.valid?
+  end
+end
--- a/web/models/stripe_external_account.ex
+++ b/web/models/stripe_external_account.ex
@@ -1,0 +1,29 @@
+defmodule CodeCorps.StripeExternalAccount do
+  use CodeCorps.Web, :model
+
+  schema "stripe_external_accounts" do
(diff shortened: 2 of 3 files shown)


web/models/stripe_external_account.ex
```diff
@@ -0,0 +1,29 @@
...
+    timestamps()
+  end
+
+  @doc """
+  Builds a changeset based on the `struct` and `params`.
+  """
+  def changeset(struct, params \\ %{}) do
+    struct
+    |> cast(params, [:id_from_stripe, :account_id_from_stripe, :account_holder_name, :account_holder_type, :bank_name, :country, :currency, :default_for_currency, :fingerprint, :last4, :routing_number, :status])
+    |> validate_required([:id_from_stripe, :account_id_from_stripe, :account_holder_name, :account_holder_type, :bank_name, :country, :currency, :default_for_currency, :fingerprint, :last4, :routing_number, :status])
```
joshsmith: Should probably split these out into an `@`.

Also don't know if we should mark all as `required`. I think only ones required would be `id_from_stripe` and `account_id_from_stripe`.

test/models/stripe_external_account_test.exs
```diff
@@ -0,0 +1,18 @@
+defmodule CodeCorps.StripeExternalAccountTest do
+  use CodeCorps.ModelCase
+
+  alias CodeCorps.StripeExternalAccount
+
+  @valid_attrs %{account_holder_name: "some content", account_holder_type: "some content", account_id_from_stripe: "some content", bank_name: "some content", country: "some content", currency: "some content", default_for_currency: "some content", fingerprint: "some content", id_from_stripe: "some content", last4: "some content", routing_number: "some content", status: "some content"}
+  @invalid_attrs %{}
```
joshsmith, line 7: Per the comment below, can probably use less attrs here for it to be valid.

joshsmith (commented):
@paulsullivanjr we could use some minor changes here.

Thanks for doing this!

joshsmith:
🙌  nice work @paulsullivanjr

paulsullivanjr:
thanks. I can make the changes you suggested. Should be able to do it tonight.

PR #577, opened 2016-12-15
Fetch webhook events from Stripe

### LINKED ISSUE #550: Events in webhooks need to be fetched from the Stripe API
# Problem

We need to fetch events received by webhooks from the API before processing them in order to avoid possible replay attacks.

The only data that we should be capturing is the `id_from_stripe`, the `type`, and a `user_id` if it has it.

The `id_from_stripe`, along with the `user_id` (the Stripe Connect account's ID), should then be used to retrieve the Event from the API and proceed with any processing.

> When using Connect, you can also receive notifications of events that occur in connected accounts. For these events, there will be an additional user_id attribute in the received Event object.

Since we're making this change, we might as well fold in some other changes.

## Subtasks
- [x] Add `user_id` to `StripeEvent` (can be `null`)
- [x] Add `endpoint` to `StripeEvent` (`platform` or `connect`, not null, no default)
- [x] Attempt to insert the `id_from_stripe`, `type`, `user_id`, `endpoint`, and a `status` of `processing` into the database immediately, then kick off the async process and respond to Stripe, otherwise: do nothing
- [x] Fetch `Event` from Stripe's API before processing – do not pass any attributes collected from webhook, simply use the `id_from_stripe` and `user_id` (if needed) to fetch from Stripe

joshsmith:
# What's in this PR?

Adds everything in #550 that we need, namely retrieving both connect and platform events from Stripe before processing.

I believe there are some opportunities for refactoring, so would really love advice on that.

## References
Fixes #550


lib/code_corps/stripe_service/webhook_processing/webhook_processor.ex
```diff
@@ -5,79 +5,104 @@ defmodule CodeCorps.StripeService.WebhookProcessing.WebhookProcessor do
...
+  def process_async(%{"id" => id, "livemode" => livemode, "user_id" => user_id} = json, handler) do
+    case event_matches_environment?(livemode) do
+      true -> do_process_async(id, user_id, handler, json)
+      false -> {:error, :ignored_by_environment}
     end
   end
-
-  defp do_process_async(json, handler) do
-    Task.Supervisor.start_child(:webhook_processor, fn -> do_process(json, handler) end)
+  def process_async(%{"id" => id, "livemode" => livemode} = json, handler) do
```
rossta, line 38: It looks like you can consolidate this def with the one on line 32 unless I'm missing something:

```elixir
def process_async(%{"id" => id, "livemode" => livemode} = json, handler) 
  case event_matches_environment?(livemode) 
    true -> do_process_async(id, json["user_id"], handler, json)
    false -> {:error, :ignored_by_environment}
  end
end

Comment (joshsmith) [reply lib/code_corps/stripe_service/webhook_processing/webhook_processor.ex:38]: @rossta the one on line 32 is doing pattern matching against the user_id key, so on platform events the signature will be off.

File: lib/code_corps/stripe_service/webhook_processing/webhook_processor.ex [hunk 2]
```diff
@@ -5,79 +5,104 @@ defmodule CodeCorps.StripeService.WebhookProcessing.WebhookProcessor do
...
+    do
+      handle_event(json, event, handler)
     else
       {:error, :already_processing} -> nil
     end
   end
 
-  defp find_or_create_event(id_from_stripe, type) do
+  defp event_matches_environment?(livemode) do
+    case Application.get_env(:code_corps, :stripe_env) do
```
rossta, line 61: The `livemode` value is just `true/false`, no? How about just the following as the method body? 

`livemode && Application.get_env(:code_corps, :stripe_env) == :prod`

lib/code_corps/stripe_service/webhook_processing/webhook_processor.ex
```diff
@@ -5,79 +5,104 @@ defmodule CodeCorps.StripeService.WebhookProcessing.WebhookProcessor do
...
+    %StripeEvent{} |> StripeEvent.create_changeset(%{endpoint: endpoint, id_from_stripe: id_from_stripe, type: type, user_id: user_id}) |> Repo.insert
   end
 
   defp set_errored(%StripeEvent{} = event) do
     event |> StripeEvent.update_changeset(%{status: "errored"}) |> Repo.update
   end
 
+  defp set_processed(%StripeEvent{} = event) do
+    event |> StripeEvent.update_changeset(%{status: "processed"}) |> Repo.update
+  end
```
rossta, line 104: I often struggle with the question of whether or not to introduce another helper method for helper methods like `set_(errored|processed|unhandled)` when all three have similar implementation. However, another layer of indirection isn't always helpful and I think it's fine the way you have it here with already-small methods.

lib/code_corps/stripe_service/webhook_processing/webhook_processor.ex
joshsmith: @rossta we still want to be able to process events in staging, so we would never handle events with this situation.

rossta:
@JoshSmith Nice work! I made a few comments on little stuff. Probably warrants a more-thorough review from someone who knows the codebase better than me.

lib/code_corps/stripe_service/webhook_processing/webhook_processor.ex
rossta: Yah, but isn't it just passing the `user_id` value into the same place where it would be `nil` in the other example? So I thought `json["user_id"]` as the second param to the `true` case would handle both implementations but I could be wrong.

lib/code_corps/stripe_service/webhook_processing/webhook_processor.ex
joshsmith: Yeah, I wasn't sure about this, hemmed and hawed, but ultimately this is probably okay. Was thinking we could just have done `set_status/2` with a string as a second argument.

lib/code_corps/stripe_service/webhook_processing/webhook_processor.ex
rossta: Gotcha

joshsmith: I tried thinking about how to improve this because it reads strange, but think it's isolated enough to not make a difference, and is kind of addressed in the moduledoc.

lib/code_corps/stripe_service/webhook_processing/webhook_processor.ex
joshsmith: @rossta yeah, I'm not sure how I feel about it. On the one hand I like the explicitness of saying that there are two different ways we handle this, but I suppose it's the same outcome.

joshsmith: I did miss your code example somehow earlier so just read "consolidate" somehow.

rossta: I get what you mean about being explicit. In that case, I'm usually inclined to extract an intention-revealing variable or method to help highlight the difference. That may not be the best way to handle it as I'm still learning my way around the conventions in the codebase.

joshsmith: It's possible that this interface should be redesigned to have two public fn's, one for processing `platform` and one for `connect` events, and then the controller does some pattern matching against `user_id`. I believe that every `connect` event should have a `user_id`, but I could be wrong.

lib/code_corps/stripe_service/webhook_processing/webhook_processor.ex
```diff
@@ -5,79 +5,104 @@ defmodule CodeCorps.StripeService.WebhookProcessing.WebhookProcessor do
...
-    event |> StripeEvent.update_changeset(%{status: "processed"}) |> Repo.update
+  defp infer_endpoint_from_handler(ConnectEventHandler), do: {:ok, "connect"}
+  defp infer_endpoint_from_handler(PlatformEventHandler), do: {:ok, "platform"}
+  defp infer_endpoint_from_handler(_), do: {:error, :invalid_handler}
+
+  defp retrieve_event_from_api(id, nil), do: @api.Event.retrieve(id)
+  defp retrieve_event_from_api(id, user_id), do: @api.Event.retrieve(id, connect_account: user_id)
+
+  defp create_event(id_from_stripe, endpoint, type, user_id) do
+    %StripeEvent{} |> StripeEvent.create_changeset(%{endpoint: endpoint, id_from_stripe: id_from_stripe, type: type, user_id: user_id}) |> Repo.insert
```
green-arrow, line 95: Do we follow any conventions on when to split out a pipe to a new line?

green-arrow (approved):
Looks good to me

lib/code_corps/stripe_service/webhook_processing/webhook_processor.ex
joshsmith: We nominally use `credo` but don't have it enabled on the repo and honestly haven't run it in CLI in awhile.

joshsmith:
I'm going to refactor this to make a single public API that pattern matches against the handlers. I think given that there are two endpoints, it makes sense to split the API here to respond to the two handlers for those endpoints. At the controller level, we can pattern match against `"user_id"` and pass it in from there.

Think this will be a good refactoring, but will do it as a separate commit in case y'all disagree.

joshsmith:
Decided not to refactor. If needed later, I'll circle back. For now, that's an optimization.

Issue #573, opened 2016-12-11
Add a webhook for account.external_account.created


joshsmith:
# Problem

We need a webhook to listen on `account.external_account.created` (right now for bank accounts), so that we can fetch the bank account from Stripe's API and then create the `StripeBankAccount`, attached to the `StripeConnectAccount` for the `user_id` in the event.

joshsmith:
~This is blocked by https://github.com/code-corps/stripity_stripe/pull/153 which is necessary for us to traverse the object from Stripe.~

joshsmith:
Actually, I was wrong, this is blocked by #571.

paulsullivanjr:
Hopefully I can start working on this tonight.

PR #583, opened 2016-12-19
Add StripeConnectAccount update endpoint for assigning an external account

### LINKED ISSUE #572: Add StripeConnectAccountController update for setting external account
# Problem

We want to add an `update` on `StripeConnectAccountController` to allow for setting `external_account` to a bank account token string.

We likely want to listen for the bank account being created inside a webhook rather than trying to deal with it in one go, to reduce complexity. Can be a separate task.

begedin:
# What's in this PR?

Adds an update endpoint for StripeConnectAccount, which only allows assigning an external account. Everything else is reported as an unauthorized request.

## References
Fixes #572

### DIFF
--- a/test/lib/code_corps/stripe_service/stripe_connect_account_service_test.exs
+++ b/test/lib/code_corps/stripe_service/stripe_connect_account_service_test.exs
@@ -3,6 +3,7 @@ defmodule CodeCorps.StripeService.StripeConnectAccountServiceTest do
 
   use CodeCorps.ModelCase
 
+  alias CodeCorps.{StripeConnectAccount}
   alias CodeCorps.StripeService.StripeConnectAccountService
 
   describe "create" do
@@ -11,12 +12,22 @@ defmodule CodeCorps.StripeService.StripeConnectAccountServiceTest do
 
       attributes = %{"country" => "US", "organization_id" => organization.id}
 
-      {:ok, %CodeCorps.StripeConnectAccount{} = connect_account} =
-              StripeConnectAccountService.create(attributes)
+      {:ok, %StripeConnectAccount{} = connect_account} =
+        StripeConnectAccountService.create(attributes)
 
       assert connect_account.country == "US"
       assert connect_account.organization_id == organization.id
       assert connect_account.managed == true
+    end
+  end
+
+  describe "add_external_account/2" do
+    test "assigns the external_account property to the record" do
+      account = insert(:stripe_connect_account)
+
+      {:ok, %StripeConnectAccount{} = updated_account} =
+        StripeConnectAccountService.add_external_account(account, "ba_test123")
+      assert updated_account.external_account == "ba_test123"
     end
   end
 end
--- a/lib/code_corps/stripe_service/adapters/stripe_connect_account.ex
+++ b/lib/code_corps/stripe_service/adapters/stripe_connect_account.ex
@@ -66,6 +66,7 @@ defmodule CodeCorps.StripeService.Adapters.StripeConnectAccountAdapter do
       stripe_account
       |> Map.from_struct
       |> transform_map(@stripe_mapping)
+      |> add_nested_attributes(stripe_account)
       |> keys_to_string
       |> add_non_stripe_attributes(attributes)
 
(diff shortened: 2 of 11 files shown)


joshsmith (approved):

joshsmith:
@begedin LGTM

joshsmith:
@begedin to be clear, though, this does not fix #572 in its entirety.

joshsmith:
We still need to allow for updates to the StripeConnectAccount for all the other changes that can be made from the Ember app, like the personal/business information, identity document, etc.

begedin:
I would either rewrite the issue description in #572 then, or change it's title and open a new issue. The description makes it seem like only the `external_account` field is supposed to be updatable by it.

I would actually lean against too specific update actions (this one, and the one proposed for attaching a verification document), but the issue description made it seem to me like this is what's required.

joshsmith:
Agree wholeheartedly on not having separate update actions, and think we should avoid that here. I'm definitely onboard with that. The only case where they are advisable, I think, is when they kick off complex separate actions that result in hitting Stripe's API at multiple endpoints. That's obviously avoidable here and instead we're listening for the external account solely in a webhook. Single update here, single update to Stripe.

test/lib/code_corps/stripe_service/stripe_connect_account_service_test.exs
```diff
@@ -12,12 +12,22 @@ defmodule CodeCorps.StripeService.StripeConnectAccountServiceTest do
...
+
+  describe "add_external_account/2" do
+    test "assigns the external_account property to the record" do
+      account = insert(:stripe_connect_account)
+
+      {:ok, %StripeConnectAccount{} = updated_account} =
+        StripeConnectAccountService.add_external_account(account, "ba_test123")
+      assert updated_account.external_account == "ba_test123"
+    end
+  end
```
joshsmith, line 32: Really I think this should just be `update` now, no?

joshsmith (changes requested):

joshsmith:
I think the issue for #572 was written incorrectly by me and instead needs to focus on adapting to changes brought into `stripity_stripe` by @SViccari.

begedin:
This needs more work. The process should be

* update the Stripe API connect account record with `external_account`.
* this will trigger a `account.external_account.created` webhook, handled in #573. Handling the webhook will insert an external account record into our database.
* however, we should update the `stripe_connect_account` record as part of this request immediately, using the response we get from the stripe API. Otherwise, the account creation status fields will not update in time.

joshsmith:
@begedin stripity_stripe account update is merged now.

begedin:
@JoshSmith I'm not seeing any code related to `Stripe.Account.update` in `stripity_stripe`, merged or otherwise.

What did get merged was the rewrite of the conversion process.

I'll start a PR of my own.

begedin:
@JoshSmith What this does now is:

* controller receives an update request with the parameter "external_account"
* `handle_update` matches against this parameter only. There is a catch-all match that just returns a 403 (we can add in  matchers for document/recipient in other PRs).
* `StripeConnectAccountService.add_external_account` puts this change to the stripe api (new PR for `Account.update`).
* Api replies with a `Stripe.Account` record, which we use to update the local record.

NOTES:

* Right now, a Stripe list object is not handled properly with stripity stripe. Once that happens, we need to do a minor change here.
* I had to modify `StripeTesting.Account` a good amount to allow for updates. Sadly, the `Account` object is complex, so I had to add helpers to build a fixture with nested data. I don't like it, but I couldn't think of a better way.
* Tests are failing because we need the `stripity_stripe` PR merged. I used a local package. https://github.com/code-corps/stripity_stripe/pull/166

joshsmith:
@begedin can you bring in the new stripity_stripe or rerun the build so it does so?

begedin:
@JoshSmith stripity_stripe is up to date now.

Regardless of merge order, this is bound to have some conflicts with the other PRs. I recommending merging #588 first, because using that as a baseline will make resolving conflicts easier.

joshsmith (approved):

PR #596, opened 2016-12-31
Fix StripeConnectAccount

joshsmith:
# What's in this PR?

Attempt to fix StripeConnectAccount's update fn to allow for updating.


begedin:
I did some long overdue fixes here, which we needed due to some of the changes the PR contains

* `MapUtils.keys_to_string` now does a deep transform and also supports lists
* The `json_payload` `ExMachina` strategy has been replaced with a more accurate `build_json_payload` helper.
* Added an `:update` policy for stripe connect account, still needs testing

## Note on replacing strategy

The strategy had several problems. The first one is that it uses the associated record's default view in order to serialize. The problem there is that it wrongly assumes that the attributes we send in a `:create` or an `:update` request is the exact same set we receive back in a `:show`

This is wrong for several records now.

Additionally, it adds **all attributes** defined by the view to the payload. A `PATCH` request by definition should only contain the attributes that are changed. The new helper supports that definition.

begedin:
Started cleanup of model tests

Added two new helpers, `assert_validation_triggered(changeset, field, type)` and `assert_error_message(changeset, field, message)`. This allows a more generalized approach to checking changeset validation errors so it's a just a matter of replacing the existing assertions with these.

lib/code_corps/stripe_service/adapters/stripe_connect_account.ex
```diff
@@ -93,17 +125,31 @@ defmodule CodeCorps.StripeService.Adapters.StripeConnectAccountAdapter do
     |> Map.take(@non_stripe_attributes)
   end
 
+  defp remove_read_only_values(attributes, [head | tail]) do
+    remove_read_only_values(attributes, head)
+    remove_read_only_values(attributes, tail)
```
begedin: How does this work? Seems to me like it only returns the result from removing read only values from the final item in the list (after it recursively gets through the tail). Are we missing some chaining here?

lib/code_corps/stripe_service/util.ex
```diff
@@ -10,6 +10,53 @@ defmodule CodeCorps.StripeService.Util do
   """
   def transform_map(api_map, mapping), do: mapping |> Enum.reduce(%{}, &map_field(&1, &2, api_map))
 
+  def transform_attributes(attributes, mapping) do
+    attributes_map = map_keys_to_atoms(attributes)
+    mapping |> Enum.reduce(%{}, &map_attribute(&1, &2, attributes_map))
+  end
```
begedin, line 16: Might not be a bad Idea to have a unit test for this specifically. Separate issue might be good enough for now, though.

begedin (dismissed):
Have a question on one bit of code, but that's about it for now. Took me a while to get through.

lib/code_corps/stripe_service/adapters/stripe_connect_account.ex
joshsmith: I think we should remove this altogether, actually.

lib/code_corps/stripe_service/util.ex
joshsmith: 👍 for separate issue.

lib/code_corps/stripe_service/stripe_connect_plan.ex
```diff
@@ -22,11 +22,12 @@ defmodule CodeCorps.StripeService.StripeConnectPlanService do
...
   end
 
-  defp get_create_attributes do
+  @spec get_create_attributes(binary) :: map
+  defp get_create_attributes(project_id) do
     %{
       amount: 1, # in cents
       currency: "usd",
-      id: "month",
+      id: "month_project_" <> to_string(project_id),
```
begedin, line 30: Did you mean `month_plan_` here or is `_project_` intentional?

lib/code_corps/stripe_service/adapters/stripe_connect_account.ex
```diff
@@ -62,6 +62,36 @@ defmodule CodeCorps.StripeService.Adapters.StripeConnectAccountAdapter do
     {:verification_fields_needed, [:verification, :fields_needed]}
   ]
 
+  @stripe_update_read_only [
```
begedin: Is this still used for something? I tried searching for it through the file but got nothing.

begedin (changes requested):
Minor questions. Otherwise seems good to go.

lib/code_corps/stripe_service/stripe_connect_account.ex
```diff
@@ -16,43 +14,17 @@ defmodule CodeCorps.StripeService.StripeConnectAccountService do
...
-        # to the account changeset and return that to be rendered.
-        {:error, :stripe_file_upload, %Ecto.Changeset{} = _, _} ->
-          account_changeset |> Ecto.Changeset.add_error(:legal_entity_verification_document, "is invalid")
-          {:error, account_changeset}
-      end
+      {:ok, updated_account}
+    else
+      {:error, %Ecto.Changeset{} = changeset} -> {:error, changeset}
+      {:error, %Stripe.APIErrorResponse{} = error} -> {:error, error}
+      _ -> {:error, :unhandled}
```
begedin, line 27: We will likely want to sweep through the code at some point and remove these `_ -> unhandled` cases. They basically swallow any unexpected errors in the process. We should explicitly list any failures we support and handle. If anything unexpected fails, we should treat it as a bug. 

This is basically equivalent with us wrapping a huge chunk of code into a `try-catch` that catches any sort of exception.

lib/code_corps/stripe_service/stripe_connect_plan.ex
joshsmith: @begedin project is intentional since plans are per-project.

begedin (approved):

PR #604, opened 2017-01-03
Hook up stripe external account with connect account, expose bank_account info

### LINKED ISSUE #597: Expose and/or save bank account  last 4 and routing number
# Problem

In the front-end our `bank-account` component displays a `bankAccountRoutingNumber` and `bankAccountLast4` that the API's `StripeConnectAccount` does not appear to track. This might possibly be in the `external_account`'s information, but no such relationship exists in the API.

paulsullivanjr:
# What's in this PR?

This is in progress and not complete. Initial changes to associated the external account with StripeConnectAccount to provide access to the last 4 digits of the account and routing numbers. 

## References
Fixes #597

### DIFF
--- a/lib/code_corps/stripe_service/stripe_connect_external_account_service.ex
+++ b/lib/code_corps/stripe_service/stripe_connect_external_account_service.ex
@@ -1,16 +1,24 @@
 defmodule CodeCorps.StripeService.StripeConnectExternalAccountService do
-  alias CodeCorps.{Repo, StripeExternalAccount}
+  alias CodeCorps.{Repo, StripeConnectAccount, StripeExternalAccount}
   alias CodeCorps.StripeService.Adapters.StripeExternalAccountAdapter
 
   @api Application.get_env(:code_corps, :stripe)
 
   def create(id_from_stripe, account_id_from_stripe) do
-    with {:ok, %Stripe.ExternalAccount{} = bank_account} <- @api.ExternalAccount.retrieve(id_from_stripe, connect_account: account_id_from_stripe),
-         {:ok, params} <- StripeExternalAccountAdapter.to_params(bank_account)
+    with {:ok, %Stripe.ExternalAccount{} = external_account} <- @api.ExternalAccount.retrieve(id_from_stripe, connect_account: account_id_from_stripe),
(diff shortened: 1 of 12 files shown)


begedin:
@paulsullivanjr In the interest of wrapping this up and launching the feature as soon as possible, I hope you don't mind my taking this PR over. Good work up to this point!

@begedin makes sense. Definitely didn't want to hold you guys up from finishing!

lib/code_corps/stripe_service/stripe_connect_external_account_service.ex
```diff
@@ -1,16 +1,24 @@
...
+         {:ok, %StripeConnectAccount{} = connect_account} <- get_connect_account(account_id_from_stripe),
+         {:ok, params} <- StripeExternalAccountAdapter.to_params(external_account, connect_account.id)
     do
       %StripeExternalAccount{}
       |> StripeExternalAccount.changeset(params)
       |> Repo.insert
     end
   end
+
+  defp get_connect_account(account_id_from_stripe) do
```
begedin, line 18: The reason we do this at the service instead of the adapter level is because we want the service to return a useful error response in case the connect account was not found.

Specifically, the service will return `{:ok, created_external_account}`  if everything is ok, and `{:error, :not_found}` if the associated connect account was not found locally. 

Due to this behavior, the event handling system will mark the event as errored.

We could technically push it further into the adapter, but the code would be more complicated.

begedin:
@JoshSmith Should be good to go.

The relationship is stored on the external account record. My reasoning is that we may end up at one point having a 

`:stripe_connect_account` -> `has_many` -> `:stripe_external_account`

With that in mind, 

`:stripe_connect_account` -> `has_one` -> `:stripe_external_account`

makes more sense at the moment

There is no chance of us having 

`:stripe_external_account` -> `has_many` -> `:stripe_connect_account`

so 

`:stripe_external_account` -> `has_one` -> `:stripe_connect_account`

makes much less sense.

joshsmith:
I think we should also add a `bank_account_bank_name` to this so that we can render that out to the user, as well.

joshsmith (approved):

PR #613, opened 2017-01-04
537 add invoices

### LINKED ISSUE #537: Handle Stripe invoice.payment_succeeded webhook event
# Problem

* We should handle the `invoice.payment_succeeded` event
* We store the invoice in a local table
  
```Elixir
field :amount_due
field :application_fee

field :id_from_stripe
field :charge_id_from_stripe
field :customer_id_from_stripe
field :subscription_id_from_stripe

belongs_to :user # can figure it out through customer
belongs_to :subscription

# any other fields that make sense
```

* we then track the event. tracking should contain a user id

begedin:
# What's in this PR?

Work in progress for #537

TODO:

* tests
* implement adapter mapping
* implement figuring out user from customer in adapter
* implement tracking properties
* write do_track method better

## References
Fixes #537

### DIFF
--- a/lib/code_corps/stripe_service/adapters/stripe_invoice.ex
+++ b/lib/code_corps/stripe_service/adapters/stripe_invoice.ex
@@ -1,0 +1,68 @@
+defmodule CodeCorps.StripeService.Adapters.StripeInvoiceAdapter do
+  alias CodeCorps.{Repo, StripeConnectCustomer, StripeConnectSubscription}
+
+  import CodeCorps.MapUtils, only: [keys_to_string: 1]
+  import CodeCorps.StripeService.Util, only: [transform_map: 2]
+
+  # Mapping of stripe record attributes to locally stored attributes
+  # Format is {:local_key, [:nesting, :of, :stripe, :keys]}
+  @stripe_mapping [
+    {:id_from_stripe, [:id]},
+    {:amount_due, [:amount_due]},
+    {:application_fee, [:application_fee]},
+    {:attempt_count, [:attempt_count]},
+    {:attempted, [:attempted]},
+    {:charge_id_from_stripe, [:charge]},
+    {:closed, [:closed]},
+    {:currency, [:currency]},
+    {:customer_id_from_stripe, [:customer]},
+    {:date, [:date]},
+    {:description, [:description]},
+    {:ending_balance, [:ending_balance]},
+    {:forgiven, [:forgiven]},
+    {:next_payment_attempt, [:next_payment_attempt]},
+    {:paid, [:paid]},
+    {:period_end, [:period_end]},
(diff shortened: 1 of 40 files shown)


joshsmith:
This is a little janky. I would prefer better ways of handling all the related resources, but I couldn't think of any. It works, though.

We also should probably write a better set of property tracking for invoices.

lib/code_corps/stripe_service/adapters/stripe_invoice.ex
```diff
@@ -0,0 +1,69 @@
...
+  alias CodeCorps.{Repo, StripeConnectCustomer, StripeConnectSubscription}
+
+  import CodeCorps.MapUtils, only: [keys_to_string: 1]
+  import CodeCorps.StripeService.Util, only: [transform_map: 2]
+
+  # Mapping of stripe record attributes to locally stored attributes
+  # Format is {:local_key, [:nesting, :of, :stripe, :keys]}
+  @stripe_mapping [
+    {:id_from_stripe, [:id]},
+    {:amount, [:amount]},
```
begedin: I'm not seeing `:amount` in the API docs? Where do we get it from?

lib/code_corps/stripe_service/stripe_invoice_service.ex
```diff
@@ -0,0 +1,28 @@
...
+  alias CodeCorps.{Repo, StripeConnectAccount, StripeConnectCustomer, StripeInvoice}
+  alias CodeCorps.StripeService.Adapters.StripeInvoiceAdapter
+
+  @api Application.get_env(:code_corps, :stripe)
+
+  @spec create(binary, binary) :: {:ok, %StripeInvoice{}} | {:error, %Ecto.Changeset{}}
+  def create(invoice_id_from_stripe, customer_id_from_stripe) do
+    with account_id <- get_connect_account(customer_id_from_stripe),
+      {:ok, %Stripe.Invoice{} = invoice} <- @api.Invoice.retrieve(invoice_id_from_stripe, connect_account: account_id),
+         {:ok, params} <- StripeInvoiceAdapter.to_params(invoice)
```
begedin, line 12: Some weird indentation issue here.

begedin (commented):
Generally looks good, just confused about the `:amount` field.

Also, I can't approve or request changes, since I'm the pull request author 👍

lib/code_corps/stripe_service/adapters/stripe_invoice.ex
joshsmith: You're right. Idk how I got it.

joshsmith: Need to change in stripity.

joshsmith (approved):

PR #616, opened 2017-01-06
Add inbox flag to task lists

joshsmith:
# What's in this PR?

This adds the concept of `inbox` `true | false` to `TaskList` for determining whether a list is the inbox for new tasks in the project.


begedin (approved):
This looks good to me, but just for the sake of discussion, wouldn't something like `:is_default` be easier to comprehend both for us and the users.

The way I see it, many projects might have a task list that's outright called "inbox", us included. `:is_default` seems much more explicit to me, with less potential for confusion.

This is not something I have a strong opinion about, though, so good to merge.

joshsmith:
@begedin There are four default task lists, though.

PR #622, opened 2017-01-07
Set proper order for default task lists of a project

### LINKED ISSUE #592: Default task lists are not ordered
# Problem

In the `Project.create_changeset` we're making use of `put_assoc(:task_lists, TaskList.default_task_lists())` in order to create the default task lists for a project (Inbox, etc).

Unfortunately, since this function does not use the `TaskList.changeset`, then `set_order()` is never called (from `EctoOrdered`). This means the `order` is never set via the `position` attribute used in `default_task_lists()`.

## Subtasks
- [ ] Write a failing regression test to catch the issue
- [ ] Make the test pass by setting the order of the new default task lists when creating a new project

dleve123:
# What's in this PR?

This PR changes `TaskList.default_task_lists/0` to return changesets, which means that the `order` of each `TaskList` is now properly set within the scope of a project via `TaskList.changeset/2`.

## References
Fixes #592

### DIFF
--- a/web/models/task_list.ex
+++ b/web/models/task_list.ex
@@ -16,7 +16,7 @@ defmodule CodeCorps.TaskList do
   end
 
   def default_task_lists() do
-    task_list_data = [
+    [
       %{
         name: "Inbox",
         position: 1
@@ -30,11 +30,7 @@ defmodule CodeCorps.TaskList do
         name: "Done",
         position: 4
       }
-    ]
-
-    for task_list <- task_list_data do
-      TaskList.changeset(%TaskList{}, task_list)
-    end
+    ] |> Enum.map(fn (params) -> changeset(%TaskList{}, params) end)
   end
 
   @doc """
--- a/web/models/task_list.ex
+++ b/web/models/task_list.ex
@@ -2,6 +2,8 @@ defmodule CodeCorps.TaskList do
   use CodeCorps.Web, :model
   import EctoOrdered
 
+  alias CodeCorps.TaskList
+
   schema "task_lists" do
     field :name, :string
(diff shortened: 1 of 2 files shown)


dleve123:
cc: @JoshSmith – I iterated towards what I think is a more robust/idiomatic solution to this bug, but started with the solution you described at https://github.com/code-corps/code-corps-api/pull/622/commits/c66af717835d27d634d1bcfdf29beb7f539a19d3. Let me know what you think of the end solution (d4508b6)

web/models/task_list.ex
```diff
@@ -29,6 +31,10 @@ defmodule CodeCorps.TaskList do
         position: 4
       }
     ]
+
+    for task_list <- task_list_data do
+      TaskList.changeset(%TaskList{}, task_list)
+    end
   end
```
begedin, line 34: Just a thought, but this might be nicely chained using

```Elixir
[%{name: "Index", position: 1}, ...]
|> Enum.map(fn (params) -> changeset(%TaskList{}, params) end)
```

dleve123: @begedin Thanks for the comment! I'm not super familiar with idiomatic Elixir, but seems like a thing is to use a lot of `|>`s, so I'll refactor accordingly and push a commit with that change.

joshsmith (approved):

joshsmith:
Thanks @dleve123! Would you mind squashing to a single commit? Then we can merge.

joshsmith:
@dleve123 just a quick ping here.

joshsmith:
Closing in favor of #635 so we can push this up.

Issue #625, opened 2017-01-12
Verify that no webhooks accept data from the event directly


joshsmith:
# Problem

Per #612 we should ensure that webhooks work idempotently by simply fetching the latest data from Stripe's API.

We'll need to step through each endpoint to verify this is the case.

begedin:
Went through all of our current webhooks handlers and none of them use the data stored in the event at the moment, so this should be safe to close.

PR #627, opened 2017-01-12
Moved code from handler for "account.updated" into StripeConnectAccountService

begedin:
# What's in this PR?

Figured this out while working on #625 

For consistency, I moved the code from the "account.updated" handler into `StripeConnectAccountService.update_from_stripe`.

I also removed some `else` cases from all 3 functions there, since they either just propagate the error response (basically doing nothing), or they outright swallow errors.

We should also add an issue, good for contributors to add a test for `StripeConnectAccountService.update_from_stripe/1`  - created #628

### DIFF
--- a/lib/code_corps/stripe_service/stripe_connect_account.ex
+++ b/lib/code_corps/stripe_service/stripe_connect_account.ex
@@ -4,28 +4,44 @@ defmodule CodeCorps.StripeService.StripeConnectAccountService do
 
   @api Application.get_env(:code_corps, :stripe)
 
+  @doc """
+  Used to create a remote `Stripe.Account` record as well as an associated local
+  `StripeConnectAccount` record.
+  """
   def create(attributes) do
     with {:ok, from_params} <- StripeConnectAccountAdapter.from_params(attributes),
          {:ok, %Stripe.Account{} = account} <- @api.Account.create(from_params),
          {:ok, params} <- StripeConnectAccountAdapter.to_params(account, attributes)
     do
-      %StripeConnectAccount{}
-      |> StripeConnectAccount.create_changeset(params)
-      |> Repo.insert
+      %StripeConnectAccount{} |> StripeConnectAccount.create_changeset(params) |> Repo.insert
     end
   end
 
+  @doc """
+  Used to update both the local `StripeConnectAccount` as well as the remote `Stripe.Account`,
+  using attributes sent by the client
+  """
   def update(%StripeConnectAccount{id_from_stripe: id_from_stripe} = account, %{} = attributes) do
     with {:ok, from_params} <- StripeConnectAccountAdapter.from_params(attributes),
          {:ok, %Stripe.Account{} = stripe_account} <- @api.Account.update(id_from_stripe, from_params),
-         {:ok, params} <- StripeConnectAccountAdapter.to_params(stripe_account, attributes),
-         {:ok, %StripeConnectAccount{} = updated_account} <- account |> StripeConnectAccount.webhook_update_changeset(params) |> Repo.update
+         {:ok, params} <- StripeConnectAccountAdapter.to_params(stripe_account, attributes)
     do
-      {:ok, updated_account}
+      account |> StripeConnectAccount.webhook_update_changeset(params) |> Repo.update
+    end
+  end
+
+  @doc """
+  Used to update the local `StripeConnectAccount` record using data retrieved from the Stripe API
+  """
+  def update_from_stripe(id_from_stripe) do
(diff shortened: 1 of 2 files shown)


lib/code_corps/stripe_service/stripe_connect_account.ex
```diff
@@ -4,28 +4,44 @@ defmodule CodeCorps.StripeService.StripeConnectAccountService do
...
+  def update_from_stripe(id_from_stripe) do
+    with {:ok, %Stripe.Account{} = stripe_account} <- @api.Account.retrieve(id_from_stripe),
+         %StripeConnectAccount{} = local_account <- Repo.get_by(StripeConnectAccount, id_from_stripe: id_from_stripe),
+         {:ok, params} <- stripe_account |> StripeConnectAccountAdapter.to_params(%{})
+    do
+      local_account |> StripeConnectAccount.webhook_update_changeset(params) |> Repo.update
     else
-      {:error, %Ecto.Changeset{} = changeset} -> {:error, changeset}
-      {:error, %Stripe.APIErrorResponse{} = error} -> {:error, error}
-      _ -> {:error, :unhandled}
```
begedin, line 28: This is outright error hiding and should go away, so I removed it.

lib/code_corps/stripe_service/stripe_connect_account.ex
```diff
@@ -4,28 +4,44 @@ defmodule CodeCorps.StripeService.StripeConnectAccountService do
...
+  """
+  def update_from_stripe(id_from_stripe) do
+    with {:ok, %Stripe.Account{} = stripe_account} <- @api.Account.retrieve(id_from_stripe),
+         %StripeConnectAccount{} = local_account <- Repo.get_by(StripeConnectAccount, id_from_stripe: id_from_stripe),
+         {:ok, params} <- stripe_account |> StripeConnectAccountAdapter.to_params(%{})
+    do
+      local_account |> StripeConnectAccount.webhook_update_changeset(params) |> Repo.update
     else
-      {:error, %Ecto.Changeset{} = changeset} -> {:error, changeset}
-      {:error, %Stripe.APIErrorResponse{} = error} -> {:error, error}
```
begedin, line 27: This one serves no function. It passes out the error response. Not having this line leaves behavior exactly the same.

lib/code_corps/stripe_service/stripe_connect_account.ex
```diff
@@ -4,28 +4,44 @@ defmodule CodeCorps.StripeService.StripeConnectAccountService do
...
+  Used to update the local `StripeConnectAccount` record using data retrieved from the Stripe API
+  """
+  def update_from_stripe(id_from_stripe) do
+    with {:ok, %Stripe.Account{} = stripe_account} <- @api.Account.retrieve(id_from_stripe),
+         %StripeConnectAccount{} = local_account <- Repo.get_by(StripeConnectAccount, id_from_stripe: id_from_stripe),
+         {:ok, params} <- stripe_account |> StripeConnectAccountAdapter.to_params(%{})
+    do
+      local_account |> StripeConnectAccount.webhook_update_changeset(params) |> Repo.update
     else
-      {:error, %Ecto.Changeset{} = changeset} -> {:error, changeset}
```
begedin, line 26: Removing this one means the call in the `do` block will just do the same thing automatically.

lib/code_corps/stripe_service/stripe_connect_account.ex
```diff
@@ -4,28 +4,44 @@ defmodule CodeCorps.StripeService.StripeConnectAccountService do
...
+  """
   def update(%StripeConnectAccount{id_from_stripe: id_from_stripe} = account, %{} = attributes) do
     with {:ok, from_params} <- StripeConnectAccountAdapter.from_params(attributes),
          {:ok, %Stripe.Account{} = stripe_account} <- @api.Account.update(id_from_stripe, from_params),
-         {:ok, params} <- StripeConnectAccountAdapter.to_params(stripe_account, attributes),
-         {:ok, %StripeConnectAccount{} = updated_account} <- account |> StripeConnectAccount.webhook_update_changeset(params) |> Repo.update
+         {:ok, params} <- StripeConnectAccountAdapter.to_params(stripe_account, attributes)
     do
-      {:ok, updated_account}
+      account |> StripeConnectAccount.webhook_update_changeset(params) |> Repo.update
```
begedin, line 29: This call will return `{:ok, updated_account}` anyway. No point in first calling it in the `with` block, then just passing it through in the `do` block.

joshsmith (approved):

PR #631, opened 2017-01-15
Project has can-activate-donations on its view

### LINKED ISSUE #630: Expand Validators.ProjectCanEnableDonations to also require payments to be fully setup
# Problem

We should expand the `Validators.ProjectCanEnableDonations` validator module so it requires the stripe connect account to be fully setup for payments (`transfers_enabled` needs to be `true`)

This is part of code-corps/code-corps-ember#947

## References

Progress on code-corps/code-corps-ember#947

joshsmith:
# What's in this PR?

Closes #630.

### DIFF
--- a/lib/code_corps/stripe_service/validators/project_can_enable_donations.ex
+++ b/lib/code_corps/stripe_service/validators/project_can_enable_donations.ex
@@ -13,12 +13,9 @@ defmodule CodeCorps.StripeService.Validators.ProjectCanEnableDonations do
 
   These are:
 
-  * At least one `CodeCorps.DonationGoal`
-
-  and only one of:
-
-  - `Organization` with a `StripeConnectAccount`, in production env, which has `charges_enabled: true`
-  - `Organization` with a `StripeConnectAccount`, not in production env
+  - At least one `CodeCorps.DonationGoal`
+  - `Organization` with a `StripeConnectAccount` which
+    has `charges_enabled: true` and `transfers_enabled: true`
 
   If the project has these relationships set up, it returns `{:ok, project}`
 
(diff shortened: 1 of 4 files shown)


lib/code_corps/stripe_service/validators/project_can_enable_donations.ex
```diff
@@ -17,8 +17,10 @@ defmodule CodeCorps.StripeService.Validators.ProjectCanEnableDonations do
 
   and only one of:
 
-  - `Organization` with a `StripeConnectAccount`, in production env, which has `charges_enabled: true`
-  - `Organization` with a `StripeConnectAccount`, not in production env
+  - `Organization` with a `StripeConnectAccount`, in production env, which
+    has `charges_enabled: true` and `transfers_enabled: true`
+  - `Organization` with a `StripeConnectAccount` and
+    `transfers_enabled: true`, not in production env
```
begedin: I'm not sure what my original reasoning was for this (I was the one originally implementing the validator), but I'm not seeing the benefits in splitting up the environments here.

I think it may have just been for easier manual testing, so I'm considering it might be a good idea to just ditch the check for the correct environment. 

It should just be 
- `Organization` with a `StripeConnectAccount`, which has `charges_enabled: true` and `transfers_enabled: true`

I would remove the method on line 38 and update the documentation.

begedin (changes requested):

lib/code_corps/stripe_service/validators/project_can_enable_donations.ex
joshsmith: @begedin I think it's because `charges_enabled` is always `true` in some environments? Not sure.

begedin (approved):

PR #635, opened 2017-01-15
Set proper order for default task lists of a project

### LINKED ISSUE #592: Default task lists are not ordered

joshsmith:
# What's in this PR?

This PR changes `TaskList.default_task_lists/0` to return changesets, which means that the `order` of each `TaskList` is now properly set within the scope of a project via `TaskList.changeset/2`.

## References
Fixes #592 

Is effectively #622 but trying to rebase/squash.

### DIFF
--- a/test/models/project_test.exs
+++ b/test/models/project_test.exs
@@ -62,6 +62,21 @@ defmodule CodeCorps.ProjectTest do
       changeset = Project.create_changeset(%Project{}, %{organization_id: 1})
       assert {:ok, 1} == changeset |> fetch_change(:organization_id)
     end
+
+    test "associates the ordered default task lists to the project" do
+      organization = insert(:organization)
+      changeset = Project.create_changeset(
+        %Project{},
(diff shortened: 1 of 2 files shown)


dleve123:
@JoshSmith - apologies for the too-late rebase here, and thanks for opening up a new PR. Really busy week this past week and went skiing over the weekend

joshsmith:
@dleve123 no worries at all! Thanks again for the PR. Apologies on my part for closing/opening new PR, but decent way to keep your commit in!

PR #638, opened 2017-01-15
Add error handling to token controller

joshsmith:
# What's in this PR?

Adds some blank username/password handling that was missing.


begedin (approved):
Looks good. Should we add an ember issue to display these errors, or should ember handle it on it's own?

joshsmith:
It already does display them.

Issue #609, opened 2017-01-04
Fix issues with webhooks


joshsmith:
# Problem

We have some webhooks that are failing or are not processed. Here's a breakdown of things based on state.

We'll need to decide whether we address all of these now and in what priority. Some appear to be needed, most do not.

- 👍  looks good
- ❓ urgent question
- ❔ non-urgent question, probably doesn't need handled
- 🚫  needs fixed

### `processed`
account.updated 👍 
account.external_account.created 👍 

### `unhandled`
plan.created - nothing particularly interesting happens here since we've just created the plan
customer.source.created - likely does not need handled since it comes from us creating a card
customer.created - probably does not need handled
charge.succeeded - probably covered by invoice payment succeeded
invoice.created - probably does not need handled
invoice.payment_succeeded - handled right now, I believe
customer.updated - handled as much as we can now 
application_fee.created - probably does not need handled
customer.subscription.created - likely does not need handled

joshsmith:
@begedin Should we have handlers that explicitly ignore events with a reason that they were ignored? Might be nice to have it documented for why, originally, we had ignored them.

begedin:
You're correct in thinking `invoice.payment_succeeded` is handled

PR #639, opened 2017-01-15
Explicitly ignore events listed in #609

### LINKED ISSUE #634: Add handler that explicitly ignores events with a reason that they were ignored
# Problem

We should add a handler that simply sets an event to "ignored" and a reason. Can pattern match. This can serve as in-database documentation for why it was an event type was ignored.

begedin:
# What's in this PR?

This PR first reorganizes the `WebhookProcessor` module so that the environment/livemode filtering goes into its own module. 

Then it separates `process_async` and `process` so that `process can be more easily tested`.

Then it extracts the event handling logic from `WebhookProcessor` into `EventHandler`, so this can be further and more easily tested, while relying less on `StripeTesting` (though there is still some reliance.

Lastly, it hooks into `EventHandler` with a new `IgnoredEventHandler` module, which is in charge of explicitly ignoring some Stripe event types. 

The term _explicitly igoring_ here means
* The `event.status` will be set to _"ignored"_
* The new `event.ignored_reason` field will be set to an informative message based on event type.

## References
Fixes #634


begedin:
Ready for review, I believe. There are 2 commits, but I do not think we should squash, since each of them is a distinct change.

lib/code_corps/stripe_service/events/account_updated.ex
```diff
@@ -1,5 +1,5 @@
 defmodule CodeCorps.StripeService.Events.AccountUpdated do
-  def handle(%{"data" => %{"object" => %{"id" => id_from_stripe}}}) do
+  def handle(%{data: %{object: %{id: id_from_stripe}}}) do
```
begedin, line 2: One of the changes involved passing in the fetched `Stripe.Event` into the handlers, instead of the raw json.

lib/code_corps/stripe_service/webhook_processing/ignored_event_handler.ex
```diff
@@ -0,0 +1,55 @@
...
+  """
+  @spec handle(StripeEvent.t) :: {:ok, StripeEvent.t}
+  def handle(%StripeEvent{type: type} = local_event) do
+    with ignored_reason <- get_reason(type) do
+      local_event |> set_ignored(ignored_reason)
+    end
+  end
+
+  @spec get_reason(String.t) :: String.t
+  defp get_reason("application_fee.created"), do: "Event provides no useful information."
```
joshsmith: Would rewrite to:

> "We don't make use of the application fee."

lib/code_corps/stripe_service/webhook_processing/ignored_event_handler.ex
```diff
@@ -0,0 +1,55 @@
...
+  defp get_reason("customer.created"), do: "Customers are only created from the client. No need to handle."
+  defp get_reason("customer.source.created"), do: "Cards are only created from the client. No need to handle"
+  defp get_reason("customer.subscription.created"), do: "Subscriptions are only created from the client. No need to handle."
+  defp get_reason("invoice.created"), do: "Event provides no useful information."
+  defp get_reason("plan.created"), do: "Plans are only created from the client. No need to handle."
```
joshsmith: Would remove "No need to handle." everywhere.

lib/code_corps/stripe_service/webhook_processing/ignored_event_handler.ex
```diff
@@ -0,0 +1,55 @@
...
+      local_event |> set_ignored(ignored_reason)
+    end
+  end
+
+  @spec get_reason(String.t) :: String.t
+  defp get_reason("application_fee.created"), do: "Event provides no useful information."
+  defp get_reason("customer.created"), do: "Customers are only created from the client. No need to handle."
+  defp get_reason("customer.source.created"), do: "Cards are only created from the client. No need to handle"
+  defp get_reason("customer.subscription.created"), do: "Subscriptions are only created from the client. No need to handle."
+  defp get_reason("invoice.created"), do: "Event provides no useful information."
```
joshsmith: Would rewrite to:

> "We prefer to handle other lifecycle events for invoices, like payment_succeeded."

joshsmith (approved):

PR #643, opened 2017-01-18
Update docs for postgres env vars

joshsmith:
# What's in this PR?

Minor improvement to docs based on feedback.


arubis (approved):
Looks very helpful for the use case I ran into; thanks!

PR #647, opened 2017-01-19
Separate IgnoredEventHandler behavior for connect and platform events.

### LINKED ISSUE #645: Split up ignored event handler to target connect and platform events separately
# Problem

At the moment, we have a list of events we specifically ignore, and a system which handles this list in `CodeCorps.StripeService.WebhookProcessing.IgnoredEventHandler`

This system incorrectly assumes the event type (ex. `account.updated`) is the only thing that effects our handling or ignoring of the event. The second factor we should take into account is if it's a platform or a connect event we are receiving.

This means that we need two ignored event handlers, one for platform events, the other for connect events, with separate lists for each.

Just as a basic example, "customer.updated" is handled for platform events, but should probably be ignored for connect events.

## Subtasks
- [ ] Rewrite the ignored events system so it works separately for connect and platform events
- [ ] Ignore "customer.updated" for connect events, but keep handling it for platform events.
- [ ] All of the currently ignored events should be ignored both on platform and connect webhooks

begedin:
# What's in this PR?

This PR changes behavior of the `IgnoredEventHandler` module to work separately for connect platform events. This allows us, for example, to ignore a certain event type when it hits the platform webhook, while still handling it on the connect webhook, or vice-versa.

It also allows us to have separate ignore reasons for different endpoints.

Finally, it adds an explicit ignore for the "customer.updated" event on the connect webhook. The event is still being handled normally on the platform webhook.

## References
Fixes #645

### DIFF
--- a/lib/code_corps/stripe_service/webhook_processing/ignored_event_handler.ex
+++ b/lib/code_corps/stripe_service/webhook_processing/ignored_event_handler.ex
@@ -1,10 +1,24 @@
 defmodule CodeCorps.StripeService.WebhookProcessing.IgnoredEventHandler do
   alias CodeCorps.{StripeEvent, Repo}
+  alias CodeCorps.StripeService.WebhookProcessing.{
+    ConnectEventHandler, PlatformEventHandler
+  }
 
-  @ignored_event_types [
+  @ignored_platform_event_types [
     "account.external_account.created",
     "application_fee.created",
     "customer.created",
+    "customer.source.created",
+    "customer.subscription.created",
+    "invoice.created",
+    "plan.created"
+  ]
+
+  @ignored_connect_event_types [
+    "account.external_account.created",
+    "application_fee.created",
+    "customer.created",
+    "customer.updated",
     "customer.source.created",
     "customer.subscription.created",
     "invoice.created",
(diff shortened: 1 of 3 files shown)


lib/code_corps/stripe_service/webhook_processing/ignored_event_handler.ex
```diff
@@ -32,21 +47,36 @@ defmodule CodeCorps.StripeService.WebhookProcessing.IgnoredEventHandler do
...
+  defp get_connect_reason("plan.created"), do: "Plans are only created from the client."
+
+  @spec get_platform_reason(String.t) :: String.t
+  defp get_platform_reason("account.external_account.created"), do: "External accounts are stored locally upon updating a connect account."
+  defp get_platform_reason("application_fee.created"), do: "We don't make use of the application fee object."
+  defp get_platform_reason("customer.created"), do: "Customers are only created from the client."
+  defp get_platform_reason("customer.source.created"), do: "Cards are only created from the client. No need to handle"
+  defp get_platform_reason("customer.subscription.created"), do: "Subscriptions are only created from the client."
+  defp get_platform_reason("invoice.created"), do: "We prefer to handle other lifecycle events for invoices, like payment_succeeded."
+  defp get_platform_reason("plan.created"), do: "Plans are only created from the client."
```
begedin, line 78: I'm aware there is some code duplication here and we could reduce it but I feel it's clearer to keep the mapping of messages to event types separate. It also allows us to change the actual text based on the type of webhook that got hit.

joshsmith (approved):

PR #653, opened 2017-01-21
Add stripe charge model and event tracking

### LINKED ISSUE #649: Track "charge.succeeded" event on the connect webhook endpoint
# Problem

We need to track the `charge.succeeded` event in order to track revenue.

## Subtasks
- [ ] Store the event in the database
- [ ] Call segment tracking on it

begedin:
# What's in this PR?

Work in progress

## References
Fixes #649


lib/code_corps/stripe_testing/charge.ex
```diff
@@ -0,0 +1,7 @@
+defmodule CodeCorps.StripeTesting.Charge do
+  import CodeCorps.StripeTesting.Helpers
+
+  def retrieve(_id, _opts) do
+    {:ok, load_fixture(Stripe.Charge, "charge")}
+  end
```
begedin, line 6:
```Elixir
def retrieve(id, _opts) do
  {:ok, load_fixture(Stripe.Charge, id)
end
```

Would allow us to, in the future, specify ID and let stripe "retrieve" a specific fixture. We don't need it right now, but at the same time, the only fixture we try to retrieve right now is "charge".

joshsmith (approved):

PR #658, opened 2017-01-22
Elasticsearch implementation

### LINKED ISSUE #60: Add Elasticsearch Capabilities to Application
Files it is used in:
- [ ] `Skill_controller.ex`

Possible clients to use:
[Tirexs](https://hexdocs.pm/tirexs/api-reference.html)

paulsullivanjr:
# What's in this PR?
This is currently work in progress. But wanted to get an initial PR out. 

## References
Fixes #60


config/dev.exs
```diff
@@ -51,6 +51,10 @@ config :code_corps, :analytics, CodeCorps.Analytics.InMemoryAPI
 config :code_corps, :stripe, Stripe
 config :code_corps, :stripe_env, :dev
 
+# Configure elasticsearch
+config :code_corps, :elasticsearch_url, "http://0.0.0.0:9200"
```
joshsmith: Maybe should set these URLs via ENV vars?

config/dev.exs
```diff
@@ -51,6 +51,10 @@ config :code_corps, :analytics, CodeCorps.Analytics.InMemoryAPI
...
+config :code_corps, :elasticsearch_index,  "skills"
```
joshsmith: Is there a reason we should be setting these indexes in the config?

joshsmith (commented):
A couple comments/questions here. I'm completely ignorant to how elastix works so these are get-my-feet-wet questions.

config/dev.exs
paulsullivanjr: That makes sense. I had put these in the config files initially for convenience.

config/dev.exs
paulsullivanjr: We shouldn't. I had put these in the config files initially for convenience. But it needs to go away.

paulsullivanjr:
Elastix provides basic functionality for working with elasticsearch. It's a pretty small library (4 modules) and not very long or complex.

mix.exs
```diff
@@ -4,7 +4,7 @@ defmodule CodeCorps.Mixfile do
   def project do
     [app: :code_corps,
      version: "0.0.1",
-     elixir: "1.3.4",
+     elixir: "1.4.0",
```
npendery: Looks like the circleci is failing because of this, would undo for the PR as the upgrade to 1.4 should probably be in its own PR

paulsullivanjr: Good point, overlooked that. I'll do that this evening.

web/helpers/elastic_search_helper.ex
```diff
@@ -0,0 +1,76 @@
...
+    #count = response.body["hits"]["total"]
+
+    hits = response.body["hits"]["hits"] || []
+    Enum.map(hits, fn(x) -> x["_source"]["title"] end)
+  end
+
+  def to_map(foo, bar), do: %{ String.to_atom(foo) => bar}
+
+  defp settings_map do
+  %{
```
npendery: Indentation

npendery:
This looks awesome! Great job. One thing I would suggest is maybe making the `field_filter` extensible for other fields. Maybe have a map as an input with `%{field: "type"}` and `field_filter` is generated from this map so if we want to add multiple fields at once we can. Also, not an elasticsearch expert but I believe just by adding more mapping types we could put other models in the same index so maybe have another layer to the input map with type? Doesnt seem like we need it right now but just a thought. This is definitely a good step forward towards a higher level elixir elasticsearch client!

paulsullivanjr:
Thanks for the kind words. I like the ideas you proposed and think they should work. I'll start working on those ideas soon.

joshsmith:
Hey @paulsullivanjr I just wanted to check in on this and see if there's anything we should be doing here?

paulsullivanjr:
@JoshSmith I want to discuss what other steps and pieces we'll need to integrate this. I'll catch up with you on slack tonight or tomorrow. I want to get this done so I can get started other other issues.

joshsmith:
Awesome, I will be around to discuss.

joshsmith:
Hey @paulsullivanjr just adding that suggestions comment here for context to this PR:

1. Adding a skills/search endpoint for searching, handled directly in SkillController
2. Endpoint point needs to return a collection of Skill structures, need to research how to create from elasticsearch results. The Elasticsearch indexes will need to contain the Skill records from the database including id's.
3. Keep an ELASTICSEARCH_URL in .env

web/helpers/elastic_search_helper.ex
```diff
@@ -47,7 +47,7 @@ defmodule CodeCorps.ElasticSearchHelper do
   end
 
   def process_response(%HTTPoison.Response{status_code: 200} = response, type) do
-    response.body["hits"]["hits"] |> Enum.map(fn(x) -> x["_source"][type] end)
+    response.body["hits"]["hits"] |> Enum.map(fn(x) -> x["_source"] end)
```
joshsmith: What is `x` here? I think this variable could benefit from a longer name.

web/controllers/skill_controller.ex
```diff
@@ -18,4 +22,9 @@ defmodule CodeCorps.SkillController do
     |> title_filter(params)
     |> limit_filter(params)
   end
+
+  def search(_conn, params) do
+    CodeCorps.ElasticSearchHelper.search(@elasticsearch_url, @elasticsearch_index, @elasticsearch_type, query)
```
joshsmith: Looks like `query` is undefined here.

web/controllers/skill_controller.ex
```diff
@@ -23,7 +23,7 @@ defmodule CodeCorps.SkillController do
     |> limit_filter(params)
   end
 
-  def search(_conn, params) do
+  def search(_conn, %{"query" => query} = params) do
```
joshsmith: You can remove the ` = params` portion here if you're not using the `params` as a whole object.

joshsmith:
We could easily get `Skill`s back by simply:

- using `results = [%{"id" => 1}, %{"id" => 2}] |> Enum.map(&(&1["id"]))`
- with a where clause: `skills = Skill |> where([s], s.id in [1, 2]) |> Repo.all` where `[1, 2]` are the `results` above

Issue #659, opened 2017-01-23
Send donation receipt emails


begedin:
# Problem

When the user successfully donates, they need to be sent a receipt. My guess is, this should happen when a charge succeeds, meaning they get a receipt every month.

## Subtasks
- [ ] add a “receipt” email to lib/code_corps/emails/receipt  
- [x] replace `config :code_corps, CodeCorps.Mailer, adapter: Bamboo.LocalAdapter` with PostMark adapter - #660 
- [ ] setup postmark account for production and maybe staging? (edited)
- [ ] make postmark template
- [ ] probably remove existing mailer test, replace with test for receipt
- [ ] setup test env to use test adapter
- [ ] make test work
- [ ] dev env can still use local adapter, we need to see if it works with postmark (since postmark does templates), see if “sent_emails” endpoint works
- [ ] either when handling `invoice.payment_succeeded`, or `charge.succeeded`, send the receipt email to the user who performed it. Should probably put that in the event handler itself, not the service which updates the database. Seems more fitting
- [ ] Add test to check if that happens.

joshsmith:
Closed by #661.

PR #661, opened 2017-01-24
Add receipt emails

begedin:
# What's in this PR?

Generates and sends a receipt email when handling the "charge.succeeded" event. The recipient is the user related to the charge customer.

Since the receipt contains information about the project, we get that through the invoice.

Since there is a non-zero chance that the invoice is not stored locally at the time the charge event is being handled, it is instead explicitly retrieved from stripe.

If something goes wrong while generating the email, the event will be marked as `:errored`

## References

Progress on: #659


lib/code_corps/emails/receipt_email.ex
```diff
@@ -0,0 +1,46 @@
...
+    }
+  end
+
+  # TODO: Maybe add invoice relationship to charge and let it
+  # load from there?
+  defp project_title(%StripeConnectCharge{invoice_id_from_stripe: invoice_id_from_stripe}) do
+    invoice =
+      StripeInvoice
+      |> Repo.get_by(id_from_stripe: invoice_id_from_stripe)
+      |> Repo.preload(stripe_connect_subscription: [stripe_connect_plan: :project])
```
joshsmith: It's entirely possible that the invoice event has not yet hit our database. And this only works for subscriptions, since there is no invoice for one-off charges.

Perhaps we should be making use of `metadata` in order to associate these. I see no way we could otherwise associate a charge (to an organization) with a project (where the donation is actually made).

joshsmith:
It's entirely possible that the invoice event has not yet hit our database. And this only works for subscriptions, since there is no invoice for one-off charges.

Perhaps we should be making use of metadata in order to associate these. I see no way we could otherwise associate a charge (to an organization) with a project (where the donation is actually made).

config/prod.exs
```diff
@@ -18,6 +18,8 @@ config :code_corps, CodeCorps.Endpoint,
   force_ssl: [rewrite_on: [:x_forwarded_proto]],
   secret_key_base: System.get_env("SECRET_KEY_BASE")
 
+config :code_corps, System.get_env("SITE_URL")
```
joshsmith: Why did we use this in this case but not the others? Would probably recommend doing codecorps.org here.

begedin: I assumed we might want to set it through the environment. I'm completely ok with setting it here directly, though.

joshsmith (approved):

PR #664, opened 2017-01-26
Simplified subscription creation

### LINKED ISSUE #590: Simplify subscription creation
# Problem

Right now, this is how the creation of a subscription record works:

* We get the quantity and the current user
* We do a `store.createRecord` using quantity and current user.
* We call save, with an `adapterOptions` parameter hash, containing a `projectId` key

```Javascript
subscription.save({ adapterOptions: { projectId: currentProjectId} })
```

The `adapterOptions` hash is handled during serialization and the project id is added as a plain JSON API attribute to the payload.

Then, on the API side, the controller receives the request and passes the parameters into `StripeService.StripeConnectSubscription.find_or_create/2`, which, during a long process, loads the project and the plan for that project, then assigns the plan to the subscription during creation.

My guess is, this was done when we didn't yet have a stripe connect plan relationship with the project on ember side. It's needlessly complicated.

I propose that we do the following:

## In Ember we simply

* fetch the `user`, `quantity` and `stripeConnectPlan`
* create and save a `stripeConnectSubscription` record using those properties
* get rid of the whole `adapterOptions` logic
* get rid of the `projectId` virtual attribute

## In API we

* expect the `find_or_create` request to receive a plan id
* rework project loading to load by associated plan id (we still need the project to validate that it is ready to accept donations)
* everything else remains the same
* clean up subscription controller tests. they do not need the custom payloads anymore

begedin:
Introduced CodeCorps.ViewCase

# What's in this PR?

This PR makes the subscription process slightly simpler and more robust. Initially, I thought I would change the parameters expected by the `find_or_create` function, but ended up going with a different approach.

The PR adds a `has_one :project, through: [:stripe_connect_plan, :project]` relationship to the subscription. This allows us to preload more easily and allows ember to simply consider the subscription to be related to user and project directly, instead of going through the plan. The plan is, this way, not exposed to the regular user through the subscription, only to the owner through the project. 

Requires ember PR to be merged as well.

While implementing this change, I decided to also introduce a `CodeCorps.ViewCase` to be used within our views instead of `CodeCorps.ConnCase`. It uses less imports and should make things slightly simpler.

## References
Dependant on code-corps/code-corps-ember#992
Fixes #590


web/models/stripe_connect_subscription.ex
```diff
@@ -42,6 +42,7 @@ defmodule CodeCorps.StripeConnectSubscription do
     field :status, :string
 
     belongs_to :stripe_connect_plan, CodeCorps.StripeConnectPlan
+    has_one :project, through: [:stripe_connect_plan, :project]
```
joshsmith: Would place this below the `belongs_to`'s.

test/views/stripe_connect_subscription_view_test.exs
```diff
@@ -0,0 +1,37 @@
...
+    subscription = insert(:stripe_connect_subscription, stripe_connect_plan: plan, user: user)
+
+    rendered_json =  render(CodeCorps.StripeConnectSubscriptionView, "show.json-api", data: subscription)
+
+    expected_json = %{
+      "data" => %{
+        "attributes" => %{
+          "quantity" => subscription.quantity,
+          "inserted-at" => subscription.inserted_at,
+          "updated-at" => subscription.updated_at
```
joshsmith, line 17: Alpha ordering.

test/support/view_case.ex
```diff
@@ -0,0 +1,25 @@
+defmodule CodeCorps.ViewCase do
```
joshsmith, line 1: Good thinking here! 👍

lib/code_corps/stripe_service/stripe_connect_subscription_service.ex
```diff
@@ -20,19 +18,22 @@ defmodule CodeCorps.StripeService.StripeConnectSubscriptionService do
...
 
   # Side effects
 
   - If the subscription is created or found, associated project totals will get updated
   - If the subscription is created or found, associated donation goal states will be updated
   """
-  def find_or_create(%{"project_id" => project_id, "quantity" => _, "user_id" => user_id} = attributes) do
-    with {:ok, %Project{} = project} <- get_project(project_id) |> ProjectSubscribable.validate,
-         {:ok, %User{} = user} <- get_user(user_id) |> UserCanSubscribe.validate
+  def find_or_create(%{"project_id" => project_id, "user_id" => user_id, "quantity" => _} = attributes) do
```
joshsmith: Would alpha-order the attrs `%{"project_id" => project_id, "quantity" => _, "user_id" => user_id}`

lib/code_corps/stripe_service/stripe_connect_subscription_service.ex
```diff
@@ -141,10 +142,12 @@ defmodule CodeCorps.StripeService.StripeConnectSubscriptionService do
     }
   end
 
-  defp to_insert_attributes(attrs, %StripeConnectPlan{id: stripe_connect_plan_id}) do
+  defp local_create_attributes(attrs, %StripeConnectPlan{id: stripe_connect_plan_id}) do
```
joshsmith: I had specified `insert` here since we use it for a `Repo.insert` operation.

joshsmith (approved):
Approved though with some caveats and minor requested changes. Does not need re-review to merge unless we have some disagreements here on changes.

begedin:
@JoshSmith I made the changes you requested, but I'll delay merging until the related ember PR is merged as well, since it will cause incompatibility otherwise.

PR #669, opened 2017-01-31
Update apib doc and add swagger doc along with swagger html UI

npendery:
# What's in this PR?

* Add swagger doc to replace apiblueprint
* Add html page to view the docs via [ReDoc](https://github.com/Rebilly/ReDoc)


joshsmith:
@npendery after some talking with some of the Apiary folks, I'm not sure if we want to move away from API Blueprint just yet.

Think we should close this and focus on improving the blueprint?

PR #672, opened 2017-02-01
Add cloudinary uploads

joshsmith:
# What's in this PR?

This adds everything we need for handling cloudinary uploads:

- Migration task for changing the icons over from S3 to Cloudinary
- Adds default colors to be generated when users/projects/organizations are created
- Removes some outdated tests

Still needs some tests. Need to identify which.


lib/code_corps/helpers/cloudinary_url.ex
```diff
@@ -0,0 +1,11 @@
+defmodule CodeCorps.Helpers.CloudinaryUrl do
+
+  @host Application.get_env(:arc, :asset_host)
```
begedin, line 3: Are we still using arc for this, or will it go away? Need to rename the config, if that's the case.

lib/code_corps/helpers/random_icon_color.ex
```diff
@@ -0,0 +1,14 @@
...
+  alias Ecto.Changeset
+
+  @icon_color_generator Application.get_env(:code_corps, :icon_color_generator)
+
+  def generate_icon_color(changeset, icon_color_key) do
+    case changeset do
+      %Changeset{valid?: true, changes: changes} ->
+        Changeset.put_change(changeset, icon_color_key, @icon_color_generator.generate())
+      _ ->
+        changeset
```
begedin, line 11: Is this separation here actually necessary?

Is it costly to generate the icon color, which is why we want to skip it in case of an invalid changeset?

begedin: Either way, I would do a test. Since we're injecting the generator via config, it shouldn't be too problematic.

lib/code_corps/helpers/cloudinary_url.ex
```diff
@@ -0,0 +1,11 @@
...
+
+  def for(nil, _options, version, default_color, type) do
+    "#{@host}/icons/#{type}_default_#{version}_#{default_color}.png"
+  end
+  def for(public_id, options, _version, _default_color, _type) do
+    Cloudex.Url.for(public_id, options)
```
begedin, line 9: We might wanna inject Cloudex via config as well, so we can test `for/5`

begedin (commented):
Some thoughts. Posting this as I see the "This page is out of date." message, so not sure how it's going to stick.

lib/code_corps/helpers/random_icon_color.ex
begedin: Also, not actually seeing the generator module anywhere in the PR. Is everything pushed up?

joshsmith:
Should be able to add a `default_url` config option to the `Cloudex.Url.for()`

lib/code_corps/helpers/cloudinary_url.ex
joshsmith: It will go away. You can comment in #673, too. Can't remove all that until after this is merged into prod.

lib/code_corps/helpers/random_icon_color.ex
joshsmith: This generator has been around for awhile.

begedin (approved):
This generally looks good, but the coverage dropped quite a bit. If we are not adding tests for the new stuff here, we should at least create issues.

Otherwise, good to go.

joshsmith:
@begedin there are new things here, but that's not why coverage dropped, I think. Believe the drop in coverage is due primarily to dropping tests on files that are intended to be removed as part of #673.

begedin:
@JoshSmith 

Unit tests for

```
lib/code_corps/helpers/cloudinary_url.ex
lib/code_corps/helpers/random_icon_color.ex
```

Would still be welcome, I think, so I would add issues for those.

PR #677, opened 2017-02-01
Added dialyzer locally (not on CI)

begedin:
# What's in this PR?

Work in progress, still trying to get it work on CI. I will also have to actually fix reported issues once it does work.

Update: Got the build to fail if there are dialyzer warnings. It also made the build quite long, so need to figure out how to add caching or something.

### DIFF
--- a/lib/code_corps/stripe_service/stripe_connect_subscription_service.ex
+++ b/lib/code_corps/stripe_service/stripe_connect_subscription_service.ex
@@ -1,4 +1,9 @@
 defmodule CodeCorps.StripeService.StripeConnectSubscriptionService do
+  @moduledoc """
+  Used to perform actions on `StripeConnectSubscription` records while propagating
+  to and from associated `Stripe.Subscription` records.
+  """
+
   alias CodeCorps.{
     Project, Repo, StripeConnectCustomer, StripeConnectAccount,
     StripeConnectPlan, StripeConnectSubscription, User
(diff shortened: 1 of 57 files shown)


joshsmith:
In the circle_pre_build you may need to `mix dialyzer.plt`, not positive.

joshsmith:
The exit code `2` means:

> No problems were found during the analysis, but warnings were emitted.

begedin:
I'm OK with the build failing on exit code 2, though, since it does mean that our specs aren't really correct.

joshsmith:
I might suggest running the dialyxir test on a separate line. Should make Circle reporting more transparent.

lib/code_corps/stripe_service/stripe_connect_subscription_service.ex
```diff
@@ -29,6 +34,13 @@ defmodule CodeCorps.StripeService.StripeConnectSubscriptionService do
   - If the subscription is created or found, associated project totals will get updated
   - If the subscription is created or found, associated donation goal states will be updated
   """
+  @spec find_or_create(map) :: {:ok, StripeConnectSubscription.t} |
+                               {:error, Ecto.Changeset.t} |
+                               {:error, Stripe.APIErrorResponse.t} |
+                               {:error, :project_not_found} |
+                               {:error, :project_not_ready} |
+                               {:error, :user_not_found} |
+                               {:error, :user_not_ready}
```
joshsmith, line 33: Maybe remove the doc for the return values since `@spec` does that?

joshsmith (approved):

lib/code_corps/stripe_service/stripe_connect_subscription_service.ex
begedin: I thought I did, definitely did in the other place I added it. Done now.

PR #683, opened 2017-02-02
Refactor segment tracking

begedin:
# What's in this PR?

Refactors segment tracking to work mainly as a plug that hooks in just before sending a response.

For all other cases, we can use the new direct calling approach which is also supported by the refactoring.


begedin:
This should be good for review.

joshsmith (approved):

joshsmith:
Really great improvement here.

PR #694, opened 2017-02-12
Add unit test for CodeCorps.Helpers.CloudinaryUrl and RandomIconColor

YuriiZdyrko:
Added tests for:
lib/code_corps/helpers/cloudinary_url.ex
lib/code_corps/helpers/random_icon_color.ex

Moved `Cloudex` to configuration to allow testing.
## References
Closes #675

PR #695, opened 2017-02-13
Add test support for Cloudex

begedin:
# What's in this PR?

This PR makes it so Cloudex can run in dev and test mode, even if the API keys are missing. Without this, the application will not start in either of the two modes.

I thought we had an issue for this, but I cannot find it. With this and #694, Cloudex should work much better.

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -61,3 +61,12 @@ config :code_corps, CodeCorps.Mailer,
 
 config :code_corps,
   postmark_receipt_template: "123"
+
+# If the dev environment has no CLOUDEX_API_KEY set, we want the app to still run,
+# with cloudex in test API mode
+if System.get_env("CLOUDEX_API_KEY") == nil do
+  config :cloudex, :cloudinary_api, Cloudex.CloudinaryApi.Test
+  config :cloudex, api_key: "test_key", secret: "test_secret", cloud_name: "test_cloud_name"
+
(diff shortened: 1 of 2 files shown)


config/dev.exs
```diff
@@ -61,3 +61,12 @@ config :code_corps, CodeCorps.Mailer,
...
 config :code_corps,
   postmark_receipt_template: "123"
+
+# If the dev environment has no CLOUDEX_API_KEY set, we want the app to still run,
+# with cloudex in test API mode
+if System.get_env("CLOUDEX_API_KEY") == nil do
+  config :cloudex, :cloudinary_api, Cloudex.CloudinaryApi.Test
+  config :cloudex, api_key: "test_key", secret: "test_secret", cloud_name: "test_cloud_name"
+
+  IO.puts("NOTE: No Cloudex configuration found. Cloudex is runnning in test mode.")
```
joshsmith, line 71: Is there a way to do this as a `warn` with a yellow log color.

Also, is this `test_key`, etc an actual Cloudinary value, or just something we're using?

begedin: It's just a random value. We need to provide a value, any value, or Cloudex raises an error on startup.

begedin:
@JoshSmith I replaced `IO.puts` with a `require Logger` and a `Logger.warn` to get the yellow text. Should be good to go now.

joshsmith (approved):

PR #697, opened 2017-02-14
Fix user tasks

joshsmith:
# What's in this PR?

This is a small fix for user tasks.

May need to improve this. See if you think there's anything that can change @begedin (like tests).


begedin (approved):
To me, this seems good to go. It exposes `user_task` on the task model and view, and the associated test passes.

PR #698, opened 2017-02-15
Add user filter by first, last, username

begedin:
# What's in this PR?

Adds the ability to do basic filtering on the user endpoint. The query filters by first name, last name or username and limits to 5 results. I have little to no experience  with elastic search so this felt like a preferable first step, with the intention of pushing it out as quickly as possible.

### DIFF
--- a/lib/code_corps/helpers/query.ex
+++ b/lib/code_corps/helpers/query.ex
@@ -65,6 +65,21 @@ defmodule CodeCorps.Helpers.Query do
 
   # end comment queries
 
+  # user queries
+
+  def user_filter(query, %{"query" => query_string}) do
+    query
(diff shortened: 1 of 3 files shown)


joshsmith:
This is definitely a good minimal first step. But I want to see what progress @paulsullivanjr can make in wrapping up #658 and we will need to circle back to recover this to index a combination of first name/last name and username, then add some fuzzy matching of the strings (like with the skill typeahead) on the user's list items on the front-end side.

Do we want to open some follow-up issues in both the API and Ember?

begedin:
Not sure of any other follow-ups needed, but I've created #700 
Let me know if this needs any updates or if it's good to merge.

PR #701, opened 2017-02-17
Fix cloudex configuration

joshsmith:
# What's in this PR?

This fixes some Cloudex config issues that made it into a merge.

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -62,11 +62,10 @@ config :code_corps, CodeCorps.Mailer,
 config :code_corps,
   postmark_receipt_template: "123"
 
-# If the dev environment has no CLOUDEX_API_KEY set, we want the app to still run,
-# with cloudex in test API mode
+# If the dev environment has no CLOUDEX_API_KEY set, we want the app
+# to still run, with cloudex in test API mode
 if System.get_env("CLOUDEX_API_KEY") == nil do
-  config :cloudex, :cloudinary_api, Cloudex.CloudinaryApi.Test
-  config :cloudex, api_key: "test_key", secret: "test_secret", cloud_name: "test_cloud_name"
-
   IO.puts("NOTE: No Cloudex configuration found. Cloudex is runnning in test mode.")
+  config :code_corps, :cloudex, CloudexTest
+  config :cloudex, api_key: "test_key", secret: "test_secret", cloud_name: "test_cloud_name"
 end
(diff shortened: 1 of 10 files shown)


config/dev.exs
```diff
@@ -62,11 +62,10 @@ config :code_corps, CodeCorps.Mailer,
 config :code_corps,
   postmark_receipt_template: "123"
 
-# If the dev environment has no CLOUDEX_API_KEY set, we want the app to still run,
-# with cloudex in test API mode
+# If the dev environment has no CLOUDEX_API_KEY set, we want the app
+# to still run, with cloudex in test API mode
 if System.get_env("CLOUDEX_API_KEY") == nil do
-  config :cloudex, :cloudinary_api, Cloudex.CloudinaryApi.Test
```
begedin, line 68: This is actually a documented `Cloudex` approach of enabling test mode for the API. Don't have a link right now, but should probably look into the Clodex docs to see what it does. Not sure if we wanna remove it.

config/dev.exs
```diff
@@ -62,11 +62,10 @@ config :code_corps, CodeCorps.Mailer,
...
-  config :cloudex, api_key: "test_key", secret: "test_secret", cloud_name: "test_cloud_name"
-
   IO.puts("NOTE: No Cloudex configuration found. Cloudex is runnning in test mode.")
+  config :code_corps, :cloudex, CloudexTest
```
begedin, line 69: I think this is what fixes it. The removal of the other change probably doesn't break anything right now, but might be misleading.

lib/code_corps/helpers/cloudex_test.ex
```diff
@@ -1,4 +1,4 @@
-defmodule Cloudex.CloudinaryApi.Test do
```
begedin, line 1: This is probably the other part of the fix.

begedin (commented):
Looking from my phone, it's hard to judge, but you might be tackling two separate things here. Look into the cloudex docs to see how they handle testing and see which parts we need to change here and which we don't.

It's possible this is all good, but I'm not sure from my phone.

config/dev.exs
joshsmith: We don't make use of the API, so don't really need it right now.

config/dev.exs
joshsmith: Yeah, I'm going to move this out of `helpers` too and into a dedicated `cloudex`.

begedin (approved):
Considering your comments, looks good to go.

PR #702, opened 2017-02-18
Fix limit query to not error on blank values

joshsmith:
# What's in this PR?

Quick fix of the limit query to not error when values are `""`.

### DIFF
--- a/lib/code_corps/helpers/query.ex
+++ b/lib/code_corps/helpers/query.ex
@@ -17,10 +17,11 @@ defmodule CodeCorps.Helpers.Query do
 
   # skill queries
 
-  def limit_filter(query, %{"limit" => count}) do
-    query |> limit(^count)
-  end
+  def limit_filter(query, %{"limit" => count}), do: query |> add_limit(count |> Integer.parse)
   def limit_filter(query, _), do: query
+
+  defp add_limit(query, {count, _rem}), do: query |> limit(^count)
+  defp add_limit(query, _other), do: query
 
   def title_filter(query, %{"query" => title}) do
     query |> where([object], ilike(object.title, ^"%#{title}%"))


begedin (approved):
This looks good, but I have to wonder. Why is this problematic? Is there a case where we had count which was not an integer?

Either way, makes the query safer, so good to merge, but was wondering.

begedin:
Never mind about the question. The answer was in the description.

lib/code_corps/helpers/query.ex
```diff
@@ -9,7 +9,7 @@ defmodule CodeCorps.Helpers.Query do
 
   # skill queries
 
-  def limit_filter(query, %{"limit" => count}) do
+  def limit_filter(query, %{"limit" => count}) when is_integer(count) do
```
begedin: Seems like query might be serializing this as a string, actually, so tests are failing. Either that, or our tests are sending a string `"5"`, when they should be sending an integer.

begedin (dismissed):
Tests are failing on this.

joshsmith:
I think our tests must be sending it incorrectly somehow. This worked fine in not tests.

begedin:
@JoshSmith I made the tests pass. We can't assume the param is an integer because the phoenix controller will parse it as a string. The tests aren't passing it incorrectly, the controller is receiving it as a string.

My change makes it so it will convert the limit to integer and apply it. If that fails (because the limit is not integer) or if there is no limit, the query will be unaltered.

I'm not sure this is the behavior we want. In the future, we may want to return an error indicating the parameter is invalid, but it ought to be good enough for now.

snewcomer (approved):
This looks good.  Just one question - I see scrivener_ecto installed, but don't see the `paginate` function used anywhere.  This limit filter would be replaced by something like `page_size` I think.  Is there any plan to use scrivener_ecto or is it already being used?

joshsmith:
@snewcomer we removed pagination but it's a good thought to actually do something like pagination here instead of a limit. @begedin your thoughts on this approach?

joshsmith:
We had removed pagination when swapping to task boards but planned to add it back for other things later, so didn't remove the dep.

PR #703, opened 2017-02-18
User being fetched twice during authentication sequence

### LINKED ISSUE #560: User being fetched twice during authentication sequence
# Problem

The user is fetched from the database first during the call to [`plug Guardian.Plug.LoadResource`](https://github.com/code-corps/code-corps-api/blob/develop/web/router.ex#L26) and then it is fetched again during the call to [`plug CodeCorps.Plug.CurrentUser`](https://github.com/code-corps/code-corps-api/blob/develop/web/router.ex#L34).

It may be possible to alleviate the second call by calling `Guardian.Plug.current_resource(conn)` [here](https://github.com/code-corps/code-corps-api/blob/develop/web/plugs/current_user.ex#L12) in order to fetch the user from the conn instead of from the database.

This will probably need some testing to ensure that `Guardian.Plug.current_resource/1` works as expected.

YuriiZdyrko:
# What's in this PR?

- changed plugs/current_user.exs to fetch User from conn
- added unit test

## References
Fixes #560

Progress on: #

### DIFF
--- a/web/plugs/current_user.ex
+++ b/web/plugs/current_user.ex
@@ -1,19 +1,18 @@
 defmodule CodeCorps.Plug.CurrentUser do
-  alias CodeCorps.GuardianSerializer
+  @moduledoc """
+  Puts authenticated Guardian user into conn.assigns[:current_user]
+  """
 
+  @spec init(Keyword.t) :: Keyword.t
   def init(opts), do: opts
 
+  @spec call(Plug.Conn.t, Keyword.t) :: Plug.Conn.t
   def call(conn, _opts) do
-    case Guardian.Plug.current_token(conn) do
+    case Guardian.Plug.current_resource(conn) do
+      user = %CodeCorps.User{} ->
+        Plug.Conn.assign(conn, :current_user, user)
       nil ->
         conn
-      current_token ->
-        with {:ok, claims} <- Guardian.decode_and_verify(current_token),
-             {:ok, user} <- GuardianSerializer.from_token(claims["sub"]) do
-                Plug.Conn.assign(conn, :current_user, user)
-        else
-          {:error, _reason} -> conn
-        end
     end
   end
 end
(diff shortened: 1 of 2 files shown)


web/plugs/current_user.ex
```diff
@@ -1,19 +1,18 @@
...
 
+  @spec call(Plug.Conn.t, Keyword.t) :: Plug.Conn.t
   def call(conn, _opts) do
-    case Guardian.Plug.current_token(conn) do
+    case Guardian.Plug.current_resource(conn) do
+      user = %CodeCorps.User{} ->
+        Plug.Conn.assign(conn, :current_user, user)
       nil ->
         conn
-      current_token ->
```
begedin: I see you removed this completely. Is it your experience that this case does not happen at all? Makes me wonder why we had it in the first place. @JoshSmith Any ideas?

begedin (commented):
Looks really good, IMO and definitely good to go, but would like to hear @JoshSmith thoughts on this, since he was the one who originally implemented it.

Any clue why we initially went with `current_token` and not `current_resource`?

joshsmith:
I'm not exactly sure why this is working vs what I was doing manually. Is the user decoded from the token this way? I need to test against the frontend too I think.

YuriiZdyrko:
Looks like previous implementation duplicated part of LoadResource's work.
https://hexdocs.pm/guardian/Guardian.Plug.LoadResource.html

joshsmith:
Interesting! Can you rebase onto develop and I'll get this merged asap?

joshsmith (approved):

YuriiZdyrko:
Rebased

joshsmith:
@banzay actually the "merge into" button doesn't do a rebase. You'll need to do something in your local branch like `get fetch --all` and `git rebase origin/develop` in order to rebase onto `develop`. Happy to pair on it if you haven't done it, but I think there are some docs in this repo about it.

joshsmith:
@banzay actually figured out that I can do this myself. Learning!

YuriiZdyrko:
👍

PR #707, opened 2017-02-24
WIP Add project acceptance email

joshsmith:
# What's in this PR?

This is the start of the project acceptance email.


begedin:
@JoshSmith I'm not sure why tests are behaving differently for cloudex on CI. Could use your input on this one.

lib/code_corps/emails/project_user_acceptance_email.ex
```diff
@@ -0,0 +1,35 @@
...
+    |> to(user.email)
+    |> template(template_id(), build_model(project, user))
+  end
+
+  defp build_model(%Project{} = project, %User{} = user) do
+    %{
+      project_logo: ImagePresenter.large(project),
+      project_title: project.title,
+      project_url: project |> preload() |> url(),
+      subject: "TODO: Put subject here",
```
begedin: You still need a subject. I've added this field because I believe Postmark is expecting it. If that's wrong, feel free to remove it. In hindsight, I should've not allowed the test to pass here.

web/controllers/project_user_controller.ex
```diff
@@ -27,4 +29,25 @@ defmodule CodeCorps.ProjectUserController do
...
+
+  defp maybe_send_email(%{assigns: %{ changeset: changeset}} = conn) do
+    case {changeset |> Ecto.Changeset.get_change(:role), changeset.data.role} do
+      {"contributor", "pending"} -> send_acceptance_email(conn.assigns.project_user)
+      _ -> nil
+    end
+
+    conn
+  end
+  defp maybe_send_email(conn), do: conn
```
begedin, line 46: This function actually defines behavior which is not exactly basic. To paraphrase it:

If the changeset changes a role that was `pending` to a role that is `contributor`, send acceptance email, otherwise do nothing.

I think that warrants a test and would probably be more easily testable if we put it in a separate module.

Not sure if you wanna do it here or in a separate PR, but we tend to delay these things and I feel we should probably strive to do that less.

begedin (changes requested):

web/controllers/project_user_controller.ex
joshsmith: I think it needs to be done in a separate module. Not much complexity should be happening in the controller itself.

joshsmith:
@begedin the content is mostly done here, but you should attempt to do the separate model work and any other cleanup you think is required.

If you can give me a list of things that need to be done for this to go live (putting template on Postmark, etc), then I can do that first thing tomorrow morning.

begedin:
@JoshSmith I've fixed the existing tests. What's left to do is:

- [ ] Move the email sending code from the project controller into probably an external plug module. We can start there and further expand. It's likely we well be sending emails on any membership role changes in the future, so it's a decent start.
- [ ] Write tests for this plug, if you feel they are necessary. The only test we have right now is that the email is being sent in the ideal case. I'm not sure if there are any edge cases we need to account for.

joshsmith:
@begedin perhaps we can wrap this in a separate PR? As-is it's technically mergeable and we do have higher priorities.

begedin:
I'm fine with creating an issue and wrapping up.

begedin (approved):

joshsmith:
@begedin if you have time to squash and merge and open the follow-up issue, will help. Slow wifi at airport hobbled me here.

PR #719, opened 2017-02-25
Make user an owner of an organization directly

### LINKED ISSUE #709: Make a user an owner of an organization directly
# Problem

We need to add an `owner` belongs to relationship on `Organization` for `User`.

joshsmith:
# What's in this PR?

Adds an `owner` of `User` to the `Organization`.

## References
Fixes #709


test/models/organization_test.exs
```diff
@@ -3,27 +3,43 @@ defmodule CodeCorps.OrganizationTest do
...
+      assert changeset.valid?
+      assert changeset.changes.slug == "code-corps"
+    end
+
+    test "with invalid attributes" do
+      changeset = Organization.create_changeset(%Organization{}, @invalid_attrs)
+      refute changeset.valid?
+    end
+
+    test "ensures owner (user) actuallys exist" do
```
christopherstyles: Small typo here 🤓

christopherstyles (changes requested):
This looks good and is working just fine. The only suggestion I would make is adding API documentation. 

In addition to adding API docs for this update, it also looks like the `+ Attributes` property is missing from the Organization _create_ request block [here](https://github.com/code-corps/code-corps-api/blob/develop/blueprint/api.apib#L518), which results in no visible attributes in the generated documentation.

christopherstyles (approved):

PR #721, opened 2017-02-28
Add a note to the docs about Elixir version management

christopherstyles:
# What's in this PR?

Currently Elixir `1.4.1` is specified in mix dependencies. If a user has a different version installed, an error is displayed when running mix tasks.

This update adds a note to the installation docs about using [Kiex](https://github.com/taylor/kiex) to manage multiple Elixir versions.


joshsmith (approved):

joshsmith:
🙌

PR #741, opened 2017-03-08
Fixes #739 and fixes #740 by adding website to projects

### LINKED ISSUE #739: Add website field to Project
# Problem

The `Project` should have a `website` field added, like exists right now for `User`.

We'll want to pull out `prefix_url` and possibly also `validate_format` into functions in a module or modules that are unit tested and imported into `User` and `Project`.

### LINKED ISSUE #740: Add should_link_externally column to Project
# Problem

We should add a column of `should_link_externally` to the `Project` model. What we render to the client should depend on whether they have a `website`. This should be a `boolean` that's `false` by default.

joshsmith:
# What's in this PR?

Adds `website` and `should_link_externally` columns to `Project` and exposes them on the view.


ignu:
LGTM! :ship: :shipit: 

My only feedback is, as someone new to the project, I'm not sure what `should_link_externally` is for, even after reading the ticket. Is it just opening the project url in a new window?

ignu (approved):

joshsmith:
@ignu it's going to be very very temporary, but good feedback nonetheless.

Issue #498, opened 2016-11-28
User Reset Password


snewcomer:
# We need to add a process to request and subsequently reset the user password

### These two can be done in parallel

- #728 Create `User.reset_password_changeset/2`
- #729 Add an `AuthToken` model 

### Once #728 and #729 are done, these two can also be done in parallel

- #730 Add `POST /forgot-password` to request a password reset email
- #733 to generate the reset link for the ember app as part of #730
- #731 Add a `POST /reset-password` to perform the actual reset of password

### This last one is purely optional 

- #732 Send notification email with option to do another reset when the process is done, as an additional security measure.

begedin:
@snewcomer any update on this?

Here's how I would approach it:

## Subtasks

### These two can be done in parallel

- #728 Create `User.reset_password_changeset/2`
- #729 Add an `AuthToken` model 

### Once #728 and #729 are done, these two can also be done in parallel

- #730 Add `POST /password_reset` to request a password reset email
- #731 Add a `POST /reset_password` to perform the actual reset of password

begedin:
@snewcomer I've updated the original issue description with more detailed instructions on how to do this. If you're continuing work on this, feel free to assign yourself to individual smaller issues to mark your progress. 

This one is not marked as "in progress", so I'm not sure if your being assigned to it is just accidental.

snewcomer:
@begedin Awesome.  Thanks for the description!  Starting to tackle the ember side right now.

PR #742, opened 2017-03-08
[USER]: reset_password_changeset

### LINKED ISSUE #728: Create `User.reset_password_changeset/2`
# Problem

We need to define `reset_password_changeset(struct, params)` on `User`

- casts `:password` and `:password_confirmation`
- validates that the two are equal
- if valid, encrypts the password and puts it as `:encrypted_password` into the changeset
- if invalid, should add an error
- we can make use of `User.put_pass_hash` to achieve parts of this
- we probably need to add a virtual `:password_confirmation` field to the user schema

## References

Progress on: #498

snewcomer:
# What's in this PR?

`reset_password_changeset` method on the user

## References
Fixes #728 

Progress on: #498

### DIFF
--- a/test/models/user_test.exs
+++ b/test/models/user_test.exs
@@ -132,4 +132,16 @@ defmodule CodeCorps.UserTest do
       assert message == "invalid transition yehaww from signed_up"
     end
   end
+
+  test "reset_password_changeset with valid passwords" do
+    changeset = User.reset_password_changeset(%User{}, %{password: "foobar", password_confirmation: "foobar"})
+    assert changeset.valid?
+    assert changeset.changes.encrypted_password
+  end
+
+  test "reset_password_changeset with invalid passwords generates message" do
+    changeset = User.reset_password_changeset(%User{}, %{password: "wat", password_confirmation: "foobar"})
+    refute changeset.valid?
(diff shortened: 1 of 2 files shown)


joshsmith (dismissed):

test/models/user_test.exs
```diff
@@ -132,4 +132,19 @@ defmodule CodeCorps.UserTest do
...
+    assert changeset.changes.encrypted_password
+  end
+
+  test "reset_password_changeset with invalid passwords generates message" do
+    changeset = User.reset_password_changeset(%User{}, %{password: "wat", password_confirmation: "foobar"})
+    refute changeset.valid?
+    [error | _] = changeset.errors
+    {attribute, {message, _}} = error
+    assert attribute == :password_confirmation
+    assert message == "passwords do not match"
```
begedin: You can use our `assert_error_message` from `tests/support/model_case`

```Elixir
assert_error_message(changeset, :password_confirmation, "passwords_do_not_match")
```

If you prefer to be more explicit, which is what' I'm leaning towards lately:

```Elixir
assert error_message(changeset, :password_confirmation) == "passwords do not match"
```

begedin (dismissed):
Just a tiny change and then we're good to go.

joshsmith (approved):

PR #744, opened 2017-03-08
[AUTH TOKEN]: model and test that value was created

### LINKED ISSUE #729: Add an AuthToken model
# We need to add an `AuthToken` model

```Elixir
schema "auth_tokens" do
    field :value, :string
    belongs_to :user, User

    timestamps()
end
```

The changeset for creating it should take a user struct. From that struct, we can use [`Phoenix.Token.sign/4`](https://hexdocs.pm/phoenix/Phoenix.Token.html#sign/4) to generate a token and store it as value.

This provides us with a flexible token model, which we can use to perform password resets, or later, if we want to, passwordless sign-ins

## Subtasks
- define model
- define and test changeset

## References

Progress on: #498

snewcomer:
# What's in this PR?
- [x] Auth token model
- [x] changeset to gen value for value field.

## References
Fixes #729 

Progress on: #498


test/models/auth_token_test.exs
```diff
@@ -0,0 +1,13 @@
...
+
+  alias CodeCorps.AuthToken
+
+  test "changeset with valid attributes" do
+    user = insert(:user)
+    changeset = AuthToken.changeset(%AuthToken{}, user)
+    assert changeset.valid?
+    assert changeset.changes.value
+  end
+
```
joshsmith: Would remove this trailing space.

web/models/auth_token.ex
```diff
@@ -0,0 +1,20 @@
+defmodule CodeCorps.AuthToken do
+  use CodeCorps.Web, :model
+
+  schema "auth_token" do
+    field :value, :string
+    belongs_to :user, CodeCorps.User
```
joshsmith: Would place a line between `field` and the `belongs_to`.

priv/repo/migrations/20170308222552_create_auth_token.exs
```diff
@@ -0,0 +1,14 @@
+defmodule CodeCorps.Repo.Migrations.CreateAuthToken do
+  use Ecto.Migration
+
+  def change do
+    create table(:auth_token) do
```
joshsmith, line 5: Do we mean this to be a password reset token only? If so, should we call it that instead?

Also, I think you want the table name to be plural here.

joshsmith: @snewcomer I see that @begedin intends this to be used for potentially passwordless authentication (i.e. the Slack "magic link") in addition to password reset. I'd just document that in the `@moduledoc` for this, then.

begedin: something like 

```Elixir
@moduledoc """
Represent's one of the user's many possible authentication tokens, created using
`Phoenix.Token.sign/4`. 

Many can coexist and be valid at the same time. They can be used for password resets or passwordless logins. 

They can also expire.
"""
```
Should do the trick.

priv/repo/migrations/20170308222552_create_auth_token.exs
```diff
@@ -0,0 +1,14 @@
...
+      add :value, :string
+      add :user_id, references(:users, on_delete: :nothing)
```
begedin: If I'm reading the documentation [here](https://hexdocs.pm/ecto/Ecto.Migration.html#references/2) correctly, this is basically us telling ecto, "if the referenced user is deleted, leave their tokens and do nothing". Basically, it's the other way around from what you're thinking it is. I would change this to `on_delete: :delete_all`.

It's a bit confusing, but this [stackoverflow answer](http://stackoverflow.com/a/34692705) explains it.

In a migration, `references(:parent, on_delete: :action)` will perform the specified action on associated records in this table, when a record in `:parent` gets deleted.

In a schema, `has_many :children, on_delete: :action` will perform the action on the children when the record belonging to this schema is deleted.

But to put it short, just changing to `on_delete: :delete_all` will do the trick here.

begedin (dismissed):
A minor change on my part, and a couple minor changes requested by @JoshSmith 

Once you get those in, feel free to immediately rebase and squash the commits, so we can approve and merge. 

Good work!

priv/repo/migrations/20170308222552_create_auth_token.exs
joshsmith: Added.

joshsmith (approved):

PR #745, opened 2017-03-09
Update project controller to take website correctly

joshsmith:
# What's in this PR?

Minor fix.

Didn't want to write a regression test since it looks like we may want to move away from multiple changeset types over the long term.


jderr-mx (approved):
Looks good to me

PR #752, opened 2017-03-14
UserTaskMatcher gets tasks with most overlapping skills fixes #736

### LINKED ISSUE #736: Implement basic task matcher module
# Problem

From #668 we need a very basic task matcher module. 

We would require a module, for example, `UserTaskMatcher` with a `&match_user/2`

```Elixir
defmodule CodeCorps.UserTaskMatcher do
  @default_task_count
  def match_user(%CodeCorps.User{} = user, task_count // @default_task_count) do
    get_list_of_tasks_with_most_overlapping_skills(tasks_count) 
  end

  defp get_list_of_tasks_with_most_overlapping_skills(amount) do
    # implementation goes here. probably a simple query initially
    # join user skills and tasks skills, group and count by task, order by count
    # limit the number of results to amount
  end
end
```

## Subtasks

- write described module
- write tests for it

# References

- taken from #668

ignu:
# What's in this PR?

Add UserTaskManager to return tasks with the most overlapping skills

## References
Fixes #736

### DIFF
--- a/lib/code_corps/user_task_matcher.ex
+++ b/lib/code_corps/user_task_matcher.ex
@@ -1,0 +1,25 @@
+defmodule CodeCorps.UserTaskMatcher do
+  @moduledoc """
+  Find the top tasks most matching a User's skills
+  """
+
+  alias CodeCorps.{Repo, Task, User, TaskSkill}
+  import Ecto.Query
+
+  @spec match_user(User.t, number) :: [Task.t]
+  def match_user(%CodeCorps.User{} = user, tasks_count) do
+    query = from t in TaskSkill,
+      join: skill in assoc(t, :skill),
+      join: user_skill in assoc(skill, :user_skills),
+      join: user in assoc(user_skill, :user),
+      where: user.id == ^user.id,
+      group_by: t.task_id,
+      order_by: count(t.task_id),
+      limit: ^tasks_count,
+      select: t.task_id
+
+    matches = query |> Repo.all
+
+    Task |> where([t], t.id in ^matches) |> Repo.all
+  end
+end
--- a/test/lib/code_corps/user_task_matcher_test.exs
+++ b/test/lib/code_corps/user_task_matcher_test.exs
@@ -1,0 +1,32 @@
+defmodule CodeCorps.UserTaskMatcherTest do
+  use CodeCorps.ModelCase
+
+  import CodeCorps.UserTaskMatcher
+
+  test "can find top x tasks for a user's skill" do
+    coding = insert(:skill, title: "coding")
+    design = insert(:skill, title: "design")
+
+    account_page = insert(:task)
+    settings_page = insert(:task)
+    photoshop = insert(:task)
+
+    insert(:task)
+
+    insert(:task_skill, task: account_page, skill: design)
+    insert(:task_skill, task: account_page, skill: coding)
+    insert(:task_skill, task: settings_page, skill: design)
+    insert(:task_skill, task: settings_page, skill: coding)
+    insert(:task_skill, task: photoshop, skill: coding)
+
+    user = insert(:user)
+
+    insert(:user_skill, user: user, skill: design)
+    insert(:user_skill, user: user, skill: coding)
+
+    tasks = match_user(user, 2)
+
+    assert(length(tasks) == 2)
+    assert(length(match_user(user, 3)) == 3)
+  end
+end
--- a/web/models/skill.ex
+++ b/web/models/skill.ex
@@ -12,6 +12,7 @@ defmodule CodeCorps.Skill do
     has_many :roles, through: [:role_skills, :role]
 
     has_many :project_skills, CodeCorps.ProjectSkill
+    has_many :user_skills, CodeCorps.UserSkill
     has_many :projects, through: [:project_skills, :project]
 
     timestamps()


lib/code_corps/user_task_matcher.ex
```diff
@@ -0,0 +1,25 @@
...
+  def match_user(%CodeCorps.User{} = user, tasks_count) do
+    query = from t in TaskSkill,
+      join: skill in assoc(t, :skill),
+      join: user_skill in assoc(skill, :user_skills),
+      join: user in assoc(user_skill, :user),
+      where: user.id == ^user.id,
+      group_by: t.task_id,
+      order_by: count(t.task_id),
+      limit: ^tasks_count,
+      select: t.task_id
```
joshsmith, line 19: Have you been able to run an EXPLAIN on this query by chance to see how well it performs?

lib/code_corps/user_task_matcher.ex
```diff
@@ -0,0 +1,25 @@
+defmodule CodeCorps.UserTaskMatcher do
+  @moduledoc """
+  Find the top tasks most matching a User's skills
+  """
+
+  alias CodeCorps.{Repo, Task, User, TaskSkill}
+  import Ecto.Query
+
+  @spec match_user(User.t, number) :: [Task.t]
+  def match_user(%CodeCorps.User{} = user, tasks_count) do
```
joshsmith, line 10: Would write some inline documentation here and rename to `tasks_limit` or something more intention revealing.

web/models/skill.ex
```diff
@@ -12,6 +12,7 @@ defmodule CodeCorps.Skill do
     has_many :roles, through: [:role_skills, :role]
 
     has_many :project_skills, CodeCorps.ProjectSkill
+    has_many :user_skills, CodeCorps.UserSkill
```
joshsmith, line 15: Can this go last to keep alpha order?

lib/code_corps/user_task_matcher.ex
```diff
@@ -0,0 +1,25 @@
...
+    query = from t in TaskSkill,
+      join: skill in assoc(t, :skill),
+      join: user_skill in assoc(skill, :user_skills),
+      join: user in assoc(user_skill, :user),
+      where: user.id == ^user.id,
+      group_by: t.task_id,
+      order_by: count(t.task_id),
+      limit: ^tasks_count,
```
joshsmith, line 18: I actually kind of wonder if we shouldn't do the `limit` chained elsewhere to keep this function more multipurpose. I.e. Without it we could do a count.

test/lib/code_corps/user_task_matcher_test.exs
```diff
@@ -0,0 +1,32 @@
...
+    insert(:task_skill, task: photoshop, skill: coding)
+
+    user = insert(:user)
+
+    insert(:user_skill, user: user, skill: design)
+    insert(:user_skill, user: user, skill: coding)
+
+    tasks = match_user(user, 2)
+
+    assert(length(tasks) == 2)
```
joshsmith, line 29: You might be able to write this as `assert tasks |> length() == 2` instead.

test/lib/code_corps/user_task_matcher_test.exs
```diff
@@ -0,0 +1,32 @@
...
+
+  import CodeCorps.UserTaskMatcher
+
+  test "can find top x tasks for a user's skill" do
+    coding = insert(:skill, title: "coding")
+    design = insert(:skill, title: "design")
+
+    account_page = insert(:task)
+    settings_page = insert(:task)
+    photoshop = insert(:task)
```
joshsmith, line 12: These could maybe be rethought a little to tell a clearer story. Like one could be `front_end_task` and one `back_end_task` and one `design_task`. You'd have `coding_skill` and `design_skill` and then match accordingly. Your final test would be 1 and 3 for length instead, using a limit of 1 and 3.

joshsmith: This comment probably conflicts slightly with the suggestion to chain, but with the right ordering could be resolvable.

joshsmith (commented):
This looks really great and is going to be a huge deal if we can combine it with an "unassigned tasks" query. That means we'll be able to suggest tasks to a user automatically via email, and eventually through our in-app messenger, too.

joshsmith:
Really great work here btw!

web/models/skill.ex
ignu: should `roles` and `role_skills` be in alpha order too?

test/lib/code_corps/user_task_matcher_test.exs
ignu: That's what i was going for with skills that were implementing an entire page (thus needing `coding` and `design`) versus one that was just PhotoShop.

I did start to wonder sometimes about using just the most overlapping tasks as the sort priority. I could see someone like me selecting 'design' as a tag (even though that's a stretch). Then I'd get a list of results back that were design heavy, even though I'd be better suited for a task that simply overlaps my strongest skill, `coding`.

(Seems like this would be more clear when we saw real data instead of speculating what the data would be.)

ignu:
That's a good question, what use is this if we don't return just unassigned tasks?

web/models/skill.ex
joshsmith: Yes!

test/lib/code_corps/user_task_matcher_test.exs
joshsmith: Yeah I think this is simply the most naive approach right now. I'd prefer some sort of recommender system generally but we're a ways off from that.

joshsmith:
@ignu I can see it being useful in other circumstances, like as a filter for the task board.

joshsmith:
@ignu just wanted to check on this with you.

PR #753, opened 2017-03-14
Reset password route

### LINKED ISSUE #731: Add a POST /password/reset endpoint
# We need to add a `POST /password/reset` endpoint as the last step to reset a user's password

* Takes in `[:value, :password, :password_confirmation]`
* Finds `AuthToken` by `:value`, as well as associated `User`
* Uses [`Phoenix.Token.verify/4`] to verify the token is correct and not expired or anything like that.
* If all is in order, updates the user with new password
  * For this, w need a `User.reset_password_changeset` which casts and validates `:password` and `:password_confirmation`. It should ensure the two match before encrypting and assigning the new password. This can be done as a separate issue in #728
  * Endpoint should render validation errors if the changeset is invalid

## Subtasks
- [x] add endpoint to router
- [x] add `PasswordResetController.reset_password`
- [ ] past this point, we are blocked by #728
- [x] ensure endpoint renders 200 OK if reset passes
- [ ] ensure endpoint renders 422 if reset fails due to password mismatch
- [ ] add controller tests

## References

Progress on: #498

snewcomer:
# What's in this PR?

password reset controller finding AuthToken.

@begedin should I also change the AuthToken model `value` to `token`?

## References
Fixes #731

Progress on: #498

### DIFF
--- a/priv/repo/structure.sql
+++ b/priv/repo/structure.sql
@@ -2,11 +2,12 @@
 -- PostgreSQL database dump
 --
 
--- Dumped from database version 9.5.1
--- Dumped by pg_dump version 9.5.1
+-- Dumped from database version 9.6.1
+-- Dumped by pg_dump version 9.6.1
 
 SET statement_timeout = 0;
 SET lock_timeout = 0;
+SET idle_in_transaction_session_timeout = 0;
 SET client_encoding = 'UTF8';
 SET standard_conforming_strings = on;
 SET check_function_bodies = false;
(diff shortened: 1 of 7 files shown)


snewcomer:
Wondering if we can just look up the token or should look up the token/user in the AuthToken model.

joshsmith:
This is looking pretty good. Where did tests go down?

begedin (approved):
Looks good, @snewcomer 

There is no need to change the property name in the model itself. My suggestion was just about having the public API interface be clear to the outside client. The model is internal to the API and should be clearer to the dev, so `auth_token.value` will make more sense that way. This is good to go.

begedin:
@JoshSmith Looks like tests going down is coveralls being coveralls. It reports them going down in files untouched by this PR.

@snewcomer You'll have to rebase and rerun migrations to deal with the `structure.sql` conflict, but this is good to merge after.

snewcomer:
Going to squash and merge!

joshsmith:
🎉  yay!

PR #758, opened 2017-03-15
Update ja_resource to 0.3

### LINKED ISSUE #757: Update ja_resource and check if controller warnings are eliminated
# Problem

`ja_resource` relies on a `__using__` macro to decorate our controllers. With the recent elixir upgrade, this started triggering a lot of warnings in our controllers.

There is a recent release which should eliminate a lot of these warnings, so we should update and reopen the issue in their repo if we still have trouble.

## References

https://github.com/vt-elixir/ja_resource/issues/49

begedin:
# What's in this PR?

The update eliminates all of the controller warnings we had, except for one, which was caused by us and I also fixed here.

## References
Fixes #757


joshsmith (approved):

joshsmith:
Can't change label on phone but looks good to merge.

PR #772, opened 2017-03-19
Add forgot_password controller action

### LINKED ISSUE #730: Add a POST /password/forgot endpoint
# We need to Add a `POST /password/forgot` endpoint

* Requires an email parameter
* Looks for a user with the specified email
* Generates an `AuthToken` model defined in #729
* Generates an email with a link which **takes the user to the ember app route to reset password** - this could be difficult, there's a separate issue #733
* Sends that email to the user
* Endpoint should return a 200 either way, even if the submitted email was wrong and the password reset link was not actually sent.

## Subtasks
- [x] add endpoint to router
- [x] add `PasswordController` with `forgot_password` function
- [x] ensure endpoint returns a 200 OK, either way
- [ ] add behavior described above to endpoint - this part is partially blocked by #729, but the rest can be done
- [x] add controller tests for endpoint

## References

Progress on: #498

snewcomer:
# What's in this PR?
- [x] add forgot pwd endpoint
- [x] controller tests
- [x] Add documentation
- [ ] create forgot_password email

## References
Fixes #730 

Progress on: #498

### DIFF
--- a/blueprint/api.apib
+++ b/blueprint/api.apib
@@ -481,6 +481,25 @@ Until the Code Corps platform is open to new organizations, only admin users can
 
     + Attributes (Unprocessable Entity Response)
 
+# Group Password
+
+This endpoint is a parent route for reset and forgot password endpoints
+
+## Forgot [/forgot]
+
+### Forgot Password [POST]
+
++ Request
+
+    + Attributes (Forgot Password Request)
+
+    + Headers
+
+            Accept: application/vnd.api+json
+            Authorization: Bearer <token>
+
++ Response 200 (application/vnd.api+json; charset=utf-8)
+
 # Group Previews
 
 Preview resources allow a user to create a temporary HTML preview for any kind of markdown.
@@ -2623,6 +2642,16 @@ The platform stores Stripe customers and cards so they can be reused across diff
 
 ## Organizations Response (object)
 + data(array[Organization Resource])
++ include JSON API Version
+
+## Forgot Password Resource (object)
++ include Forgot Password Resource Identifier
+
+## Forgot Password Resource Identifier
++ email: `myemail@gmail.com` (string, required)
+
+## Forgot Password Request (object)
++ data(Forgot Password Resource)
 + include JSON API Version
 
 ## Project User Attributes (object)
--- a/lib/code_corps/emails/forgot_password_email.ex
+++ b/lib/code_corps/emails/forgot_password_email.ex
@@ -1,0 +1,20 @@
+defmodule CodeCorps.Emails.ForgotPasswordEmail do
+  import Bamboo.Email
+  import Bamboo.PostmarkHelper
+
+  alias CodeCorps.Emails.BaseEmail
+
+  def create(user, token) do
+    BaseEmail.create
+    |> to(user.email)
+    |> template(template_id(), [link: link(token)])
+  end
+
+  defp template_id, do: Application.get_env(:code_corps, :postmark_forgot_password_template)
+
+  defp link(token) do
+    Application.get_env(:code_corps, :site_url)
+    |> URI.merge("password/reset?token=#{token}")
+    |> URI.to_string
+  end
+end
--- a/web/controllers/password_controller.ex
+++ b/web/controllers/password_controller.ex
@@ -1,0 +1,14 @@
+defmodule CodeCorps.PasswordController do
+  use CodeCorps.Web, :controller
+
+  alias CodeCorps.{Services.ForgotPasswordService}
+
+  @doc"""
+  forgot_password should take an email and generate an AuthToken model and send an email
+  """
+  def forgot_password(conn, %{"email" => email}) do
+    ForgotPasswordService.forgot_password(email) 
+    conn |> put_status(:ok) |> render("show.json", email: email)
+  end
+
+end
--- a/web/controllers/password_reset_controller.ex
+++ b/web/controllers/password_reset_controller.ex
@@ -11,8 +11,7 @@ defmodule CodeCorps.PasswordResetController do
   and return email. 422 if pwd do not match or auth token does not exist
   """
   def reset_password(conn, %{"token" => token, "password" => password, "password_confirmation" => password_confirmation}) do
-    user = conn.assigns.current_user
-    with %AuthToken{value: auth_token} <- Repo.get_by(CodeCorps.AuthToken, %{ value: token, user_id: user.id }),
+    with %AuthToken{value: auth_token, user: user} <- Repo.get_by(AuthToken, %{ value: token }) |> Repo.preload(:user),
       {:ok, _} <- Phoenix.Token.verify(CodeCorps.Endpoint, "user", auth_token, max_age: 1209600) do
         with %Changeset{valid?: true} <- User.reset_password_changeset(user, 
                                                                        %{password: password, password_confirmation: password_confirmation}) do
--- a/web/views/password_view.ex
+++ b/web/views/password_view.ex
@@ -1,0 +1,10 @@
+defmodule CodeCorps.PasswordView do
+  use CodeCorps.Web, :view
+
+  def render("show.json", %{email: email}) do
+    %{
+      email: email
+    }
+  end
+
+end
--- a/lib/code_corps/services/forgot_password.ex
+++ b/lib/code_corps/services/forgot_password.ex
@@ -1,0 +1,18 @@
+defmodule CodeCorps.Services.ForgotPasswordService do
+
+  alias CodeCorps.{AuthToken, Emails, Mailer, Repo, User}
+
+  @doc"""
+  forgot_password should take an email and generate an AuthToken model and send an email
+  """
+  def forgot_password(email) do
+    with %User{} = user <- Repo.get_by(User, email: email),
+        { :ok, %AuthToken{} = %{ value: token } } <- AuthToken.changeset(%AuthToken{}, user) |> Repo.insert
+    do
+      Emails.ForgotPasswordEmail.create(user, token) |> Mailer.deliver_now()
+    else
+      nil -> nil
+    end
+  end
+
+end
--- a/test/controllers/password_controller_test.exs
+++ b/test/controllers/password_controller_test.exs
@@ -1,0 +1,32 @@
+defmodule CodeCorps.PasswordControllerTest do
+  @moduledoc false
+
+  use CodeCorps.ApiCase, resource_name: :password
+  use Bamboo.Test
+
+  alias CodeCorps.AuthToken
+
+  test "creates and renders resource when email is valid", %{conn: conn} do
+    user = insert(:user)
+    attrs = %{"email" => user.email}
+    conn = post conn, password_path(conn, :forgot_password), attrs
+    response = json_response(conn, 200)
+
+    assert response == %{ "email" => user.email }
+
+    %AuthToken{value: token} = Repo.get_by(AuthToken, user_id: user.id)
+    assert_delivered_email CodeCorps.Emails.ForgotPasswordEmail.create(user, token)
+  end
+
+  test "does not create resource and renders 200 when email is invalid", %{conn: conn} do
+    user = insert(:user)
+    attrs = %{"email" => "random_email@gmail.com"}
+    conn = post conn, password_path(conn, :forgot_password), attrs
+    response = json_response(conn, 200)
+
+    assert response == %{ "email" => "random_email@gmail.com" }
+
+    refute_delivered_email CodeCorps.Emails.ForgotPasswordEmail.create(user, nil)
+  end
+
+end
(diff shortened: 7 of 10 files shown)


lib/code_corps/emails/forgot_password_email.ex
```diff
@@ -0,0 +1,14 @@
...
+
+  alias CodeCorps.Emails.BaseEmail
+
+  def create(user) do
+    BaseEmail.create
+    |> to(user.email)
+    |> template(template_id(), {})
+  end
+
+  defp template_id, do: Application.get_env(:code_corps, :postmark_receipt_template)
```
snewcomer: @JoshSmith @begedin this isn't right.  Wasn't sure what to do with the configuration for this template_id.  Just put random numbers in the config files?

web/controllers/password_controller.ex
```diff
@@ -0,0 +1,37 @@
...
+
+        try_create_forgot_password(user)
+        |> send_forgot_password_email
+
+        conn
+        |> put_status(200)
+        |> render("show.json", email: user.email)
+
+    else
+      nil ->
```
snewcomer: @begedin pretty sure I can just get rid of this `with/else` block if always send 200.  Thoughts?

lib/code_corps/emails/forgot_password_email.ex
begedin: Make it a `:postmark_forgot_password_template`. 

We still need to add the actual template, though, but that's not something you can create. Just assume it's there, under the above configuration key. 

@JoshSmith, you created the receipt template, so I'm referring to you to actually create the template with that ID and add it to the environment.

lib/code_corps/emails/forgot_password_email.ex
```diff
@@ -0,0 +1,14 @@
+defmodule CodeCorps.Emails.ForgotPasswordEmail do
+  import Bamboo.Email
+  import Bamboo.PostmarkHelper
+
+  alias CodeCorps.Emails.BaseEmail
+
+  def create(user) do
```
begedin: The email will have to render the link and the link will contain the token value, so we'll need the token here as well.

web/controllers/password_controller.ex
begedin: `with` would actually still be helpful here

```Elixir
with %User{} = user <- Repo.get_by(User, email: email),
     {:ok, %AuthToken{} = token} <- AuthToken.create_changeset(%AuthToken{}, user) |> Repo.insert()
do
  Emails.ForgotPasswordEmail.create(user, token) |> Mailer.deliver_now()
else
  nil -> nil
end

conn |> put_status(:ok) |> render("show.json", email: email)
```

So you render out the same email the user typed in. The with clause is helpful so that you are able to create and send the email if the address the user provided is valid. The response happens anyway, outside the with clause.

lib/code_corps/emails/forgot_password_email.ex
```diff
@@ -0,0 +1,14 @@
...
+    BaseEmail.create
+    |> to(user.email)
+    |> template(template_id(), {})
```
begedin: The last parameter here is the substitution data model for the template. The template will have to render a reset password link the client will handle. This means we need to generate the link. This is something we don't really know how it will look until we get to client side, so for now, I would go with the placeholder approach.

```Elixir
|> template(template_id(), [link: link(token)])

# ...

defp link(%AuthToken{value: token}) do
  "#{token}" # TODO: we can replace this with the proper string later
end
```

begedin (changes requested):
I hope my comments helped. Still needs some work, but part of it is something either myself or @JoshSmith  will have to do, so our goal here is to get it to a point where we can merge without that part.

web/controllers/password_controller.ex
```diff
@@ -2,36 +2,20 @@ defmodule CodeCorps.PasswordController do
...
-        %Changeset{valid?: true} <- AuthToken.changeset(%AuthToken{}, user) do
-
-        try_create_forgot_password(user)
-        |> send_forgot_password_email
-
-        conn
-        |> put_status(200)
-        |> render("show.json", email: user.email)
-
+        { :ok, %AuthToken{} = %{ value: token } } <- AuthToken.changeset(%AuthToken{}, user) |> Repo.insert
```
snewcomer: @begedin should I be calling Repo.insert here like this?  Or incorporate that apart of the model?

snewcomer (commented):
Also, should `password/reset` be passed through the `ensure_auth` plug?  I see two scenarios.  You forgot your password, so you get a link to reset your password.  The other is you change your password when you are logged in.

web/controllers/password_controller.ex
```diff
@@ -0,0 +1,21 @@
...
+  forgot_password should take an email and generate an AuthToken model and send an email
+  """
+  def forgot_password(conn, %{"email" => email}) do
+    with %User{} = user <- Repo.get_by(User, email: email),
+        { :ok, %AuthToken{} = %{ value: token } } <- AuthToken.changeset(%AuthToken{}, user) |> Repo.insert
+    do
+      Emails.ForgotPasswordEmail.create(user, token) |> Mailer.deliver_now()
+    else
+      nil -> nil
+    end
```
joshsmith: I would split the `with` functionality off into something in `lib`, some kind of a service or similar. Seeing Chris McCord talk about this, he doesn't appear to think that complex behavior should lie in the controller, nor really in the model.

web/controllers/password_controller.ex
joshsmith: See my comment below @snewcomer.

snewcomer: updated

snewcomer:
@begedin @JoshSmith any updates here with the forgot password email template (part that you guys need to do?

joshsmith:
@snewcomer so we need to still write an HTML email for this, right?

What content are we able to put into this email so I can sit down to write it?

lib/code_corps/emails/forgot_password_email.ex
```diff
@@ -0,0 +1,18 @@
...
+  def create(user, token) do
+    BaseEmail.create
+    |> to(user.email)
+    |> template(template_id(), [link: link(token)])
+  end
+
+  defp template_id, do: Application.get_env(:code_corps, :postmark_forgot_password_template)
+
+  defp link(token) do
+    "#{token}" # TODO: we can replace this with the proper string later
```
joshsmith: I'm guessing the URL here will take the `token` and use it in the `/reset` route on the client as a query param?

web/views/password_view.ex
```diff
@@ -0,0 +1,10 @@
+defmodule CodeCorps.PasswordView do
+  use CodeCorps.Web, :view
+
+  def render("show.json", %{email: email}) do
+    %{
+      email: email
+    }
+  end
```
joshsmith, line 8: We could use a view test here and definitely assert that the token is never being rendered.

web/controllers/password_reset_controller.ex
```diff
@@ -12,7 +12,7 @@ defmodule CodeCorps.PasswordResetController do
   """
   def reset_password(conn, %{"token" => token, "password" => password, "password_confirmation" => password_confirmation}) do
     user = conn.assigns.current_user
-    with %AuthToken{value: auth_token} <- Repo.get_by(CodeCorps.AuthToken, %{ value: token, user_id: user.id }),
+    with %AuthToken{value: auth_token} <- Repo.get_by(AuthToken, %{ value: token, user_id: user.id }),
```
joshsmith: I think this should probably `get_by` the `token` only, since we don't necessarily have the `user` here. It won't be in the `assigns` in this case.

lib/code_corps/emails/forgot_password_email.ex
begedin: @JoshSmith That's the general idea. Not sure if we want to go with the config way of doing this (set this route as a config value) or something else.

web/controllers/password_reset_controller.ex
snewcomer: Fixed https://github.com/code-corps/code-corps-api/pull/772/commits/602f46e9685fc208423a063d1c2ae945acc7e3c6

web/views/password_view.ex
snewcomer: Fixed https://github.com/code-corps/code-corps-api/pull/772/commits/602f46e9685fc208423a063d1c2ae945acc7e3c6

joshsmith:
@begedin can you comment on what's left to do here?

begedin:
# TODOs

- [x] In `emails/forgot_password_email.ex`, change `link/1` to

```Elixir
defp link(token) do
  Application.get_env(:code_corps, :site_url)
  |> URI.merge("password/reset?token=#{token}")
  |> URI.to_string
end
```

That should be good enough for now. We are not 100% sure that's what the client link to the reset route is going to be, but it's mergeable and we can easily adjust later.

- [x] Add a test to ensure the link is properly encoded. Basically, ensure that the result of calling `create/2` is a struct that contains a proper URL for the test environment. We can just add these assertions to the existing create test, or we can make a separate test. 

That's about all I can think off.

begedin:
@JoshSmith Once @snewcomer is done, you or myself will have to create a postmark template for the email and then add information to the config so that the following function:

```Elixir
defp template_id, do: Application.get_env(:code_corps, :postmark_forgot_password_template)
```

can work.

snewcomer:
@begedin Done.  @JoshSmith How long does it take to wip up the forgot password email?

blueprint/api.apib
```diff
@@ -481,6 +481,45 @@ Until the Code Corps platform is open to new organizations, only admin users can
...
++ Request
+
+    + Headers
+
+            Accept: application/vnd.api+json
+            Authorization: Bearer <token>
+
++ Response 201 (application/vnd.api+json; charset=utf-8)
+
+    + Attributes (Reset Password Response)
```
begedin: This goes into the request. The response itself should be something else, probably just blank content. I'd say, leave this part out until the PR that adds the controller action, unless you mean to add the action in this PR as well.

blueprint/api.apib
```diff
@@ -481,6 +481,45 @@ Until the Code Corps platform is open to new organizations, only admin users can
...
++ Request
+
+    + Headers
+
+            Accept: application/vnd.api+json
+            Authorization: Bearer <token>
+
++ Response 200 (application/vnd.api+json; charset=utf-8)
+
+    + Attributes (Forgot Password Response)
```
begedin: The attributes are part of the request as well, no?

begedin (changes requested):

blueprint/api.apib
snewcomer: @begedin which action are you referring to?

begedin: I'm sorry, my bad. I forgot the controller action was actually already implemented. That being said, not that I had another look at it, it seems like it might need fixing. The implementation seems strange to me.

- It accepts a token, a password and a password confirmation.
- It verifies the token
- It creates a changeset.
- If the changeset is valid, it renders the user email, but **doesn't really alter the user**
- If the changeset is not valid, it renders a 422

I'll create an issue to correct this.

I would say, leave the Reset Password documentation out of this PR and just keep the Forgot Password in. I'll post the issue here when I create it.

begedin: Created #795, #796

begedin: To conclude, leave the reset_password api docs out of this PR. The action needs some minor fixing, so we should add the docs separately as we fix that.

lib/code_corps/emails/forgot_password_email.ex
```diff
@@ -0,0 +1,20 @@
...
+    BaseEmail.create
+    |> to(user.email)
+    |> template(template_id(), [link: link(token)])
+  end
+
+  defp template_id, do: Application.get_env(:code_corps, :postmark_forgot_password_template)
+
+  defp link(token) do
+    Application.get_env(:code_corps, :site_url)
+    |> URI.merge("password/reset?token=#{token}")
```
snewcomer, line 17: @begedin In order for the model hook on the ember side to consume this token, we could do something like `this.route('reset', { path: '/reset/:token_id'})` and get it in the params of the model hook.  What do you think about this change?  This could change this to `URI.merge("password/reset/#{token}")`

begedin: I added a comment to your ember PR. You can easily do it without dynamic URL segments. 

That being said, I'm not sure which is better. I feel the query param approach, which is this one here, is more commonly seen with reset tokens. However, I can't argue that the dynamic URL approach, i.e. the one you're proposing, doesn't look cleaner.

snewcomer:
Squashed and ready to merge if we think it is g2g.

joshsmith:
🎉 

Thanks so much again for this @snewcomer. Many apologies it took so long to get merged. You are, as always, a trooper on this.

PR #776, opened 2017-03-27
Add sign up context

joshsmith:
# What's in this PR?

This adds a `sign_up_context` to `User` that allows us to track how a user signed up.

Also adds skippable state transitions.


lib/code_corps/transition/user_state.ex
```diff
@@ -1,8 +1,24 @@
...
   def next("selected_categories", "select_roles"), do: {:ok, "selected_roles"}
+  def next("selected_categories", "skip_roles"), do: {:ok, "skipped_roles"}
+  def next("skipped_categories", "select_roles"), do: {:ok, "selected_roles"}
+  def next("skipped_categories", "skip_roles"), do: {:ok, "skipped_roles"}
+
+  # Select/skip skills
   def next("selected_roles", "select_skills"), do: {:ok, "selected_skills"}
+  def next("selected_roles", "skip_skills"), do: {:ok, "selected_skills"}
+  def next("skipped_roles", "select_skills"), do: {:ok, "selected_skills"}
+  def next("skipped_roles", "skip_skills"), do: {:ok, "selected_skills"}
```
begedin: I wrote tests that are failing here, due to expecting a "skip_skills" transition to return a `"skipped_skills"`. Basically, I assumed this was an oversight. 

If I was mistaken and we really want the same final state in all 4 cases here, that state should, in my opinion, something not-misleading, such as `"onboarded"`.

begedin (dismissed):

begedin (approved):

PR #778, opened 2017-03-29
Update docs with ecto.create command

joshsmith:
# What's in this PR?

Updates docs to be more specific about setting up PostgreSQL and the `mix` tasks needed, with some troubleshooting pointers.


terrancebryant (approved):
This step helped me get and running

PR #782, opened 2017-03-31
Upgrade to Phoenix 1.3

joshsmith:
# What's in this PR?

This is a WIP branch to update to Phoenix 1.3.


begedin:
Working on getting this done

begedin:
@JoshSmith I've made the entire switch and all tests are now passing, so I would call this good to merge.

However, this will have to be an incremental shift, since some concepts are changing with 1.3

Not sure if you've watched this: https://www.youtube.com/watch?v=tMO28ar0lW8

A good amount of useful information there.

joshsmith:
Can you clarify this @begedin:

> this will have to be an incremental shift

begedin:
> Can you clarify this @begedin:
this will have to be an incremental shift

Firstly, this is an RC, so there may still be changes.

Secondly,

It's not just about moving `web` into `lib`. 

Phoenix is trying to abandon the concept of models, some parts don't make sense to stay even in lib/web now, etc., etc. Moving web into lib and making the tests pass is just the tiny first step of upgrading. I'm assuming we will gradually introduce the new concepts and switch the old code to them.

There are a lot of things that are likely going to happen with the upgrade, so this PR is just literally changing a package version and moving some code around. Moving the code part wasn't even strictly necessary, since 1.3 is backwards compatible, but it was a good idea to do it now in one swoop rather than later.

begedin:
Basically, what I'm saying is, this is, I think, in a mergeable state, but

1. I'm not sure we should do it for an RC
2. It's just the initial step.

snewcomer:
I'm not sure what the plan is here but looks like 1.3 is out.  Lmk if you need any help converting to the new structure.

PR #785, opened 2017-04-19
Add Policy Tests

### LINKED ISSUE #734: Write tests for lib/code_corps/helpers/policy.ex
# Problem

This is a helper that we use often, but is completely untested at the moment. It's tested as a side effect of testing individual policies, but we should still test it directly.

## Subtasks
- [ ] Write tests for all public functions of lib/code_corps/helpers/policy.ex

As a good pointer, look at the `@spec` statements and `@doc` blocks for each public function. Tests should make sure all cases mentioned in those two are covered.

securingsincity:
# What's in this PR?
Adds tests for Policy Helpers

## References
Fixes #734 

Progress on: #

### DIFF
--- a/test/lib/code_corps/helpers/policy_test.exs
+++ b/test/lib/code_corps/helpers/policy_test.exs
@@ -1,0 +1,223 @@
+defmodule CodeCorps.Helpers.PolicyTest do
+  use CodeCorps.ModelCase
+  alias Ecto.Changeset
+  alias CodeCorps.{
+    Organization, User, Helpers.Policy,
+    ProjectUser
+  }
+
+  def create_project_user_with_role(role) do
+    user = insert(:user)
+    project = insert(:project)
+    insert(:project_user, project: project, user: user, role: role)
+    {project, user}
+  end
+
+  describe "owned_by/2" do
+    test "returns false when organization is not owned by user" do
+      refute Policy.owned_by?(%Organization{owner_id: 1}, %User{id: 2})
+    end
+
+    test "returns false when invalid arguments are passed" do
+      refute Policy.owned_by?(nil, 2)
+    end
+
+    test "returns false if a project is not owned by the user" do
+      project = insert(:project)
+      some_other_user = %User{id: 1}
+      refute Policy.owned_by?(project, some_other_user)
+    end
+
+    test "returns true if a project is owned by the user" do
+      {project, user} = create_project_user_with_role("owner")
+      assert Policy.owned_by?(project, user)
+    end
+
+    test "returns false if a project is admined by the user" do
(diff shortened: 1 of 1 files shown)


securingsincity:
Still a WIP but getting the PR started

securingsincity:
This should be all set.

joshsmith:
@securingsincity thanks! I'll take a look at this tomorrow morning.

test/lib/code_corps/helpers/policy_test.exs
```diff
@@ -0,0 +1,221 @@
...
+
+  describe "owned_by/2" do
+    test "returns false when Organization owner_id doesn't match User user_id" do
+      refute Policy.owned_by?(%Organization{owner_id: 1}, %User{id: 2})
+    end
+
+    test "returns false when invalid arguments are passed" do
+      refute Policy.owned_by?(nil, 2)
+    end
+    test "returns false if a project is not owned by the user" do
```
begedin: Add a space here.

test/lib/code_corps/helpers/policy_test.exs
```diff
@@ -0,0 +1,221 @@
...
+
+  def create_project_user_with_role(role) do
+    user = insert(:user)
+    project = insert(:project)
+    insert(:project_user, project: project, user: user, role: role)
+    {project, user}
+  end
+
+  describe "owned_by/2" do
+    test "returns false when Organization owner_id doesn't match User user_id" do
```
begedin: I would reword this one and the other one for organization to 

`"returns false when organization is not owned by user"`
`"returns true when organization is owned by user"`

We want the tests to talk about "what we want the app to do" not "what's implemented".

test/lib/code_corps/helpers/policy_test.exs
```diff
@@ -0,0 +1,221 @@
...
+    test "should return task of a Changeset" do
+      task = insert(:task)
+      changeset = %Changeset{changes: %{task_id: task.id}}
+      result = Policy.get_task(changeset)
+      assert result.id == task.id
+    end
+  end
+
+  describe "task_authored_by?/1" do
+    test "should return true if the Task user_id and User id match" do
```
begedin: Again, probably word these as 

`"returns true if the user is the author of the task"`
`"returns false if the user is not the author of the task"`

begedin (changes requested):
Sorry for the late review. There is just some minor wording we would like to change. The code itself looks great and will be ready to merge as soon as you're done.

begedin (approved):
Looks great. Rebase and merge when ready. 👍

securingsincity:
@begedin should be squashed now. Thanks.

Issue #789, opened 2017-04-24
Add CodeCorps.Github module with connect/2 function


begedin:
# Problem

#788  defines a UserController action which calls `CodeCorps.Github.connect/2`.

This function should do the following:

## Handle the auth token part of the [web application flow](https://developer.github.com/v3/oauth/#2-github-redirects-back-to-your-site) for github authentication.

The first step is to make a post request to github to get the token. There is the https://github.com/edgurgel/tentacat API wrapper for elixir, which we will very likely end up using. 

However, this wrapper has no "friendly" endpoint to just generate a token using a code. Instead, we'd have to use a [POST request](https://hexdocs.pm/tentacat/Tentacat.html#post/4) created manually, something like:

```Elixir
post(@token_url, %{
  client_id: @client_id,
  client_secret: @client_secret,
  code: code,
  accept: :json
})
```

The `code` would be provided as a function argument, while all the other parameters are set in the environment.

This call will return an `{:ok, response_struct}` or an `{:error, error_struct}`. We should handle the response and have the private function making the call return the access token. We should also handle a generalized error response by returning a general `{:error, error_struct}`. We can expand this to handle specific errors at a later time, in separate tasks.

## Retrieve a GitHub user

We can use the auth_token to [generate a `Tentakit.Client`](https://hexdocs.pm/tentacat/Tentacat.Client.html#new/1) and use this client to [retrieve user information](https://hexdocs.pm/tentacat/Tentacat.Users.html#me/1). We need a github id for the user, in order to associate it with our own user account.

## Associate GitHub user with current user

_This is blocked until #791 is done. To work on the remainder of this PR, use a virtual `github_id` field on the `User` model and simply have the `GitHub.associate` function set that field on the user struct._

We associate the github user information provided in the previous step with the github user in this step.

## Return the updated current user 

The previous step should give us an updated `CodeCorps.User` struct freshly retrieved from the database. We should return this as `{:ok, user}` from the service.

## Testing

Since this will be making web requests to the github API, we should mock the API. We can use [dependency injection](http://openmymind.net/Dependency-Injection-In-Elixir/) to do this, or if that seems to complicated, we can fall back to using VCR. Someone will be available to help with this once a partial PR is submitted. The core team is still figuring this thing out just as everyone else, so no need to be apprehensive about asking for help here, or anywhere else.

There are examples of us already using dependency injection in how we use `stripity_stripe`, for example.

## Recap

- service receives `current_user, code`
- posts code to github
[...]

PR #792, opened 2017-04-24
791 - Associate user with github account

### LINKED ISSUE #791: Add ability to associate user with github account
# Problem

In order to connect a user account, we need to be able to associate a codecorps user with github user information.

## Solution

We need to write a function which handles this. It can be defined under `CodeCorps.GitHub.associate/2` where the first parameter is a `CodeCorps.User`, while the second is a `Tentacat.response`, or more simply, a map, containing an `id`, which is a github user id, among other things.

To perform the actual association, the simplest initial approach is to have a `github_id` field on the user. We also need a changeset which performs the association.

## Subtasks
- [ ] add `github_id` string field to user
- [ ] add `github_associate_changeset` to user, which casts and requires this field
- [ ] add test for the changeset
- [ ] add `CodeCorps.GitHub.associate/2`, which retrieves a user and a params map, calls the changeset, then updates user and returns the outcome (`{:ok, user}`, or `{:error, changeset}`)
- [ ] add tests for the function, one for each outcome.

securingsincity:
# What's in this PR?
* Adds a new `github_id` field on the user.
* Adds a new GitHub association changeset for the user
* Adds a new GitHub module that performs associations to a user. 

My only concern is the `Github` module. I'm not sure if that's where that file should live or if it should be in the services folder. Would love some feedback @begedin @JoshSmith 

## References
Fixes #791

Progress on: Milestone 11

### DIFF
--- a/lib/code_corps/github.ex
+++ b/lib/code_corps/github.ex
@@ -1,0 +1,10 @@
+defmodule CodeCorps.Github do
+
+  alias CodeCorps.{User, Repo}
+
+  def associate(user, params) do
+    user
+    |> User.github_associate_changeset(params)
+    |> Repo.update()
+  end
+end
--- a/priv/repo/migrations/20170424215355_add-github-id-to-user.exs
+++ b/priv/repo/migrations/20170424215355_add-github-id-to-user.exs
@@ -1,0 +1,9 @@
+defmodule :"Elixir.CodeCorps.Repo.Migrations.Add-github-id-to-user" do
+  use Ecto.Migration
+
+  def change do
+    alter table(:users) do
+      add :github_id, :string
+    end
+  end
+end
--- a/test/lib/code_corps/github_test.exs
+++ b/test/lib/code_corps/github_test.exs
@@ -1,0 +1,20 @@
+defmodule CodeCorps.GithubTest do
+  use CodeCorps.ModelCase
+  alias CodeCorps.Github
+
+  describe "associate/2" do
+    test "should update the user with the github_id" do
+      user = insert(:user)
+      params = %{github_id: "foobar"}
+      {:ok, result} = Github.associate(user, params)
+      assert result.github_id == "foobar"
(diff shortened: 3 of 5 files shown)


lib/code_corps/github.ex
```diff
@@ -0,0 +1,9 @@
+defmodule CodeCorps.Github do
+
+  alias CodeCorps.{User, Repo}
+
+  def associate(user, params) do
+    changeset = User.github_associate_changeset(user, params)
+    Repo.update(changeset)
```
snewcomer: could pipe changeset into Repo.update if you want.

priv/repo/migrations/20170424215355_add-github-id-to-user.exs
```diff
@@ -0,0 +1,9 @@
+defmodule :"Elixir.CodeCorps.Repo.Migrations.Add-github-id-to-user" do
+  use Ecto.Migration
+
+  def change do
+    alter table(:users) do
+      add :github_id, :string, default: ""
```
snewcomer: is the default: "" needed?

snewcomer (commented):
Looks great!  Just wondering how that associate function will be called?  Separate route?

priv/repo/migrations/20170424215355_add-github-id-to-user.exs
npendery: it also looks like GitHub stores these as integers, so we should do the same

securingsincity: @npendery i looked into this and I was running into an issue with the guardian serializer serializing a null integer. I'm not sure what the issue is. I'll look into

begedin (approved):
Looks great and absolutely good to go. Great work!

begedin:
@snewcomer The function will be called from the function defined in #789 initially.

PR #793, opened 2017-04-25
GitHub connect user controller action

### LINKED ISSUE #788: Add route to connect with GitHub
# Problem

The client will use the [web application flow](https://developer.github.com/v3/oauth/#web-application-flow) to connect with github. This provides the client with a `code` parameter, which it posts to the server.

We should add a `POST /github_connect` route and an action in the `UserController` which receives the code, calls a function to handle the API side of the flow, then signs in (to refresh) and renders the current user.

The module which gets called, if not yet implemented as part of another issue, can be temporarily defined under `/lib/codecorps/github.ex`. If not yet implemented, the function should be temporarily written as:

```elixir

defmodule CodeCorps.GitHub do
  @doc """ 
  Temporary function until the actual behavior is implemented. 
  """
  def connect(user, _code), do: user
end
```

The github connect action requires authentication.

It calls `Guardian.Plug.current_resource` to retrieve the current user, then sends it, alongside the `code` parameter into the `CodeCorps.GitHub.connect` function. It receives back an `{:ok, user}` which can just be sent outwards, to be handled by `ja_resource`.

## Subtasks
- [ ] Add the `POST /github_connect` route
- [ ] Add the `UserController.github_connect/2` action. It 
- [ ] Add a user controller test for this action.

npendery:
# What's in this PR?

Adds the `/github_connect` endpoint to hit the currently temp `connect` function under the Github lib file

- [x] Add the `POST /github-connect` route
- [x] Add the `UserController.github_connect/2` action. It 
- [x] Add a user controller test for this action.
- [x] Update documentation

## References
Fixes #788 

Progress on: Milestone 11

### DIFF
--- a/web/controllers/user_controller.ex
+++ b/web/controllers/user_controller.ex
@@ -6,6 +6,7 @@ defmodule CodeCorps.UserController do
 
   alias CodeCorps.User
   alias CodeCorps.Services.UserService
+  alias CodeCorps.Github
 
   plug :load_and_authorize_resource, model: User, only: [:update]
   plug JaResource
@@ -31,6 +32,23 @@ defmodule CodeCorps.UserController do
       {:ok, user}
     else
       {:error, changeset} -> {:error, changeset}
+    end
+  end
+
+  def github_connect(conn, code) do
+    current_user = Guardian.Plug.current_resource(conn)
+    with {:ok, user} <- Github.connect(current_user, code)
+    do
+      conn |> render(:show, data: user)
+    else
+      {:error, %Ecto.Changeset{} = changeset} ->
+        conn
+        |> put_status(:unprocessable_entity)
+        |> render(CodeCorps.ChangesetView, "error.json-api", changeset: changeset)
+      {:error, _error} ->
+        conn
+        |> put_status(:internal_server_error)
+        |> render(CodeCorps.ErrorView, "500.json-api")
     end
   end
 
--- a/web/router.ex
+++ b/web/router.ex
@@ -63,6 +63,7 @@ defmodule CodeCorps.Router do
     resources "/categories", CategoryController, only: [:create, :update]
     resources "/comments", CommentController, only: [:create, :update]
     resources "/donation-goals", DonationGoalController, only: [:create, :update, :delete]
+    post "/github-connect", UserController, :github_connect
     resources "/organizations", OrganizationController, only: [:create, :update]
     post "/password/reset", PasswordResetController, :reset_password
     resources "/previews", PreviewController, only: [:create]
--- a/blueprint/api.apib
+++ b/blueprint/api.apib
@@ -2265,6 +2265,32 @@ This endpoint handles the User resources for Code Corps.
 
     + Attributes (Record Not Found Response)
 
+### Connect user to GitHub [POST /github_connect]
+
+This endpoint allows a user to connect their CodeCorps account to their respective GitHub account.
+
++ Request
+
+    + Attributes
+
+      + code (string) - Client generated code for authenication when sending request to GitHub.
+
+    + Headers
+
+            Accept: application/vnd.api+json
+
++ Response 201 (application/vnd.api+json; charset=utf-8)
+
+    + Attributes (User Response)
+
++ Response 401 (application/vnd.api+json; charset=utf-8)
+
+    + Attributes (JSON Web Token Invalid Response)
+
++ Response 422 (application/vnd.api+json; charset=utf-8)
+
+    + Attributes (Unprocessable Entity Response)
+
 ### Check email availability [GET /users/email_available?email={email}]
 
 This endpoint allows you to check whether an email is valid (by checking simply for presence of the `@` symbol) and whether the email is not already registered.
--- a/lib/code_corps/github.ex
+++ b/lib/code_corps/github.ex
@@ -2,6 +2,11 @@ defmodule CodeCorps.Github do
 
   alias CodeCorps.{User, Repo}
 
+  @doc """
+  Temporary function until the actual behavior is implemented.
+  """
+  def connect(user, _code), do: {:ok, user}
+
   def associate(user, params) do
     user
     |> User.github_associate_changeset(params)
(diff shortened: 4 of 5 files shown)


web/controllers/user_controller.ex
```diff
@@ -34,6 +35,21 @@ defmodule CodeCorps.UserController do
     end
   end
 
+  def github_connect(conn, code) do
+    current_user = Guardian.Plug.current_resource(conn)
+    with {:ok, user} <- Github.connect(current_user, code)
+    do
+      # {:ok, user}
```
npendery: @begedin I tried doing what you suggested with just sending back the `{:ok, user}` and having `ja_resource` handle it, but was running into trouble with `ja_resource` recognizing the custom action. If you have a suggestion on how to remedy this that would be great!

begedin: @npendery I'm afraid I assumed too much about ja_resource there. You'll have to do a `conn |> render` manually as you would in a ja_resourceless controller.

npendery: Cool, thanks for the feedback! Definitely good to hear I wasn't going crazy trying to figure out how to do it :)

web/controllers/user_controller.ex
```diff
@@ -34,6 +35,21 @@ defmodule CodeCorps.UserController do
...
+      conn |> render(:show, data: user)
+    else
+      {:error, _error} ->
+        # {:error, error}
+        conn
+        |> put_status(404)
+        |> render(CodeCorps.ErrorView, "404.json-api")
+    end
```
begedin: The else block here would be something like

```elixir
else
  {:error, %Ecto.Changeset{} = changeset} -> render_validation_server_error
  {:error, _error} -> render_internal_server_error
```

There should be examples for an internal server error somewhere. Try searching for "500". Alternatively, we could leave that case out, in which case, the `with` will raise an unmatched clause error, which will also cause the api to render a 500.

begedin (changes requested):

web/controllers/user_controller.ex
```diff
@@ -34,6 +35,23 @@ defmodule CodeCorps.UserController do
     end
   end
 
+  def github_connect(conn, code) do
+    current_user = Guardian.Plug.current_resource(conn)
+    with {:ok, user} <- Github.connect(current_user, code)
```
snewcomer, line 40: @npendery should the connect function return the `access_token` from Github?

begedin: `Github.connect` ought to return the updated user, which has the `github_auth_token` field, so that's what happens.

web/controllers/user_controller.ex
```diff
@@ -34,6 +35,23 @@ defmodule CodeCorps.UserController do
...
+    do
+      conn |> render(:show, data: user)
+    else
+      {:error, %Ecto.Changeset{} = changeset} ->
+         conn
```
begedin: Minor indentation issue here.

web/router.ex
```diff
@@ -63,6 +63,7 @@ defmodule CodeCorps.Router do
     resources "/categories", CategoryController, only: [:create, :update]
     resources "/comments", CommentController, only: [:create, :update]
     resources "/donation-goals", DonationGoalController, only: [:create, :update, :delete]
+    post "/github_connect", UserController, :github_connect
```
begedin: This is on me, because I'm the one that wrote it in the issue but could you change the path to `/github-connect`,  so it's consistent with other parts of the app. I keep mixing up my dashes and underscores.

begedin (changes requested):
Couple of minor changes, but after that, it looks good to go.

web/controllers/user_controller.ex
snewcomer: @begedin So I was referring to the documentation [here](https://developer.github.com/v3/oauth/#2-github-redirects-back-to-your-site).

I think I may be missing something though.

web/router.ex
npendery: ah, yeah definitely

web/controllers/user_controller.ex
npendery: @snewcomer from my perspective we are doing the following (matching GH's step numbers on that doc you linked to):
1. Ember side we are making the first request to GH |> sending the received code to this `github_connect` function.
2. Once we get the code on the Phoenix side, we find the logged in user |> send the user and the code received to the `Github.connect` function |> that function sends the request to GH to authorize the code |> an access token is returned |> we set that to the user's `github_access_token` field |> send back to the `github_connect` function with the updated user or error.
3. When the user makes updates that will update on GH as well, we use that access token to hit GH's API

Now after saying that, Im wondering if it is better (conventional really) to update the `Repo` in the `Github.connect` function or in this `github_connect` function? ¯\ _ (ツ) _ /¯

Also @begedin, think we should send the `state` as well to `github_connect`?

begedin: >Now after saying that, Im wondering if it is better (conventional really) to update the Repo in the Github.connect function or in this github_connect function? ¯\ _ (ツ) _ /¯

We already have `Github.associate`, which does the associating part, so either `Github.connect` can call `Github.associate` internally, or we can call both in order. I'm fine with either way.

>Also @begedin, think we should send the state as well to github_connect?

The `state` part is dealt with in the first step. I don't think the api should have to deal with it. Basically:

Client
- redirects user to github with `state`
- user accepts connection
- github redirects back to client with `code` and the same `state`
- client checks the original and received state are the same, if they are, posts code to `/github-connect`

That being said, using that approach means we'd have to store the original state clientside somehow, maybe cookie or local storage, in order to check they match. Alternatively, we'd have to do something like

Client
- asks API to generate `state` and assigns to current user
- redirects user to github with `state`
- user accepts connection
- github redirects back to client with `code` and the same `state`
- client posts `code` and `state` to `/github-connect`
- API checks state, continues with the regular process

In this case, we'd need an extra endpoint to generate and assign the `state` server-side. It's probably safer, but also more complex.

npendery:
@begedin is this PR good to merge?

begedin (approved):
This one is ready to go. Good job 👍

Issue #786, opened 2017-04-19
Accept GitHub webhooks


npendery:
# Need endpoints to accept the incoming GitHub webhooks

Partially blocked until we get the initial GitHub integration set up, but can be outlined in a PR for now.

Webhooks will be set up for when a user interacts with an Issue or Issue comment in GitHub. This should modify the appropriate task in the Code Corps project.

[Link to GitHub webhook event page](https://developer.github.com/webhooks/#events)

## Subtasks
- [x] Figure out mapping for Issue -> Task

## Webhook Endpoints for
- [x] Create Issue
- [x] Update Issue
- [x] Delete Issue
- [ ] Create IssueComment
- [ ] Update IssueComment
- [ ] Delete IssueComment

begedin:
This is a good deal in progress and will be wrapped up by #850

PR #794, opened 2017-04-26
INITIAL SETUP Add github issue webhook

npendery:
# What's in this PR?

Adding webhooks to connect to GitHub

## References
Progress on: #786

### DIFF
--- a/lib/code_corps_web/router.ex
+++ b/lib/code_corps_web/router.ex
@@ -59,6 +59,8 @@ defmodule CodeCorpsWeb.Router do
 
     post "/webhooks/stripe/connect", StripeConnectEventsController, :create
     post "/webhooks/stripe/platform", StripePlatformEventsController, :create
+    post "/webhooks/github_issues", GithubIssueController, :create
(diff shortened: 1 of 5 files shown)


web/router.ex
```diff
@@ -55,6 +55,8 @@ defmodule CodeCorps.Router do
 
     post "/webhooks/stripe/connect", StripeConnectEventsController, :create
     post "/webhooks/stripe/platform", StripePlatformEventsController, :create
+    post "/webhooks/github_issues", GithubIssueController, :create
+    post "/webhooks/github_issue_comments", GithubIssueCommentController, :create
```
begedin: How do you feel about calling these actions `:handle`. I feel `:create` is a bit misleading here.

npendery: Yeah I think that makes sense 👍

begedin:
@npendery We did a lot of stuff since this was submitted, and we ended up implementing basic handling for several issue webhooks as part of a different PR, as we were figuring out the major stuff.

I'm sorry about that, but your work here is still appreciated. Simply put, we needed some webhook to figure out the process and this one was relatively straightforward compared to others.

Issue #755, opened 2017-03-15
Document /password/reset route


begedin:
# Problem

#753 introduces a `/password/reset` endpoint. It should be added to the API docs.
 
## References

Related to #753 
[Dealing with API documentation](https://github.com/code-corps/code-corps-api/blob/develop/docs/API.md)

PR #797, opened 2017-04-30
[RESET]: send token back after successful change of pwd close #796 close

### LINKED ISSUE #795: Fix reset_password controller action
# Problem

The behavior of the `PasswordResetController#reset_password` is buggy.

I'm sorry, my bad. I forgot the controller action was actually already implemented. That being said, not that I had another look at it, it seems like it might need fixing. The implementation seems strange to me.

- It accepts a token, a password and a password confirmation.
- It verifies the token
- It creates a changeset.
- If the changeset is valid, it renders the user email, but **doesn't really alter the user**
- If the changeset is not valid, it renders a 422

This should be corrected to

- It accepts a token, a password and a password confirmation.
- It verifies the token
- It creates a changeset, **passes it into `Repo.update`**
- If the result is an `{:ok, user}`, render the user
- If the result is an `{:error, changeset}`, render validation errors

## Subtasks

- [ ] Make the changes
- [ ] Update the tests

### LINKED ISSUE #796: Add API docs for reset_password action
# Problem

The PasswordResetController#reset_password action needs API blueprint documentation.

It accepts a hash containing `token`, `password` and `password_confirmation`
It responds with a user JSON API response, or with a validation error.

## Subtasks
- [ ] Add the documentation

snewcomer:
## References
Fixes #795 #796 #755 

Progress on: #498

### DIFF
--- a/blueprint/api.apib
+++ b/blueprint/api.apib
@@ -500,6 +500,28 @@ This endpoint is a parent route for reset and forgot password endpoints
 
 + Response 200 (application/vnd.api+json; charset=utf-8)
 
+## Reset [/password/reset]
+
+Reset Password resource allow a user to reset their password
+
+### Reset Password [POST]
+
++ Request
+
+    + Attributes (Reset Password Request)
+
++ Response 201 (application/vnd.api+json; charset=utf-8)
+
+    + Attributes (JSON Web Token Valid Response)
+
++ Response 401 (application/vnd.api+json; charset=utf-8)
+
+    + Attributes (JSON Web Token No User With Email Response)
+
++ Response 401 (application/vnd.api+json; charset=utf-8)
+
+    + Attributes (JSON Web Token Invalid Password Response)
+
 # Group Previews
 
 Preview resources allow a user to create a temporary HTML preview for any kind of markdown.
@@ -2849,6 +2871,15 @@ The platform stores Stripe customers and cards so they can be reused across diff
     + id: `NOT_FOUND` (string)
     + title: `404 Resource not found` (string)
     + status: 404 (number) - HTTP status code
+
+## Reset Password Resource (object)
++ token: `zzz123` (string, required)
++ password: `mypwd` (string, required)
++ password-confirmation: `mypwd` (string, required)
+
+## Reset Password Request (object)
++ data(Reset Password Resource)
++ include JSON API Version
 
 ## Role Skill Resource (object)
 + include Role Skill Resource Identifier
--- a/test/controllers/password_reset_controller_test.exs
+++ b/test/controllers/password_reset_controller_test.exs
@@ -2,31 +2,34 @@ defmodule CodeCorps.PasswordResetControllerTest do
   @moduledoc false
 
   use CodeCorps.ApiCase, resource_name: :password_reset
-  alias CodeCorps.AuthToken
+  alias CodeCorps.{User, AuthToken}
 
-  @tag :authenticated
-  test "creates and renders resource when data is valid", %{conn: conn, current_user: current_user} do
+  test "updates user password when data is valid", %{conn: conn} do
+    current_user = insert(:user)
     {:ok, auth_token} = AuthToken.changeset(%AuthToken{}, current_user) |> Repo.insert
     attrs = %{"token" => auth_token.value, "password" => "123456", "password_confirmation" => "123456"}
     conn = post conn, password_reset_path(conn, :reset_password), attrs
     response = json_response(conn, 201)
     assert response
+    encrypted_password = Repo.get(User, current_user.id).encrypted_password
+    assert Comeonin.Bcrypt.checkpw("123456", encrypted_password)
   end
 
-  @tag :authenticated
-  test "does not create resource and renders errors when password does not match", %{conn: conn, current_user: current_user} do
+  test "does not create resource and renders errors when password does not match", %{conn: conn} do
+    current_user = insert(:user)
     {:ok, auth_token} = AuthToken.changeset(%AuthToken{}, current_user) |> Repo.insert
     attrs = %{"token" => auth_token.value, "password" => "123456", "password_confirmation" => "another"}
     conn = post conn, password_reset_path(conn, :reset_password), attrs
-    assert json_response(conn, 422)
+    response = json_response(conn, 422)
+    assert %{"errors" => [%{"detail" => "Password confirmation passwords do not match"}]} = response
   end
 
-  @tag :authenticated
-  test "does not create resource and renders errors when token is invalid", %{conn: conn, current_user: current_user} do
+  test "does not create resource and renders errors when token is invalid", %{conn: conn} do
+    current_user = insert(:user)
     {:ok, _} = AuthToken.changeset(%AuthToken{}, current_user) |> Repo.insert
     attrs = %{"token" => "random token", "password" => "123456", "password_confirmation" => "123456"}
     conn = post conn, password_reset_path(conn, :reset_password), attrs
-    assert json_response(conn, 422)
+    assert json_response(conn, 404)
   end
 
 end
--- a/web/controllers/password_reset_controller.ex
+++ b/web/controllers/password_reset_controller.ex
@@ -10,28 +10,20 @@ defmodule CodeCorps.PasswordResetController do
   2. password & password_confirmation match
   and return email. 422 if pwd do not match or auth token does not exist
   """
-  def reset_password(conn, %{"token" => token, "password" => password, "password_confirmation" => password_confirmation}) do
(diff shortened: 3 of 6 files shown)


snewcomer:
I just noticed one problem.  It is assuming the user is authenticated.  Will have to look up the token on the AuthToken model and then get the user off it as well.

web/router.ex
```diff
@@ -92,6 +91,7 @@ defmodule CodeCorps.Router do
 
     post "/token", TokenController, :create
     post "/token/refresh", TokenController, :refresh
+    post "/password/reset", PasswordResetController, :reset_password
```
snewcomer, line 95: @begedin just confirming this change.....

blueprint/api.apib
```diff
@@ -481,6 +481,33 @@ Until the Code Corps platform is open to new organizations, only admin users can
...
+### Reset Password [POST]
+
++ Request
+
+    + Attributes (Reset Password Request)
+
+    + Headers
+
+            Accept: application/vnd.api+json
+            Authorization: Bearer <token>
```
begedin: If the request does not need authorisation, should this be here?

web/router.ex
begedin: This is to mimic the ember route, I'm guessing? I'm ok with the change. Really, I don't think there's a right/wrong way to name things here.

web/controllers/password_reset_controller.ex
```diff
@@ -11,14 +11,14 @@ defmodule CodeCorps.PasswordResetController do
...
-        with %Changeset{valid?: true} <- User.reset_password_changeset(user, 
-                                                                       %{password: password, password_confirmation: password_confirmation}) do
+        user = Repo.get(User, user_id)
+        with %Changeset{valid?: true} = changeset <- User.reset_password_changeset(user, 
+                                                                       %{password: password, password_confirmation: password_confirmation}),
+             {:ok, user} <- Repo.update(changeset) do
           conn
-          |> put_status(:created)
-          |> render("show.json", email: user.email)
+          |> TokenController.create(%{"username" => user.email, "password" => password})
```
begedin: I'm not sure I like that one controller should be calling another. Maybe both should be calling the same module instead? I'd say, leave it as is in this PR, but create an issue to extract that behavior into a module and link it here. What do you think?

begedin (changes requested):

web/controllers/password_reset_controller.ex
snewcomer: So are you thinking the reset password will call the `TokenController`?

snewcomer: @begedin So one thing I noticed this weekend was when calling the TokenController from this controller, it still used the PasswordResetView (I assumed it would use the token controller).  So in this case, the token controller will return the token and user id.  However, the frontend would need the `email` also since the password reset route only knows about the `password/passwordConfirmation` values in order to login right away.

snewcomer: If you agree, then I can add the email in the token controller as well.

begedin: @snewcomer My proposal was, extract the logic from the token controller into a module. Then both the `TokenController` and the `PasswordResetController` call the same function in a third module.

That would eliminate the problem with `TokenController` using it's own view.

Looking at it now, thought, the part you need from `TokenController.create` is

```Elixir
{:ok, token, _claims} = user |> Guardian.encode_and_sign(:token)

conn
|> Plug.Conn.assign(:current_user, user)
|> put_status(:created)
|> render("show.json", token: token, user_id: user.id, email: user.email)
```

I would just put that smaller part of the code here, instead of calling `TokenController.create`

begedin: @snewcomer The two with blocks are really hard to follow here.. The code could be rewritten, and IMO would be easier to follow as: 

```Elixir
def reset_password(conn, %{"token" => reset_token, "password" => password, "password_confirmation" => password_confirmation} = params) do
  with %AuthToken{user: user} <- AuthToken |> Repo.get_by(%{ value: reset_token }) |> Repo.preload(:user),
       {:ok, _} <- Phoenix.Token.verify(CodeCorps.Endpoint, "user", reset_token, max_age: 1209600),
       {:ok, updated_user} <- user |> User.reset_password_changeset(params) |> Repo.update,
       {:ok, auth_token, _claims} = updated_user |> Guardian.encode_and_sign(:token)
  do
    conn
    |> Plug.Conn.assign(:current_user, updated_user)
    |> put_status(:created)
  else
    {:error, %Changeset{} = changeset} -> conn |> put_status(422) |> render(:errors, data: changeset)
    nil -> conn |> put_status(:not_found) |> render(CodeCorps.ErrorView, "404.json")
  end
end
```

begedin (changes requested):

web/controllers/password_reset_controller.ex
snewcomer: Nice!  Yeah muchhh better.  Thanks a ton for that!  

One thing is this was the error I got for `render(:errors, data: changeset)` so I did `render(CodeCorps.ErrorView, "422.json")`

image

blueprint/api.apib
```diff
@@ -2795,6 +2821,18 @@ The platform stores Stripe customers and cards so they can be reused across diff
     + title: `404 Resource not found` (string)
     + status: 404 (number) - HTTP status code
 
+## Reset Password Resource (object)
++ include Reset Password Resource Identifier
+
+## Reset Password Resource Identifier
++ token: `zzz123` (string, required)
++ password: `mypwd` (string, required)
++ password-confirmation: `mypwd` (string, required)
```
begedin: I'd just move this under `Reset Password Resource`. If the identifier and the resource are the same, there's little point of having the identifier. Just gives an extra step in reading the file.

test/controllers/password_reset_controller_test.exs
```diff
@@ -2,31 +2,33 @@ defmodule CodeCorps.PasswordResetControllerTest do
   @moduledoc false
 
   use CodeCorps.ApiCase, resource_name: :password_reset
-  alias CodeCorps.AuthToken
+  alias CodeCorps.{User, AuthToken}
 
-  @tag :authenticated
-  test "creates and renders resource when data is valid", %{conn: conn, current_user: current_user} do
+  test "creates and renders resource when data is valid", %{conn: conn} do
```
begedin: I'd make this test name less generic. 

"updates user password when data is valid"

web/controllers/password_reset_controller.ex
```diff
@@ -10,29 +10,20 @@ defmodule CodeCorps.PasswordResetController do
...
+         {:ok, auth_token, _claims} = updated_user |> Guardian.encode_and_sign(:token)
+    do
+      conn
+      |> Plug.Conn.assign(:current_user, updated_user)
+      |> put_status(:created)
+      |> render("show.json", token: auth_token, user_id: updated_user.id, email: updated_user.email)
     else
-      nil ->
-        handle_reset_pswd_result(conn)
+      {:error, %Changeset{} = _changeset} -> conn |> put_status(422) |> render(CodeCorps.ErrorView, "422.json")
```
begedin: I'm trying to find any reference to "422.json" in `JaSerializer` and failing. If this passes without errors, then the status of 422 will be applied to the response, but I don't think using the `422.json` template actually does anything. 

Even if it does, you'd still need to provide it error data to render, otherwise we will just get a blank 422 response.

Did you try 

```Elixir
conn |> put_status(422) |> render(CodeCorps.ErrorView, :errors, data: changeset)
```

If that does not work

```Elixir
conn |> put_status(422) |> render(CodeCorps.ErrorView, 'errors.json-api', data: changeset)
```

Additionally, we'd need to add the assertion to the controller test. It's not enough to check that the response is a 422. We also need to check that it contains the "passwords do not match" error. In most cases we do not check for that because the serialization is handled by `JaSerializer`/`JaResource` internally. In this case, however, we are doing it manually.

begedin (changes requested):

web/controllers/password_reset_controller.ex
snewcomer: Ah now it makes much more sense.  I had assumed in the past that JaResource had handled 422 errors.

snewcomer:
@begedin Not sure your thoughts in getting this pwd stuff merged in, but I would think we can get some of this API work in, then get the email created, then test out it's interaction w/ the Ember side?

joshsmith:
@snewcomer yes I agree on getting this merged in and then focusing on the email.

joshsmith:
This is looking pretty good to me. Can you squash and we can get this merged?

joshsmith (approved):

snewcomer:
squashed

joshsmith:
@snewcomer looks like some small conflicts with the other one just now merged. Are you able to resolve and then I'll merge this ASAP?

PR #805, opened 2017-05-07
Integration of Tasks with GitHub issues

### LINKED ISSUE #803: When updating a task if it belongs to a GithubRepo update associated issue on github
# Problem

In the first iteration of github integration, a project can be assigned a repository on github, via a dropdown. Likewise, a task can then be assigned an issue on github.

If that is the case, when that task updates, the associated issue should update as well.

While we're handling this, we should extract task controller behavior into a separate module.

## Subtasks
- [ ] If it doesn't already exist, add `/lib/code_corps/task/task.ex` module, namespaced `CodeCorps.Task`
- [ ] Add `update/3` function to that module
   - Function accepts task, an attributes map and the current user
   - Function updates task, then updates associated github issue if necessary (using information from associated `GithubRepo` and the `github_issue_number` field.
   - Function returns `{:ok, task}` or `{:error, changeset}` or `{:error, _}`
   - If the current user is GitHub connected, the github API calls should be made on their behalf. If not, they should be made on the behalf of the default user provided by #800
- [ ] Modify `TaskController.update` to call the new module function
- [ ] Add tests for new service module/new function which assert these four cases:
  - Task with no github id is updated, `{:ok, updated_task}` was returned, no github api call was made
  - Task with github id was updated, `{:ok, updated_task}` was returned, a github call was made
  - There was a validation error
  - There was some other error

The existing controller tests can remain as is.

### LINKED ISSUE #804: When creating a task, if it belongs to a GithubRepo, create associated issue on github
# Problem

A project can be tied to one or more github repositories, through the `ProjectGithubRepo` relation

When creating a task for a project, the attributes should contain a github_repo_id, meaning a task should now belong to a `GithubRepo` (but the relationship can be `nil`).

While we're handling this, we should extract task controller behavior into a separate module.

## Subtasks
- [ ] If it doesn't already exist, add `/lib/code_corps/task/task.ex` module (namespace `CodeCorps.Task`
- [ ] Add `create/2` function to that module
   - Function accepts an attributes map and the current user
   - Function initializes task (associated with `GithubRepo`), creates an associated github issue for the assigned github repo, the stores that github issue number as the task `:github_number` field and finally saves the task
   - Function returns `{:ok, task}` or `{:error, changeset}` or `{:error, _}`
   - If the current user is GitHub connected, the github API calls should be made on their behalf. If not, they should be made on the behalf of the identity provided in #800
- [ ] Modify `TaskController.create` to call the new module function
- [ ] Add tests for new function which assert these four cases:
  - Task with no github id is created, `{:ok, task}` was returned, no github api call was made
  - Task with github id was created, `{:ok, task}` was returned, a github call was made
  - There was a validation error
  - There was some other error

npendery:
Closes #803, #804, #806

# What's in this PR?

- [x] Add Github service `create_issue\3` function
- [x] Add testing to task create function to account for github creation
- [x] Add tentacat dependency to connect to Github
- [x] Update project to include `github_owner` and `github_repo` fields

## To do

- [x] Add default user info for github issue creation

## References
Fixes #804


npendery:
@begedin & @JoshSmith, mind helping me with the test error? not sure why the `github_id` isnt being returned in the JSON response, think `ja_resource` is doing some magic so might need to work around it.

lib/code_corps/github.ex
```diff
@@ -12,4 +12,33 @@ defmodule CodeCorps.Github do
...
+        nil
+    end
+  end
+
+  defp issue_attributes(attributes) do
+    %{
+      title: attributes[:title],
+      body: attributes[:body]
+    }
+  end
```
begedin: This seems sort of redundant, but my guess is, the intention was to have an explicit adapter function?

web/controllers/task_controller.ex
```diff
@@ -40,12 +41,22 @@ defmodule CodeCorps.TaskController do
...
 
   @spec handle_create(Plug.Conn.t, map) :: Ecto.Changeset.t
-  def handle_create(_conn, attributes) do
+  def handle_create(conn, attributes) do
+    project = Project |> Repo.get(attributes["project_id"])
+    if project.github_id do
+      current_user = Guardian.Plug.current_resource(conn)
+      github_id = github_module().create_issue(attributes, project, current_user)
+      attributes = Map.merge(attributes, %{"github_id" => github_id})
+    end
```
begedin: I apologise if I'm reviewing too early and it was already your intention to do this, but this behavior would be what we'd expect to put into the task_service module described in #804

web/controllers/task_controller.ex
```diff
@@ -40,12 +41,22 @@ defmodule CodeCorps.TaskController do
...
   end
 
   @spec handle_update(Plug.Conn.t, Task.t, map) :: Ecto.Changeset.t
   def handle_update(_conn, task, attributes) do
     task |> Task.update_changeset(attributes)
   end
+
+  defp github_module do
+    Application.get_env(:code_corps, :github)
+  end
```
begedin: I'm thinking maybe we call this `github_client` or even just `github`

web/controllers/task_controller.ex
npendery: Hey @begedin, I initially did that and was wondering the reasoning for pulling this out into a service? There really isnt a lot of complicated logic (most of that happens in the Github service) and wondering why we would create a task service just to copy the lines we have here? Are we thinking of using it beyond the CRUD actions? thanks!

lib/code_corps/github.ex
npendery: Pulled it out because I was thinking of reusing it for the future `update_issue` function. Also was thinking we might expand it to be more fields (when we want to be able to assign users, labels, etc.)

npendery: Never mind, we can come back and use this once we add the functionality for that, will update

web/controllers/task_controller.ex
begedin: There are no strong reasons right now, but

- It's more in line with how Phoenix 1.3 will structure modules.
- A clean, non-controller module will be easier to test for various outcomes.
- The code will very likely grow beyond these few lines.
- Generally, we like to keep our controller code as tiny as possible.

npendery: Cool, thanks for the points! Will make the change

npendery:
@begedin Mind looking over this for me? Thanks!

lib/code_corps/github.ex
```diff
@@ -12,4 +14,30 @@ defmodule CodeCorps.Github do
...
+  def create_issue(task, current_user, github_owner, github_repo) do
+    access_token = current_user.github_access_token || default_user_token() # need to create the Github user for this token
+    client = Tentacat.Client.new(%{access_token: access_token})
+    request = Tentacat.Issues.create(
+      github_owner,
+      github_repo,
+      Map.take(task, [:title, :body]),
+      client
+    )
+    case request do
```
begedin: Maybe call this a `result` or a `request_result`. Not a big deal, though.

lib/code_corps/github_testing.ex
```diff
@@ -0,0 +1,11 @@
+defmodule CodeCorps.GithubTesting do
+  alias CodeCorps.{Task, Repo}
+
+  def create_issue(task, _current_user, _github_owner, _github_repo) do
+    send self(), {:create_issue, task.id}
+
+    task
+    |> Task.github_changeset(%{"github_id" => 1})
+    |> Repo.update()
+  end
```
begedin: Great work on this one.

priv/repo/migrations/20170505224222_add_github_repo_and_owner_to_project.exs
```diff
@@ -0,0 +1,10 @@
+defmodule CodeCorps.Repo.Migrations.AddGithubRepoAndOwnerToProject do
+  use Ecto.Migration
+
+  def change do
+    alter table(:projects) do
+      add :github_repo, :string
+      add :github_owner, :string
```
begedin, line 7: I think this means we will no longer require a `github_id`. Same would go with ember's side. Would you agree?

begedin:
@npendery This is great work and something that's going to get merged ASAP, but sadly, not before we're done with the connect feature.

priv/repo/migrations/20170505224222_add_github_repo_and_owner_to_project.exs
npendery: Yeah, reviewing the Github docs more it looks like we can use the repo name for everything we need

joshsmith:
@npendery we're working to get the GitHub connect stuff merged now and then can focus on this again shortly.

begedin:
Hey @npendery 

We're finally at a point with the rest of the project where this can be continued and hopefully wrapped up. If you're still interested in doing so, please let me know and feel free to start by resolving conflicts, etc.

Also note that, with the new knowledge, I've updated the issue a good amount, so feel free to go back to it for the update.

For converting task attributes into github issue attributes, you'll probably want to extend the `GitHub.Adapters.Task` or even add a `GitHub.Adapters.Issue`

Finally, if you're able to only get parts of this done, feel free to do so. If it's a self contained part of the solution, we will be more then happy to merge so other parts can be done separately.

Feel free to also submit separate smaller PRs, for parts of the solution, such as the adapter.

begedin:
@npendery I've rebased this, but it looks like most of the changes here are also contained in #806 and it may be easier to just continue from there. I've added some much overdue information into that one.

Let me know how you feel about continuing this, but my guess is, either way, we will be closing this one and focusing on #806

begedin:
Started moving this towards the new approach.

TODOS:

- [ ] New migration is wrong. `github_id` should be `github_issue_number`, not `github_issue_id`
- [ ] We should use tentacat, but with using it, and in other ways to, relying on bypass across the app is problematic. I'm considering other options based on stuff I've learned on a client project. Will write up a suggestion as soon as I can
- [ ] Some tests are broken due to the column rename. I've also written some placeholder tests

joshsmith:
> I'm considering other options based on stuff I've learned on a client project. Will write up a suggestion as soon as I can

@begedin I'm curious what you mean here.

begedin:
In our client project, I've found a relatively simple way to do mocks of the whole api layer.

Basically, there is a think api module which wraps external requests. This module is then mocked in tests that need it using a macro. What that ends up looking like is something along the lines of

```elixir
with_mock_api(MockModule) do
  do_stuff
end
```

In that project, we have a basic success mock, where all requests return the default successful response, a basic failure module, where all requests return the default failure response, and with the way the macro works, we can define simple modules with specific responses, "inline".

In our specific case, we would probably mock the `request` module. The important part is, we need to make sure all our external requests are centralized, which is already something we strive to do.

begedin:
@JoshSmith I've been having trouble finding a nice way to implement things here. 

The way I see it, we have two issues

- we can't really mock `Tentacat` using `bypass`
- it's difficult to keep our mocks in check with things scatted a bit across `CodeCorps.GitHub`, `CodeCorps.GitHub.Request`, etc.

My latest commit is just a bunch of scattered things that I'm not sure should be kept in, but I wanted to at least show some of it, so I pushed it up. Feel free to remove if you don't find it useful.

I would like to direct your attention to the `with_mock_api` macro which I wrote. It allows us to define a tiny module which could handle a custom request by asserting, sending a custom response, etc and wrap a code calling our api forcing it to call that module instead.

In order for this to work, we cannot load our api module from the config using a module `@variable`. Instead, it needs to be a function, so it's evaluated at execution time, not compile time.

I've defined it due to the way we use a similar approach with `SparkPost` in our client project.
We could, in addition to this, have a `with_mock_tentacat` macro as well. 

These could be useful tools depending on how we continue. Really, though, the issue here is
that our communication layer for github is a bit too scattered to be easily testable. I'm not even 100% positive using tentacat here is worth it, or even possible (I'm not sure how it works with integrations). 

I'm sorry I'm just listing problems instead of offering solutions, but I'm having difficulties wrapping my head around this yet.

## Considerations

1. We need to figure out if we can use tentacat here
2. We need to decide if it's worth using tentacat here, considering the scatteredness it creates
3. We need to consolidate our own github requests to make them more easily mockable
4. We need to consider ditching bypass in favor of the mock approach. It may affect how we consider other things listed here.

## Approaches for mocking

1. We have a module with a `request` method. All requests go through that method. We mock that module, assert it was called with proper arguments.

2. We have a module with a bunch of "flat" methods such as `create_issue`, `list_repositories`, etc, and we mock those.

If we use tentacat, approach 2 is still compatible. For example, `create_issue` would internally call `Tentacat.Issues.create`, etc. Approach 1 is not compatible directly, but we could work around it.

I'm leaning towards 2 if we decide to stick with tentacat and 1 if we decide to drop it.

joshsmith:
I think we should consider dropping tentacat given the difficulties exposed here, believe it would help us clean up our GitHub code generally, and provide some extra flexibility going forward. We're not hitting so much API surface area that it wouldn't be worth doing that.

Also agree with your approach to mocking.

Given this, could you outline how you would approach this move in an issue or issues so I can help work to wrap that up this week with time left to still get the integration actually functional?

lib/code_corps/helpers/query.ex
```diff
@@ -37,32 +37,11 @@ defmodule CodeCorps.Helpers.Query do
...
-  def task_list_filter(query, %{"task_list_ids" => task_list_ids}) do
-    task_list_ids = task_list_ids |> coalesce_id_string
-    query |> where([object], object.task_list_id in ^task_list_ids)
-  end
-  def task_list_filter(query, _), do: query
-
-  def task_status_filter(query, %{"status" => status}) do
-    query |> where([object], object.status == ^status)
-  end
-  def task_status_filter(query, _), do: query
```
begedin, line 56: Moved into separate module specific for task and tested

begedin:
@JoshSmith This now fully handles task sync with github and should be good for a review/merge.

It should also serve as a decent template for comment sync, both for implementation and tests.

begedin:
Also note that we can close #806 as well, when this is merged.

joshsmith:
@begedin it appears there's a merge conflict now from the recent merge in the policy.

begedin:
@JoshSmith Rebased and resolved

lib/code_corps_web/plugs/ids_to_integers.ex
```diff
@@ -0,0 +1,78 @@
...
+  defp convert?("stripe_platform_customer_id"), do: true
+  defp convert?("task_id"), do: true
+  defp convert?("task_list_id"), do: true
+  defp convert?("task_skill_id"), do: true
+  defp convert?("user_id"), do: true
+  defp convert?("user_category_id"), do: true
+  defp convert?("user_role_id"), do: true
+  defp convert?("user_skill_id"), do: true
+  defp convert?("user_task_id"), do: true
+  defp convert?(_other), do: false
```
joshsmith, line 72: Would it not make more sense to here to check if it's in a list? i.e. like how Ecto validations work?

joshsmith:
I just pushed up some fixes that I identified in a new commit.

Possibly outside the scope of this PR, but some other things I've noticed while testing:

- When users are created on CC because we don't have any matching info from GH:
  - They don't receive a username
  - They don't receive a `default_color`
  - They don't have an avatar (it's a broken image), although we do have a GitHub avatar they could use
  - Their sign up context is `default`; this should probably be set to `github` or `github-api` instead

- A user gets created for `code-corps-local[bot]` when an issue is opened. This likely shouldn't be happening, but I'm not quite sure how we should approach this situation.

- I got an `opened` event for `issues` (created by a GH user that had no corresponding user on CC) that resulted in no issue being created, a 200 returned, a `processed` status, and no user created on CC. That means the flow is broken and it's not clear where it's breaking down.

- It's really hard to debug what's happening when something goes wrong since we're not logging anything. I did end up logging at least the 401 returned by GitHub, which helped me identify that our timestamp code was breaking. But it took quite a long amount of time to get down the stack.

We can break these out into some follow-up issues if necessary since I'm not positive any of this relates to the PR at hand (aside from the bot situation above). I was ultimately able to create issues effectively (see https://github.com/code-corps/test/issues). 7-11 are tests created from my localhost, and 12 is the test created _on GH_ that then failed to sync properly to my localhost.

begedin:
- Created #938 for handling github user info when creating user.
- Created #939 for handling bot user when a non-connected user creates a task on cc
- Created #940 to verify what goes wrong that would cause user not to be created when issue is opened on github directly

begedin:
@JoshSmith I rebased and resolved conflicts again.

I think we need the longs, but I also think we should mute them in the test environment, since they pollute the output.

`config :logger, level: :warn` changing this to `:error` may even be enough. It would also mute the clouded warning, but really, that one should also be rewritten so an env variable is not needed in tests, and either way, the tests, if they get broken, will fail on CI.

Some notes on rebasing if there's a need for you to do so:

- `api.ex` and `mix.exs` are easy
- for `mix.lock`, simply take the entire `HEAD` version, and then remove the `tentacat` entry from it. 
- you'll have to `mix deps.get` one package at a time. If you just `get` all the packages, you'll pull the latest `stripity_stripe`, since we target the 2.0 branch, which will break things.

joshsmith:
@begedin should we place these rebasing notes in the README somewhere? Seems like it might end up being a fairly common thing to accidentally pull in these changes from the 2.0 branch since we're not locked to a particular commit hash or anything.

joshsmith (approved):

PR #807, opened 2017-05-11
[GITHUB]: connect

snewcomer:
This PR is meant to retrieving an access token for a user and updating the User record with their github_id.  

Not sure where constants `@token_url`, `@client_id`, `@client_secret` come from.

Progress on: #789


lib/code_corps/github.ex
```diff
@@ -2,10 +2,37 @@ defmodule CodeCorps.Github do
...
-  def connect(user, _code), do: {:ok, user}
+  def connect(user, code) do
+    github_post(code)
+    |> Tentacat.Client.new
+    |> Tentacat.Users.me
+    |> update_user(user)
+  end
+
+  defp update_user(%{github_id: _github_id} = github_info, user) do
+    associate(user, github_info)
```
snewcomer: Maybe just pass the `github_id` instead of the whole map?

lib/code_corps/github.ex
```diff
@@ -2,10 +2,37 @@ defmodule CodeCorps.Github do
 
   alias CodeCorps.{User, Repo}
 
+  @token_url ""
+  @client_id ""
+  @client_secret ""
+
```
snewcomer: Need help with these.

snewcomer:
Just looking for some initial feedback and help!  Haven't looked at testing, b/c wanted to see first if going down the right path.

lib/code_corps/github.ex
begedin: `client_id` and `_secret` will be environment variables which we will get when we register CodeCorps as a github application. It may even make sense not to have them as module variables, so we can inject them in tests. 

Either way, assume the environment variables `GITHUB_CLIENT_ID` and `GITHUB_CLIENT_SECRET` exist, are loaded during config and stored into a config key.

For `@token_url`, this should simply be `https://github.com/login/oauth/access_token`

lib/code_corps/github.ex
begedin: You mean, pass into `associate/2`? That would work for now, if that's the only thing we need to pass in. Later on, we might not have just the client ID but instead store a larger set of values, but we can worry about that when we come to it.

That being said, if associate already exists, and tests for it exist, no point i changing it now.

begedin:
@snewcomer I'm currently in the process of implementing the API key stuff as part of this PR, since, in order to do that, I need to work with Github API keys for CodeCorps. I'll try to implement it in a way so that tests will be reliable, but will not require other developers to provide API keys. Be sure to pull the latest before doing more work on this. I'll make sure to do the same.

begedin:
@snewcomer My commit still needs thorough testing and wrapping up, but I managed to get through the basic flow.

- Renamed user.github_id to user.github_auth_token, since that's a more accurate name.
- Extracted the actual API calls into a separate module, so we can inject a test module in tests instead.
- For testing, you can use the test module.
- In dev, you can temporarily set the test module as the one being used, or you can register a github oauth app on your github profile, add those keys to your `.env` file (look at `.env.example` to see what to name them, then run `source .env` from bash before running the app. 
- At that point, you'd need to
  - visit `https://github.com/login/oauth/authorize?scope=user:email&client_id=<your_client_id>` to get a `code`
  - use that code when calling `Github.connect(user, code)`
If you can work around this, feel free. Otherwise, I'll be able to wrap it up tomorrow.

lib/code_corps/github/api.ex
```diff
@@ -0,0 +1,28 @@
...
+  @client_secret Application.get_env(:code_corps, :github_client_secret)
+  @client_id Application.get_env(:code_corps, :github_client_id)
+
+  @base_connect_params %{
+    client_id: @client_id,
+    client_secret: @client_secret
+  }
+
+  def connect(code) do
+    with {:ok, %HTTPoison.Response{body: response}} <- code |> build_connect_params() |> do_connect(),
```
snewcomer: @begedin Just curious on the reason not to use Tentacat.

lib/code_corps/github.ex
```diff
@@ -2,38 +2,33 @@ defmodule CodeCorps.Github do
...
+  - {:ok, %CodeCorps.User{}}
+  - {:error, %Ecto.Changeset{}}
+  - {:error, "some_github_error"}
   """
   def connect(user, code) do
-    github_post(code)
-    |> Tentacat.Client.new
-    |> Tentacat.Users.me
-    |> update_user(user)
+    case code |> @api.connect do
```
snewcomer: Yeah I like this move to a separate module for testing

snewcomer (commented):
So guessing we can remove the tentacat dep?  These changes look great!

lib/code_corps/github/api.ex
begedin: I had issues getting it to work. The documented methods do not work as documented. They also raise exceptions when they should not be doing so and are named without exclamation marks.

Due to that, for this specific case, I went with a direct API call.

For the regular stuff, in the integration milestone, we should go with Tentacat.

lib/code_corps/github/api.ex
```diff
@@ -0,0 +1,28 @@
...
+      {:ok, %{"access_token" => access_token}} <- response |> Poison.decode
+    do
+      {:ok, access_token}
+    else
+      {:ok, %{"error" => error}} -> {:error, error}
```
joshsmith, line 29: It's a bit weird to me that this returns an `:ok` tuple.

priv/repo/migrations/20170515154812_rename_user_github_id_to_github_auth_token.exs
```diff
@@ -0,0 +1,7 @@
+defmodule CodeCorps.Repo.Migrations.RenameUserGithubIdToGithubAuthToken do
+  use Ecto.Migration
+
+  def change do
+    rename table(:users), :github_id, to: :github_auth_token
```
joshsmith: Are you no longer expecting to track the GitHub user ID? I would think this would be important information for us to have on-hand and indexed.

web/models/user.ex
```diff
@@ -97,8 +98,8 @@ defmodule CodeCorps.User do
 
   def github_associate_changeset(struct, params) do
```
joshsmith: Why not `github_association` changeset, since it looks like this wants to be a noun and `associate` in this case is verbed?

joshsmith (dismissed):
Minor comments that I think deserve some work before a merge.

lib/code_corps/github/api.ex
npendery: Yeah, surprised a HTTPoison request with an error response would return an `:ok` tuple, there should be a better way of differentiating a HTTPoison error and a response error

begedin: It is what it is, though. Seems that github itself returns a 200 ok, with an error hash. Not much we can do about it, other than interpret it here.

priv/repo/migrations/20170515154812_rename_user_github_id_to_github_auth_token.exs
begedin: I did not consider that. We'd need it as part of the integration feature. Technically, however, all the connect feature needs is the `auth_token`.

I'll revert this change and make it so both fields are there.

joshsmith:
@snewcomer @begedin are we able to implement the above minor suggestions and then get this merged?

begedin (approved):
@JoshSmith Made the changes, rebased and squashed. This ought to be good to merge.

I've added support for an optional parameter for the connect method, so we can explicityl inject a boundary module when testing. This allows really easy testing for various response cases and is advised to do in http://blog.plataformatec.com.br/2015/10/mocks-and-explicit-contracts/

PR #808, opened 2017-05-30
Update password reset to add email

joshsmith:
# What's in this PR?

- Adds email template for forgot password
- Fixes a `Poison.EncodeError` for a mistaken change in `CodeCorps.Cloudex.CloudinaryUrlTest`; introduces `placehold.it` URLs here that are testable
- Minor fix for password reset view
- Copy changes
- Changes reset expiration from 14 days to 1 hour

### DIFF
--- a/emails/forgot_password.html
+++ b/emails/forgot_password.html
@@ -1,0 +1,90 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
+<html xmlns="http://www.w3.org/1999/xhtml">
+  <head>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
+    <title>Here's the link to reset your password.</title>
+    <!--
+    Make sure you copy the styles from styles.css into the email template in Postmark before saving there.
+
+    <style type="text/css" rel="stylesheet" media="all">
+    </style>
+    -->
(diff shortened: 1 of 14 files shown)


joshsmith:
Looked at coverage and no meaningful reason for drop other than known `schema` issues with `excoveralls`.

web/views/password_reset_view.ex
```diff
@@ -1,9 +1,11 @@
 defmodule CodeCorps.PasswordResetView do
   use CodeCorps.Web, :view
 
-  def render("show.json", %{token: token}) do
+  def render("show.json", %{email: email, token: token, user_id: user_id}) do
     %{
-      token: token
+      email: email,
+      token: token,
```
snewcomer, line 7: Not seeing the token/user_id being consumed on the [frontend](https://github.com/code-corps/code-corps-ember/blob/84d91618ccb464b112555298600b366aeb53dc7f/app/routes/password/reset.js#L30).

emails/forgot_password.html
```diff
@@ -0,0 +1,90 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
```
snewcomer, line 1: What is the need for this file if there is a postmark template?

joshsmith: @snewcomer for us to be able to design and preview it locally.

web/views/password_reset_view.ex
joshsmith: `user_id` is used by ESA, and we might as well have the token, I think.

PR #814, opened 2017-06-02
Loosen required elixir version to 1.4 or greater

dhedlund:
The mix.exs file has the version of elixir pegged to 1.4.1.  This increases the difficulty of onboarding new contributors and may no longer be necessary.

From the git history, it appears it was pegged to a specific version back when trying to get the project to work with both docker and heroku; the commits were squashed together, so it's hard to tell which.  Since that time, the docker configurations were removed and the elixir version is now defined in elixir_buildpack.config for both heroku and circle ci.

I'm hoping that this PR will trigger circle ci verification which should hopefully verify both cases still work?

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -6,7 +6,7 @@ defmodule CodeCorps.Mixfile do
   def project do
     [app: :code_corps,
      version: "0.0.1",
-     elixir: "1.4.1",
+     elixir: "~> 1.4",
      elixirc_paths: elixirc_paths(Mix.env),
      compilers: [:phoenix, :gettext] ++ Mix.compilers,
      dialyzer: [plt_add_deps: :transitive],


begedin:
@dhedlund That was some in depth research.

I don't have a problem with doing this. Minor versions should not contain anything breaking, and we can still ensure specific versions in our deploys if needed.

@JoshSmith What do you think?

joshsmith:
I agree with this move!

joshsmith (approved):

joshsmith:
@dhedlund would you mind rebasing just to be sure there's not any new issue? Then we can merge at will.

dhedlund:
I've rebased my changes on the latest develop branch and CircleCI completed successfully.

joshsmith:
🙌  Just invited you to the repo, as well! Merged myself this time to prevent another rebase, but you've got a commit bit now so you can do it in the future yourself.

Thanks again!

PR #816, opened 2017-06-04
Code corps api#692/add validation to donation goals: positive numbers only

### LINKED ISSUE #692: Add validation to donation goals: Positive numbers only
# Problem

There should be a validation of donation goals, specifically disallowing negative numbers. At present, there is no check in the app or in the api.

jayjayjpg:
# What's in this PR?

This adds a validation for the amount attribute of donation goals to be greater than zero. It will validate any requests for the creation of donation goals or for the update of donation goals to the backend and will return an "Amount must be greater than 0" error if the validation fails.

Donation goal creation workflow:
addvalid1

Donation goal update workflow:
addvalid2

## References
Fixes #692

### DIFF
--- a/test/models/donation_goal_test.exs
+++ b/test/models/donation_goal_test.exs
@@ -26,6 +26,15 @@ defmodule CodeCorps.DonationGoalTest do
       refute changeset.valid?
       changeset |> assert_error_message(:project, "does not exist")
     end
+
+    test "amount must not be negative" do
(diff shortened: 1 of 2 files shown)


joshsmith (approved):
Excellent! Looks really great.

joshsmith:
🎉 I'm going to push to production as soon as it lands in staging.

PR #821, opened 2017-06-22
Add GitHubAppInstallation model, state transition, and tests

joshsmith:
# What's in this PR?

Implements some of what's needed in https://github.com/code-corps/code-corps-ember/issues/1331.

### DIFF
--- a/lib/code_corps/transition/github_app_installation_state.ex
+++ b/lib/code_corps/transition/github_app_installation_state.ex
@@ -1,0 +1,32 @@
+defmodule CodeCorps.Transition.GithubAppInstallationState do
+  @moduledoc """
+  Governs the transitioning from one state to the next.
+
+  The possible `state` values are:
+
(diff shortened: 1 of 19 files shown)


lib/code_corps/transition/github_app_installation_state.ex
```diff
@@ -0,0 +1,38 @@
...
+  def next("unmatched_user", "processed"), do: {:ok, "processed"}
+
+  def next(current_state, next_state) do
+    cond do
+      current_state == next_state ->
+        {:ok, current_state}
+      current_state != next_state ->
+        {:error, "invalid transition to #{next_state} from #{current_state}"}
+    end
+  end
```
begedin: you could rewrite this as two clauses:

```Elixir
def next(current_state, next_state) when current_state == next_sate, do: {:ok, next_state}
def next(current_state, next_state), do: {:error, "invalid transition to #{next_state} from #{current_state}"}
```

I think that makes it ever so slightly more readable.

begedin (commented):
Other than the minor comment, this looks good to me.

PR #822, opened 2017-06-26
Added GitHub webhooks endpoint

begedin:
This is nowhere near being done or ready to merge, but should serve as the initial infrastructure for handling github webhooks.

The idea is that we receive the action in our controller, figure out what the event type is, fire an async processing task and respond with the correct status code depending on us supporting or not supporting the event.

https://developer.github.com/v3/guides/best-practices-for-integrators/#use-appropriate-http-status-codes-when-responding-to-github

From that, I've concluded that unsupported events should be responded to with 202 (though 201 would be ok as well) and set things up so we can start adding in supported events.

Everything else sort of mimics the stripe event processing, except I've tried to follow a more conventional naming scheme.


joshsmith:
I'm thinking that some of the processing should be delegated to, for example, an `IssuesEventHandler`, since the logic seems to be complex depending on the event (see, e.g. https://developer.github.com/v3/activity/events/types/#issuesevent), and that would allow us to more easily understand and maintain the various processing that's happening when a given event occurs.

web/models/github_event.ex
```diff
@@ -0,0 +1,22 @@
+defmodule CodeCorps.GithubEvent do
+  use CodeCorps.Web, :model
+
+  schema "github_events" do
+    field :action, :string
+    field :github_delivery_id, :string
```
joshsmith, line 6: This is sent as a header with the webhook and should be used for ensuring we don't respond to multiple of the same event. There are no event IDs (!!!), so this is extremely important.

I'm still not quite sure what we're going to do about there not being event IDs. We're going to need some way of determining whether to apply a given change from GitHub to our data, and vice versa.

config/test.exs
```diff
@@ -54,3 +54,6 @@ config :code_corps,
 
 config :code_corps, :cloudex, CloudexTest
 config :cloudex, api_key: "test_key", secret: "test_secret", cloud_name: "test_cloud_name"
+
+# Plug.Conn errors on uppercased event names, so we need to mock it here
+config :code_corps, github_event_header_name: "x-github-event"
```
joshsmith: This is frustrating but I couldn't find a way around it. Fine with this for now.

lib/code_corps/github/webhook/handler.ex
```diff
@@ -0,0 +1,12 @@
+defmodule CodeCorps.GitHub.Webhook.Handler do
+  @moduledoc """
+  Receives and handles GitHub event payloads.
+  """
+
+  @doc """
+  Handles a GitHub event based on its type.
+  """
+  def handle(type, payload) do
+    # TODO: Implement
```
joshsmith: Yeah! I made no progress here!

test/models/github_event_test.exs
```diff
@@ -0,0 +1,18 @@
...
+
+  test "changeset with valid attributes" do
+    changeset = GithubEvent.changeset(%GithubEvent{}, @valid_attrs)
+    assert changeset.valid?
+  end
+
+  test "changeset with invalid attributes" do
+    changeset = GithubEvent.changeset(%GithubEvent{}, @invalid_attrs)
+    refute changeset.valid?
+  end
```
joshsmith, line 17: These are not useful tests yet.

web/controllers/github_events_controller.ex
```diff
@@ -0,0 +1,47 @@
...
+
+  defp check_support(event) when event in ~w(installation
+    installation_repositories issue_comment issues), do: :supported
+  defp check_support(event) when event in ~w(commit_comment create delete
+    deployment deployment_status fork gollum label marketplace_purchase
+    member membership milestone organization org_block page_build
+    project_card project_column project public pull_request_review_comment
+    pull_request_review pull_request push repository release status team
+    team_add watch), do: :unsupported
+  defp check_support(_event), do: :unsupported
```
joshsmith: This looks gross but also okay with proceeding unless there's a quick change to make it better.

web/controllers/github_events_controller.ex
```diff
@@ -0,0 +1,47 @@
...
+    ~w(installation installation_repositories issue_comment issues)
+  end
+
+  defp unsupported_events do
+    ~w(commit_comment create delete deployment deployment_status fork gollum
+    label marketplace_purchase member membership milestone organization
+    org_block page_build project_card project_column project public
+    pull_request_review_comment pull_request_review pull_request push
+    repository release status team team_add watch)
+  end
```
joshsmith: Forgot to remove these.

joshsmith (commented):
Some comments below in the code, but some more general comments here to consider.

We may need to split out our App's webhooks from user OAuth webhooks. This is because we appear to need _both_ the App and OAuth to do all the things on behalf of a user that we'd like to do. How we respond to each is thus likely to be different, and each app should likely have different webhook URLs for a cleaner separation of behavior.

For now, we may want to just progress with a `GithubEventsController` and two separate URLs, and simply record the source (App vs OAuth) depending on that.

I've tested out the integration with the App and certain endpoints are just altogether missing when doing something on behalf of a user with the App. That means we'll need a fallback.

Although well outside the scope of this PR, I'd like to take a moment to quickly explore the fallbacks, just for our own reference later. We can expect, when posting/editing an issue or comment, to:

- check first if the user has a GitHub account
- check the type of event
- check the GitHub user's integrations (App and OAuth), and their permissions

For example, if a user does not have a GitHub account, we'd need to post the issue as the `code-corps` App bot. If they did have one, we'd see if the event can be done by the App or needs to fall back to the integration. The App provides nicer feedback in GitHub's UI, but not all endpoints are available for us yet. So we may then need to fall back onto whether the user has connected with OAuth or the App. This latter part may be simplified if we can spend some more time thinking about what it means for someone to "connect with GitHub."

config/test.exs
begedin: Couldn't resist googling and that's just...

Ok, it's supposed to be a standard and a convention, but forcing it like this...

joshsmith (approved):

PR #829, opened 2017-06-28
Add installation event support

### LINKED ISSUE #825: Handle installation webhooks from GitHub
# Problem

We need to write code that handles installation webhooks in the `CodeCorps.GitHub.Events.Installation` module.

This should:

- Handle the `created` or `deleted` events
- Find the existing `CodeCorps.GitHubAppInstallation` for the GitHub user
- Update the `CodeCorps.GitHubAppInstallation` with a `processing` state
- Request repositories from [GitHub's endpoint](https://api.github.com/installation/repositories) using the GitHub app's installation token. It should then use these repositories to create or update `GitHubRepo` records and `ProjectGitHubRepo` records.
- Update the `CodeCorps.GitHubAppInstallation` with its new end state (hopefully `processed`)

joshsmith:
# What's in this PR?

Spike of #825. Not yet finished, but worked out the theoretical logic that should go into this. Got stuck at end of night trying to figure out control flow.

You can see #825 itself for more concreteness around what this is supposed to end up doing.

A couple thoughts:

- Create a new test case for this?
- Should we have adapters for the events? That would likely speed up the unit testing of an adapter layer, and separate any logic from these handlers. There's some possible reusability then across the different handling methods, rather than repeating the same basic extraction.
- We should update `GithubAppInstallation` to actually set the `boolean` of installed.
- It's obvious now that we will want to have a many-to-many relationship between `GithubAppInstallation` and a `Project`. This will have to be how a "group" of repositories is presented in a UI. This maybe needs some further discussion, but a mistake here early could be costly later on, especially in terms of our understanding of how these various relationships actually relate.
- We'll need to find a way to prevent someone creating more than one `GithubAppInstallation` for any given `installation_id`, and for a user from creating more than one `GithubAppInstallation` in an `initiated_on_code_corps` state.

## References
Fixes #825


lib/code_corps/github/adapters/github_repo.ex
```diff
@@ -0,0 +1,20 @@
+defmodule CodeCorps.GitHub.Adapters.GithubRepo do
```
begedin, line 1: Thinking about moving this to `lib/code_corps/github_repo/adapter.ex`. According to 

image

That's the place where it should be going, assuming `web/models/github_repo.ex` will be moved into `lib/code_corps/github_repo.ex` once we upgrade to 1.3.

That would provide us the namespace of `CodeCorps.GithubRepo.Adapter`. In the place where we use it right now, we could just alias `CodeCorps.GithubRepo` and then call `GithubRepo.Adapter.from_api`, which seams a bit nicer than what we have.

lib/code_corps/github/events/installation.ex
```diff
@@ -13,5 +24,89 @@ defmodule CodeCorps.GitHub.Events.Installation do
   - do the work
   - marked the passed in event as "processed" or "errored"
   """
-  def handle(%GithubEvent{}, %{}), do: :not_fully_implemented
+  @spec handle(GithubEvent.t, map) :: {:ok, GithubEvent.t}
+  def handle(%GithubEvent{action: "created"} = event, payload) do
+    event
+    |> start_event_processing()
+    |> do_handle(payload)
+    |> stop_event_processing(event)
```
begedin, line 32: The "processing" state seems redundant here. Sure, the processing action is slightly longer lasting that most things we do, but this is all called in succession, as part of one large async process, not as multiple async processes. It basically just adds an extra db write.

We never really call the handler for an existing event so it's not like we can match against `state: "processing"` her and abort if that is  the case.

Right now, our flow is always  "create event as unprocessed -> mark as processing -> handle payload -> mark as processed". Everything except the first step is done as a single async block, so it's not really doing us any good as it is right now. 

Feels like we're implementing for some future thing that might happen, may even likely happen, but we did a similar thing with stripe and it did not happen yet in that case. Yes, we have matching for a processing event with stripe, but I don't think our system allows for it to ever happen in a real life scenario. 

Even if it did, with stripe, there's at least the added argument of having unique event ids, so we `find_or_create` the event. With github, all we have are delivery ids, which are different with each webhook request, so we do no such thing.

lib/code_corps/github/events/installation.ex
```diff
@@ -13,5 +24,89 @@ defmodule CodeCorps.GitHub.Events.Installation do
...
+
+  defp create_installation_initiated_on_github(%User{} = user, %{"id" => github_id}) do
+    %GithubAppInstallation{}
+    |> Changeset.change(%{state: "initiated_on_github", github_id: github_id})
+    |> Changeset.put_assoc(:user, user)
+    |> Repo.insert()
+  end
+
+  defp update_installation(%User{} = user, %GithubAppInstallation{} = github_app_installation) do
+    # TODO: user and installation need to relate. What if they don't?
```
begedin: As the comment says, how do we deal with that edge case? Can it ever actually happen? Somehow, one user requests the integration from our client, then another user actually installs the app?

lib/code_corps/github/events/installation.ex
```diff
@@ -13,5 +24,89 @@ defmodule CodeCorps.GitHub.Events.Installation do
...
+
+    github_app_installation
+    |> start_app_processing()
+    |> process_repos(user)
+    |> stop_app_processing()
+  end
+
+  defp start_app_processing(%GithubAppInstallation{} = github_app_installation) do
+    github_app_installation
+    |> Changeset.change(%{state: "processing"})
```
begedin: In this case, there's slightly more to argue in favor of having a "processing" state, but again, I'm not sure it will actually happen in a real life scenario.

If it does, we need to

- abort
- mark the event as "errored" ? 
- implement a system which takes errored events and retries them? 

We do not store event data, though, so we would not be able to do that. It would be for logging purposes / manual intervention only.

lib/code_corps/github/events/installation.ex
```diff
@@ -13,5 +24,89 @@ defmodule CodeCorps.GitHub.Events.Installation do
...
+    |> Repo.update()
+  end
+
+  defp process_repos({:ok, %GithubAppInstallation{} = github_app_installation}, %User{} = user) do
+    # TODO: Consider moving into transaction
```
begedin: We create each event individually here.

- Should we move this into a transaction?
- Should just the repos be in a transaction, or the app update as well?

lib/code_corps/github/events/installation.ex
```diff
@@ -13,5 +24,89 @@ defmodule CodeCorps.GitHub.Events.Installation do
...
+    |> Changeset.change(%{state: "unmatched_user", github_id: github_id})
+    |> Repo.insert()
+  end
+
+  defp create_installation_initiated_on_github(%User{} = user, %{"id" => github_id}) do
+    %GithubAppInstallation{}
+    |> Changeset.change(%{state: "initiated_on_github", github_id: github_id})
+    |> Changeset.put_assoc(:user, user)
+    |> Repo.insert()
+  end
```
begedin: Do we need to create repos for this app as well, or do we wait for a matching user first?

lib/code_corps/github/installation.ex
```diff
@@ -0,0 +1,28 @@
...
+  """
+  @spec repositories(User.t, GithubAppInstallation.t) :: {:ok, list} | {:error, APIError.t}
+  def repositories(%User{github_auth_token: github_auth_token}, %GithubAppInstallation{github_id: installation_id}) do
+    endpoint = "user/installations/#{installation_id}/repositories"
+    case GitHub.Request.retrieve(endpoint, [access_token: github_auth_token]) do
+      {:error, %CodeCorps.GitHub.APIError{} = error} -> {:error, error}
+      {:ok, %{"total_count" => _, "repositories" => repositories}} -> {:ok, repositories}
+    end
+  end
+end
```
begedin: I'm not completely clear on the philosophy for the API part of our github namespace, nor am I sure that this is the request I should be making to get installation repositories, but it seemed like the most likely one.

begedin:
> We should update `GithubAppInstallation` to actually set the boolean of installed.

The 3 cases do this, but the default for that field is set to `true` on the db level. Might be a good idea to change the default to false.

> It's obvious now that we will want to have a many-to-many relationship between GithubAppInstallation and a Project.

Correct me if I'm wrong, but `GithubRepo` belongs to a `GithubAppInstallation`. `GithubRepoProject` belongs to a `GithubRepo` and a `Project`. Would not a `many-to-many` between a `GithubAppInstallation` and a `Project` be implicit here?

In the repo selection UI, we'd have

- a list of `GitHubRepos` belonging to the `GithubAppInstallation`
- in the background, the project would have a list of `ProjectGithubRepos` loaded
- each `GithubRepo` in the list would have a different rendering, based on there being a `ProjectGithubRepo` for that `GithubRepo`
  - if there is -> `Unfollow` (I know we were talking about having a status here, but unfollow could just as easily delete the `ProjectGithubRepo`, no?)
  - if there isn't -> `Follow` (Create the `ProjectGithubRepo`)

lib/code_corps/github/adapters/github_repo.ex
joshsmith: Let's do this refactoring after merge, maybe?

lib/code_corps/github/events/installation.ex
joshsmith: The point of the `processing` state here was that we could, in the UI, at least redirect the user to the proper project, although we could not present them yet with the list of GitHubRepos. Since I think the most common case in terms of timing upon redirect is "we've received the webhook and we're processing it", that means keeping them off the interstitial page and into the least surprising UI, which I think is preferable. It will make channels easier for us to deal with in that event, too.

lib/code_corps/github/events/installation.ex
joshsmith: A GitHub user can install a GitHub App without ever having been a Code Corps user.

joshsmith:
> The 3 cases do this, but the default for that field is set to true on the db level. Might be a good idea to change the default to false.

Can you think of a time where we'd be creating a `GitHubAppInstallation` from the `deleted` action? I think `true` does make sense as the default in this case.

lib/code_corps/github/events/installation.ex
joshsmith: We would create repos here, too.

lib/code_corps/github/events/installation.ex
begedin: I tried to give an example here and then realized the edge case I was thinking of could not happen. In fact, I'm even having trouble putting the thing I was thinking off into words, but it relied on the false idea that the primary key of our `GithubAppInstallation` record somehow becomes the github id. Basically, all wrong, so feel free to ignore. It was me thinking of something, but not fully thinking it through.

test/lib/code_corps/github/adapters/github_repo_test.exs
```diff
@@ -9,7 +9,7 @@ defmodule CodeCorps.GitHub.Adapters.GithubRepoTest do
 
   describe "from_api/1" do
     test "maps api payload correctly" do
-      %{"repositories" => [repo]} = load_fixture("user_repositories")
+      %{"repositories" => [repo]} = load_event_fixture("user_repositories")
```
begedin: You made a rename to `load_event_fixture`, but in this specific case, you're actually loading an api response fixture. I'm ok with naming, but then we'd need a `load_response_fixture` as well and probably split the two sets of fixtures into separate folders under `fixtures/github/`

begedin:
Not much to say about the changes, but I got a sort of idea.

With the different authentication methods, hitting potentially same endpoints and all the back and forth, I'm considering a different API building approach might be suitable.

We could have a `%Request{}` struct. Different layers/modules in our code would transform this base struct in different ways.

- different authentication layers would add different authentication information
- endpoint layers would add endoint urls, and associated information
- a final layer would then take the information from the request struct, make the request and return a response struct

Such an approach might be easier to test, but this is just me doing some random thinking.

begedin:
Note on transactions:

I'm unsure where to perform them, but it did occur to me that individual repo creation is something that will likely be done similarly with the `installation_repository` event, so that part of the code should probably be reusable/extractable.

lib/code_corps/github/events/installation.ex
```diff
@@ -32,80 +32,77 @@ defmodule CodeCorps.GitHub.Events.Installation do
...
+      installation
+      |> GitHub.Installation.repositories()
       |> (fn {:ok, repositories} -> repositories end).()
-      |> Enum.map(&create_repository(github_app_installation, &1))
+      |> Enum.map(&create_repository(installation, &1))
       |> Enum.map(fn {:ok, repository} -> repository end)
       |> (fn repositories -> {:ok, repositories} end).()
 
-    {:ok, github_app_installation |> Map.put(:github_repos, repositories)}
+    {:ok, installation |> Map.put(:github_repos, repositories)}
```
begedin: This was me trying to aggregate the results into something useful, but it isn't strictly necessary. Basically, everything after line 80 is unneeded. I'm fine with leaving it in, but also not sure about it causing confusion.

begedin (approved):
I have a comment, but overall, I think it looks good. Feel free to decide on that part at your discretion.

PR #840, opened 2017-07-05
Mock out pem access

### LINKED ISSUE #823: App fails to compile if no GITHUB_APP_PEM environment variable
# Problem

Without the mentioned key present in the environment, with any value whatsoever, the app fails to compile, both when running it, or when running tests.

This blocks new contributors from helping and forces us to either explicitly set the evn variable `source` the `.env` on every new tab to be able to do any work at all.

Additionally, adding an incorrect value to the env variable causes the app to compile, but then two tests fail:

```
  1) test authenticated_integration_request/5 handles an error response (CodeCorps.GitHubTest)
     test/lib/code_corps/github_test.exs:45
     ** (FunctionClauseError) no function clause matching in Joken.Signer.rs/2
     stacktrace:
       lib/joken/signer.ex:78: Joken.Signer.rs("RS256", [])
       (code_corps) lib/code_corps/github.ex:138: CodeCorps.GitHub.generate_jwt/0
       (code_corps) lib/code_corps/github.ex:157: CodeCorps.GitHub.add_jwt_header/1
       (code_corps) lib/code_corps/github.ex:230: CodeCorps.GitHub.authenticated_integration_request/5
       test/lib/code_corps/github_test.exs:46: (test)



  2) test authenticated_integration_request/5 handles a successful response (CodeCorps.GitHubTest)
     test/lib/code_corps/github_test.exs:39
     ** (FunctionClauseError) no function clause matching in Joken.Signer.rs/2
     stacktrace:
       lib/joken/signer.ex:78: Joken.Signer.rs("RS256", [])
       (code_corps) lib/code_corps/github.ex:138: CodeCorps.GitHub.generate_jwt/0
       (code_corps) lib/code_corps/github.ex:157: CodeCorps.GitHub.add_jwt_header/1
       (code_corps) lib/code_corps/github.ex:230: CodeCorps.GitHub.authenticated_integration_request/5
       test/lib/code_corps/github_test.exs:40: (test)

```

The compilation fails due to trying to `Base64.decode` a `nil` so that problem can be simply fixed by adding a `""` fallback to the variable in `config.ex`.

As for the failing tests, we need to decide what to do there. Do we skip them, or do we inject the data?

## Subtasks
- [ ] Make it so the variable is not strictly necessary
- [ ] Decide what to do with the failing tests

begedin:
Closes #823 

This PR adds a test fixture for a `.pem` file, used when authenticating a github request. It sets up the app configuration so that the `.pem` content is loaded from this fixture in the test environment, instead of being loaded from an environment variable, as it is in other cases.

This should Joken related GitHub tests to pass, for users who do not have the actual pem in their environment.

The test file simply contains a random RSA private key, since this is all we need for the process to work.


config/test.exs
```diff
@@ -54,3 +54,7 @@ config :code_corps,
 
 config :code_corps, :cloudex, CloudexTest
 config :cloudex, api_key: "test_key", secret: "test_secret", cloud_name: "test_cloud_name"
+
+# supply sample .pem with RSA_PRIVATE_KEY for tests
+config :code_corps,
+  github_app_pem: "./test/fixtures/github/app.pem" |> File.read!
```
joshsmith: We have an actual PEM on CI, so I wonder if this shouldn't be an `||` statement somehow.

joshsmith (approved):
Looks good generally. Approving so you don't have to wait for a follow-up, but would like that one comment handled.

Issue #833, opened 2017-06-29
Move repo creation in installation event to a transaction


joshsmith:
# Problem

Right now the repos are created without a transaction wrapping them. We should use Ecto's `multi` to wrap these in a transaction.

joshsmith:
Closed by #841.

Issue #830, opened 2017-06-28
Deal with the "unmatched_user" case of a github installation webhook


begedin:
NOTE: OP is mostly out of date. Refer to https://github.com/code-corps/code-corps-api/issues/830#issuecomment-314046480 for the actual requirements and description.

From https://github.com/code-corps/code-corps-ember/issues/1331 :
> `unmatched_user` - when we receive a webhook but there is no such Code Corps user with the given GitHub id. Edge case for now. This should only happen when the installation was initiated on GitHub. We can handle this by asking a user to connect with GitHub after the fact, and in the meantime matching the installation up to that user. Once the user has connected, we can move the status of the installation to initiated_on_github so they can pick a project.

How do we want to do this?. 

The relatively clear part is that, when a user connects with github, we need to see there are any unmatched installations of their in our database.

However, a github installation record in our database has no field which would connect it to a user.

Do we want to

- Add a `github_user_id` field to the `github_app_installations` table
- Fetch a list of github installations for the user from the API, then match any already in our database

I'm not finding an endpoint to do the second option, so the first one might be the only one available. That would mean that, in the case of an unmatched user, the installation needs to be stored with the github user id.

@JoshSmith What do you think?

joshsmith:
Yeah, this is something that I forgot to add in my notes last night. The first is the only workable option I can find. An `installation` _can_ be on a user's account, but an `installation` is not really _related_ to a user except by way of the user who installed it. This is intentional on GitHub's part: a person may leave a company, but they don't want that to mean the `installation` suddenly has to change "ownership". Instead, the `installation` is most typically performed on an organization level, which means anyone in the organization has access.

This is why we may need to rethink the relationships some.

begedin:
@JoshSmith Correct me if I'm wrong, but 

The `Installation::created` payload will have a "sender.id", which is what we currently take as to `github_id` to find the existing `User` by.

Could we not simply add a field to the installation -> `user_github_id` or `sender_id` and store that in the case of no user found.

Then, as a CodeCorps `User` connects with Github, they receive a `github_id`. We use it to see if there are any `GithubAppInstallationRecords` we can connect them to and if there are, associate them with the `User`.

The way I see it, if we use the field to match an installation with an already connected user, then it's no less wrong to use the field to match an installation with a freshly connected user.

If the field is not good enough for one case, it's not good enough for the other. Am I missing something?

joshsmith:
Are your questions already handled now? How out of date is this?

[...]

PR #841, opened 2017-07-05
Cleanup installation event

### LINKED ISSUE #838: Handle API error edge case when handling Installation::created event
# Problem

`mix dialyzer` reports a bunch of warnings when analyzing `lib/code_corps/github/event/installation.ex`. Most of these are due to us not handling the case of an API error when requesting installation repositories from the github API. 

Not handling this means the code will throw an error and the event will be kept in the "processing" state indefinitely. We should handle this case.

- Add a function clause for an API Error into the module
- Rewrite rest of module to account for that case - result should be event being marked as "errored"

begedin:
Closes #838 as well as #833. 

This PR makes the overall process of handling the `Installation::created` event more robust and far less prone to failure.

- Moves repo creation to a transaction
- Cleans up specs, solves minor issues with specs
- Adds points of failure for various cases 
  - issue with api
  - user found but installation not fund
  - action not yet supported
  - payload wrong in some way

In all of the above cases, before this PR, the event being handled would get stuck as "processing". Post this PR, it will correctly be put into the "errored" state, so we can properly handle it later.

This PR is built on top of #834 , so I'm basing it on that branch. Once #834 is merged, we should rebase off of `develop`

### Other minor things I did in order to get this working

- updated dialyxir (helped me with detecting failure paths)
- added missing types (for usage with dialyxir
- made the `GithubRepo` adapter explicitly fail if the structure does not match (it was failing with an exception previously)


begedin:
### UPDATE: This is not done yet. 

The PR improves upon previous behavior, so it is mergeable, but we should consider further alterations.

I think we've approached it incorrectly. The flow needs to be corrected and there are more cases we need to consider:

#### Corrected flow
try and find user
- [ ] if user not found, try and find installation by github id
  - [ ] if found, maybe update it? **not really sure how to handle**
  - [x] if not found, use installation attributes to create unmatched installation
    - [ ] right now, the way we create it is not very useful. it should be tied to a `sender_github_id` to be able to match it when, for example, a user connects with github and gets a `github_id`
     
- [ ] if user found, try and find installation by `github_id` and `user_id`
  - [ ] if found, update that one with repos - this should not actually happen, but we need to handle it
  - [ ] if not found, try and find installation associated to user and `github_id: nil` - this is the default and we do not currently do it this way
    - [ ] if found, update it with info from github, update repos
    - [ ] if not found, create one, associate with user, update repos

This should handle all cases, but I would not consider it a full, robust flow.

A user could still own multiple newly initiated installations (without a github id) on code corps. There is no guarantee only one will be found.

As you mentioned in #830, I think we need to reconsider our relationships here.

For my initial thoughts, it seems that we either need to consider integrations being enabled for users, not projects. Once they are, a user which is an owner of a project's organization can link a project with a repo. That means each user would have at most one github app installation.

An alternative would be to link installations with organizations, but a user can own multiple organizations, so I a have no idea how we would identify the correct organization upon receiving a webhook.

begedin:
Looking into this further, I'm still having difficulty with figuring out the proper procedure here.

# Perfect path

Starting on CodeCorps, the user goes to a project page and starts the integration. This creates a `GithubAppInstallation` with a `user_id`, `project_id` and a blank `github_id`.

Then user is directed to github, where they perform the installation, upon their github account.

- We get an "installation created" webhook, and handle it.
```Elixir
# there is exactly one user with the github id
%User{id: user_id} = user <- Repo.get_by(User, github_id: event["sender"]["id"]`
# there is no installation with the github id
nil <- Repo.get_by(GithubAppInstallation, github_id: event["installation"]["id"]
# there is exactly one installation belonging to user with no github id
%GithubAppInstallation{} = instalation <- Repo.get_by(GithubAppInstallation, user_id: user_id, github_id: event["installation"]["id"]
```

I this perfect path, we simply update the installation with 

```Elixir
%{installed: true, github_id: event["installation"]["id"]}
```

Then we set it's status to processing, get and create repos and set it to processed.

However, there's a whole bunch of unhappy paths

# Unhappy paths

## 1. There is no user, there is no app installation

This means someone found and installed the app directly from the github marketplace.

In this case, we can create the app installation, with the state "initiated_on_github". We should assign a `sender_github_id`, so we know which user did the install, once we potentially get that user on codecorps.

Later, when a user connects with github, if there is an installation with the proper `sender_github_id`, we associate it with that user. **However, there is no way for us to determine which project this happened for, so I'm not sure if there's a use in doing any of that.**

## 2. There is no user, there is an app with that `github_id`

This also means someone found and installed the app directly from the github marketplace.

I can only imagine this happening if 1 already happened and then somehow, the webhook triggered again.

## 3. There is a user, there is no app installation

A github user, which also happens to be registered on codecorps and is connected with github on codecorps, found the app on the marketplace and installed it.

I think we can handle this by creating the installation, filling out its data, associating with the user, etc. However, **there is no way for us to associate the installation with a specific project**.

## 4. There is a user, there is already an app installation with that `github_id`

Means the process was already previously done and the webhook was somehow triggered again. Either that, or the webhook triggered because the user installed the app to a different entity. For example, the previous installation was done to their own account, this one to an organization's account. This can happen and I'm not sure how to handle this.

I guess a user could have multiple installations and we need to store `event["installation"]["account"]["id"]`as well as `event["installation"]["account"]["type"]` onto the `GitHubAppInstallation`. Then, we need to figure out which one to work with.

## 5. There is a user, there is an app installation with a different `github_id` and an installation with no `github_id`, both belonging to that user.

I guess this is relatively safe. One went through the installation process fully, the other is the one we need to process now.

## 6. There is a user, there are multiple app installations on record with a blank `github_id`

I'm not sure it's conceivable to handle this one. 

This means that the user started connecting one project, then aborted the installation on github, then went to a different project and started the procedure again. They did this once or multiple times, then finally for one project, went through the installation procedure and we got a webhook.

We have no way of finding out which project the webhook is for. 

We can assume it's the most recent installation, but we cannot be sure of this.

# Proposal

Thinking long and hard about this one, this is what I got:

- We do not associate github app installations with a project
- We store `github_account_id`, `github_account_type`, `github_account_login`, `sender_github_id`, etc., onto the installation. Basically, informational fields which would allow us to render an app installation on the ember side.
- As the user goes to a project integrations page, they need to be
  - connected to github
  - have at least one installation
- At that point, they select installation, and then select from repos in that installation, to join with a project.
- They can also click something like "i would like to integrate with a different organization" or something, which takes them to the app installation page again, where they can install the app onto a different organization/their own account. They, can only do this if they, at the moment, have no "newly initiated installations" (meaning, installations which they've created from here, but have no `github_id` yet. Otherwise, instead, for that one installation without a github id, they get a button to "continue installing it".
- Once they pick a repo and connect it to a project, we can maybe lock that project into that installation, so they can only ever pick repos from one installation, but I'm not sure we have to.
- Alternatively, they could go to an organization integration page to select an installation from there, then go to a project page to select a repo from there

Either way,
- Unhappy path 1 is solved out of the box. We create the orphaned installation, use the `sender_github_id` to match with a user which connects later, create/update repos, etc.
- Path 2 is dealt with same as 1, except we update the installation instead of creating it.
- Path 3 is dealt with same as 1, except we now also associate the installation with the user 
- Path 4 is dealt with same as 3, except we update the installation instead of creating it
- Path 5 is now completely safe
- Path 6 can no longer happen

# NOTES:

- when updating repos, we need to do a "difference", meaning for that installation, we fetch all repos. Then we remove repos which we have a record of but they no longer exist, create repos which we have no record of and update the remaining repos.
- I'm sort of leaning towards the "organization integration" option. That way, we can potentially tie an organization to an installation similar to how github does it. All of that is something that does not need to happen initially, though. We can go with the simpler approach of just a 2-level selection, not locking in anything to anything and have the only relationship be `ProjectGithubRepo`. The way I see it, nothing should be stopping us from A project being synced with multiple repos across multiple installations. That being said, it might save us future trouble if we do lock in.

joshsmith:
A few notes here:

- The designs show an organization installation because that's the most straightforward way to connect up an installation to the person who has the privileges to do this in the first place – the organization owner.
- We _also_ want to track the requesting _organization_ (vs project, as we had been doing), if there was one, because that means the user doesn't have to go through an extra step of then connecting that organization to the installation once they return from GitHub, which is another pain in an already quite painful process for them.
- An app installation can happen without the user just browsing the GitHub marketplace. I shared a screenshot of example comments on GitHub and _anyone_ browsing _any_ GitHub repo that's hooked up to the Code Corps app could find themselves on the installation page. We have to handle that path not as an edge case but as a first class citizen in and of itself, _eventually_. I say eventually because handling that path in its totality in the best way involves more time than we have at present.

## 1. There is no user, there is no app installation

We could store a `sender_github_id`, though you can also find who installed it with the `installation` API endpoint, I believe. Better to store it, of course, but just pointing out that you're not SOL without it.

> However, there is no way for us to determine which project this happened for, so I'm not sure if there's a use in doing any of that.

This doesn't matter, per the designs I shared. The intent is that you could have an installation that _you the user have access to_ but that is not connected to the organization for that project. We'll want to be regularly refreshing the installations that a user has – as well as giving them a mechanism to refresh them manually from GitHub – so that we always know what installations are available to connect to an organization.

## 2. There is no user, there is an app with that github_id

> I can only imagine this happening if 1 already happened and then somehow, the webhook triggered again.

Yes, that's how that would happen, and should then be ignored as having already been processed.

## 3. There is a user, there is no app installation

> I think we can handle this by creating the installation, filling out its data, associating with the user, etc. However, there is no way for us to associate the installation with a specific project.

See my notes above on 1.

## 4. There is a user, there is already an app installation with that github_id

> Means the process was already previously done and the webhook was somehow triggered again. Either that, or the webhook triggered because the user installed the app to a different entity. For example, the previous installation was done to their own account, this one to an organization's account. This can happen and I'm not sure how to handle this.

No, this is not possible. An installation is a connection between a GitHub App and a GitHub account (user or organization). You cannot have an installation connect to multiple accounts.

## 5. There is a user, there is an app installation with a different `github_id` and an installation with no `github_id`, both belonging to that user.

> I guess this is relatively safe. One went through the installation process fully, the other is the one we need to process now.

Yes, this is how it ideally works when someone is creating multiple installations across their GitHub accounts.

## 6. There is a user, there are multiple app installations on record with a blank `github_id`

This should not happen at all. A blank `github_id` can only happen if initiated by a user. This happens when clicking the `Install on GitHub` button shown in the designs. If a user clicks that button and one already exists before we send them off to GitHub, we do not create another installation. It's a `find_or_create` operation.

joshsmith:
A link to the relevant designs for reference: https://projects.invisionapp.com/share/4NCEZ8EAF#/screens

begedin:
@JoshSmith Your approach with the designs and the explanation here are similar to what I proposed, albeit connecting things a bit differently. Still, I believe it does deal with the issues similarly and avoids the problems I mentioned. 

With that, I think I have enough information to make another pass through this, handle all the cases, and finally get this read for merge.

begedin:
@JoshSmith Pending review and merge of #834, this PR should be good to go as well.

Not all paths are fully covered, but none of them fails and they all do at least part of what they will likely end up doing. Here's what's covered:

- if any of the elements of the payload does not have the structure we expect, the event is marked as errored and nothing else is done
- if we find a user with the correct github id
  - if we find an installation with a blank github id -> update that installation, fetch and create repos (**this is our happy path**)
    - due to the way this works, if fetching and creation fails for some reason (api error, for example), the installation will still have been updated, and it will stay in the state `"initiated_on_code_corps"`. The refresh action will either have to try and fetch them again, or I can rework the process so the initial update is also part of the transaction.
  - if we do not find an installation with a blank github id, attempt to create one. If that works, event is processed. If that attempt fails, the event is marked as errored. It can fail if there is already that github id in our database, although that should not happen in a real-life scenario.
- if no user was matched
  - if no installation was matched by github id either, we create the installation, set it's `sender_github_id` and set it's `state` to `"initiated_on_github"`
  - if an installation was matched by `github_id`, we just update it. It should not actually happen unless we're not handling something, but the process will not fail if it does happen.

As I said, this ought to be good to go, but for future PRs, **a question**. 

** Should we fetch and create repos for all paths**, not only the happy one. 

If so, **should we also do a `diff` process on the repos**, meaning the fetched list is considered the authority, so we delete repos not on the list, create repos in the list but not present locally, update repos matched from the list?

As I said, both of these can be done in separate PRs if they are necessary.

joshsmith:
> **Should we fetch and create repos for all paths**, not only the happy one?

> If so, **should we also do a `diff` process on the repos**, meaning the fetched list is considered the authority, so we delete repos not on the list, create repos in the list but not present locally, update repos matched from the list?

Yes to both questions here.

joshsmith (approved):
Looks great. We'll create some follow-on issues now.

PR #844, opened 2017-07-10
Switch repo processing during the handling of an installation event to a diff based approach

### LINKED ISSUE #832: Ensure repos are created in other cases of app installation event and diff them
# Problem

Right now we only create the repos in one instance of the app installation, but we should really be creating them in all of the instances.

We should also do a `diff` process on the repos, meaning the fetched list is considered the authority. We should:

- delete an installation's repos that are not on the list
- create an installation's repos in the list but not present locally
- update an installation's existing repos matched from the list

begedin:
Closes #832 

# Updated Description

This PR does several things

- enabled diff based approach for github repo processing during an installation event
- revamped the `GithubAppInstallation` `:state` field. It can now be one of `~w(unprocessed processed processing errored)`
- added an `:origin` field to `GithubAppInstallation`. It can be one of `~w(codecorps github)`
- removed the update endpoint from `GithubAppInstallationController`, as well as all associated code. We discussed this and feel it's no longer necessary.
- updated API docs

# Old Description

This PR switches repo processing during the handling of an installation event to a diff based approach, using the list of repositories retrieved from the github API as the master list to base everything from.

I got this as far as I could and now need feedback. Several tests are failing, due to a simple reason.

- Installation event is received
- Installation is created or updated with `:state` set to `"unmatched_user"` or `"initiated_on_github"`, or it already exists as `"initiated_on_code_corps"`
- In the same thread/process, we start processing github repos
- In a transaction, installation `:state` is set to "processing"
- In the same transaction, repos are fetched from the API and synced with the local state
- In the same transaction, installation `:state` is set to `"processed"`

First, in my opinion, I think we have 3 different sets of information here.

- `"initiated_on_github"` and `"initiated_on_code_corps"`  respond to "where did the installation originate from
- `"unmatched_user"`, simply put, reponds to "was the user matched. The fact that `user_id` is nil provides the same information. An installation which has no user is actually also an installation `"initiated_on_github"`
- `"processed"` and `"processing"` are a temporary flag and respond to "what's happening right now"

In addition to this, `"processing"` provides us no usefulness whatsoever. Since the installation is set to `"processing"` and then also again to `"processed"` within the same transaction, the value `"processing"` is never actually commited to the database.

The way I see it

### Where did the installation originate from?

- an `:origin` field would make more sense for this
- all cases where we create an installation while handling a webhook should have the origin "github". 
- all cases where we create an organization from our controller, due to a user request, have the origin `"code_coprs"`

### Was the user matched

- if we actually do need an  `unmatched_user` status of some sort, instead of just checking if the `:user/:user_id` itself are `nil`, it would make more sense to have it on a view. It could be a boolean.

### What's happening to the record?

- the `:state` field should be used for this and its value should be one of `["processing", "processed"]` and possibly `"errored"`
- our `:repos` module should not handle everything in a single transaction. It should set status to processing, then fire off an async task to process the repos and, upon success, set it to `"processed"`. If we need this case, upon failure, it can also set it to `"errored"`

As it is right now, if we want to keep the single `:state` field, we would have to add additional values, which would be compound values such as `"processing_unmatched_user"`, etc.

Otherwise, the only outcome of the whole processing flow is that all installations will end up as `"processed"`. They will never be `"processing"` and they will for a few ms be one of `"initiated_on_github"`or `"unmatched_user"`. They might be `"initiated_on_code_corps"` for a brief moment longer, but only until we receive a webhook to handle.

@JoshSmith I could use your feedback on this.


joshsmith:
I agree with the logic here and support the changes.

lib/code_corps/github/event/installation/repos.ex
```diff
@@ -7,76 +7,155 @@ defmodule CodeCorps.GitHub.Event.Installation.Repos do
...
       Multi.new
-      |> Multi.update(:processing_installation, installation |> Changeset.change(%{state: "processing"}))
-      |> Multi.run(:process_repos, &process_repos/1)
-      |> Multi.run(:stop_processing, &stop_processing/1)
+      |> Multi.run(:processing_installation, fn _ -> {:ok, installation} end)
+      |> Multi.run(:api_response, &fetch_api_repo_list/1)
+      |> Multi.run(:repo_attrs_list, &adapt_api_repo_list/1)
+      |> Multi.run(:deleted_repos, &delete_repos/1)
+      |> Multi.run(:synced_repos, &sync_repos/1)
+      |> Multi.run(:processed_installation, &mark_processed/1)
```
begedin, line 49: Instead of doing a few large steps, which wrap several minor steps, I tried to do this as explicitly as I can.

lib/code_corps/github/event/installation/repos.ex
```diff
@@ -7,76 +7,155 @@ defmodule CodeCorps.GitHub.Event.Installation.Repos do
...
+
+  In the background, it fires a task to asynchronously call &process/1
+  """
+  @spec process_async(GithubAppInstallation.t) :: {:ok, GithubAppInstallation.t, Task.t}
+  def process_async(%GithubAppInstallation{} = installation) do
+    {:ok, %GithubAppInstallation{} = processing_installation} = installation |> set_state("processing")
+
+    task = Task.Supervisor.async(:background_processor, fn -> processing_installation |> process() end)
+
+    {:ok, processing_installation, task}
```
begedin, line 32: Returning the task is mostly in the service of  testing. We can use `Task.await/1` to resolve it in tests.

lib/code_corps/github/event.ex
```diff
@@ -8,29 +8,32 @@ defmodule CodeCorps.GitHub.Event do
...
-  def stop_processing({:error, _reason}, %GithubEvent{} = event) do
-    event |> Changeset.change(%{status: "errored"}) |> Repo.update
+
+  @spec do_stop_processing(list, GithubEvent.t) :: {:ok, GithubEvent.t}
+  defp do_stop_processing([:ok | _data], %GithubEvent{} = event), do: event |> set_status("processed")
+  defp do_stop_processing([:error | _reason], %GithubEvent{} = event), do: event |> set_status("errored")
+
+  @spec set_status(GithubEvent.t, String.t) :: {:ok, GithubEvent.t}
+  defp set_status(%GithubEvent{} = event, status) when status in ~w(processing processed errored) do
+    event |> Changeset.change(%{status: status}) |> Repo.update()
```
begedin, line 37: Changes here basically add support for varied size tuples as outcomes of handling an event payload. The only fixed rule is that the first element of the tuple needs to be `:ok|:error`

lib/code_corps/transition/github_app_installation_state.ex
```diff
@@ -1,32 +0,0 @@
...
-
-  def next("processing", "processed"), do: {:ok, "processed"}
-  def next("processing", "unmatched_user"), do: {:ok, "unmatched_user"}
-
-  def next("unmatched_user", "processing"), do: {:ok, "processing"}
-  def next("unmatched_user", "processed"), do: {:ok, "processed"}
-
-  def next(current_state, next_state) when current_state == next_state, do: {:ok, next_state}
-  def next(current_state, next_state), do: {:error, "invalid transition to #{next_state} from #{current_state}"}
-end
```
begedin, line 32: With the switch to status fields, the need for a state machine has been eliminated.

web/models/github_app_installation.ex
```diff
@@ -21,47 +24,31 @@ defmodule CodeCorps.GithubAppInstallation do
...
   """
   def create_changeset(struct, params \\ %{}) do
     struct
-    |> changeset(params)
+    |> cast(params, [:project_id, :user_id])
+    |> put_change(:state, "unprocessed")
+    |> put_change(:origin, "codecorps")
+    |> validate_required([:project_id, :user_id])
+    |> assoc_constraint(:project)
+    |> assoc_constraint(:user)
```
begedin, line 39: If I'm following things correctly, this changeset only gets called when creating from the client, in which case, the `state` and the `origin` need to have specific values. No need for a state_machine anymore, or anything of the sort. The `github_id` field is not set from here either.

web/models/github_app_installation.ex
```diff
@@ -21,47 +24,31 @@ defmodule CodeCorps.GithubAppInstallation do
...
 
+  @doc """
+  Changeset used to update a GithubAppInstallation record from CodeCorps
+  """
   def update_changeset(struct, params \\ %{}) do
     struct
-    |> changeset(params)
+    |> cast(params, [:project_id])
+    |> validate_required([:project_id])
+    |> assoc_constraint(:project)
```
begedin: I'm not sure we actually need an update like this.

With the introduction of `OrganizationGithubAppInstallation`, this changes anyway.

joshsmith (approved):

PR #848, opened 2017-07-13
Issues webhook handling

### LINKED ISSUE #828: Handle issues webhooks from GitHub
# Problem

We need to write code that handles `issues` events in the `CodeCorps.GitHub.Events.Issues` module.

This should:

- Have handlers for the `opened`, `edited`, `closed`, and `reopened` actions.
- Does not yet handle the `assigned`, `unassigned`, `labeled`, `unlabeled`, `milestoned`, or `demilestoned` actions.
- Fetches the issue from GitHub.
- Checks whether the `CodeCorps.Task` exists for the GitHub task.
- Checks whether this is the last write (i.e. another event from some source did not happen since this event transpired). If this is the last write, this write wins and we should create or update the `CodeCorps.Task`.

While we can handle this later, some thoughts on `assigned` and `unassigned` actions:

- We'll need to simply handle the assignment and unassignment of the `CodeCorps.Task` based on the GitHub account that's connected to a user. No thoughts yet on how to handle certain edge cases here. Thus, this should wait to be handled later

begedin:
Closes #828

This PR started out as a handler for the Issues "opened" webhook exclusively, but I managed to find a generalized solution for all 4 basic subtypes of this webhook. 

I also believe we will be able to extend this generalised solution for other subtypes in the future.

This is still not done, but there is an uncertainty we need to discuss, so I'm creating a PR for that purpose.

# What the handler does

- First, the handler tries to find a `GithubRepo` for the issue
- If no repo, `{:error, :unmatched_repo}` is returned, meaning the event itself will be marked as errored
- If a repo is found, it then for each `ProjectGithubRepo` child of the `GithubRepo` (found by `github_id`, `project_id`), it will
  - find or create a `Task` 
  - associate that task with a `User` (also found by github id)

What I'm unsure of is the following:

- ~~in cases other than the "opened" event, should we be creating a `Task` in case one was not found. Presumably, as the `Project` was connected to a `GithubRepo`, a sync was performed, so all tasks should be present, meaning this should not actually be occurring, but to code allows it, so I'm not enthusiastic about just ignoring it.~~ **Yes, we should be creating a task**
- ~~if we do not find a `User` what should we do? This is a realistic option and something we've discussed in the past, but we never agreed on a solution fully. Do we hold a CodeCorps user locally, to assign to the Task? Do we store the `user_github_id` as a Task field, to be joined later? Do we create a user, which can later be merged with a newly registered/connected user?~~ **We should be creating a user**
 
# TODO

- [x] Write tests for `StateManager`
- [x] Write tests for `Validator`
- [x] Update/write tests for `Handler`
- [x] Write tests for `ChangesetBuilder`
- [x] Deal with the unmatched user case
- [x] Deal with any changes for unmatched task


joshsmith:
> in cases other than the "opened" event, should we be creating a `Task` in case one was not found? Presumably, as the `Project` was connected to a `GithubRepo`, a sync was performed, so all tasks should be present, meaning this should not actually be occurring, but to code allows it, so I'm not enthusiastic about just ignoring it.

Yes, we should be creating tasks if they're not found. Ideally, of course, we have some mechanism to prevent actually _ingesting_ these events if the sync is still being performed. Bear in mind that a particularly large repository may take _several hours_ of processing time. Even our own repositories are at this stage. We should probably design with this expectation in mind. Luckily we have good test cases available to us in order to get this right.

> if we do not find a `User` what should we do? This is a realistic option and something we've discussed in the past, but we never agreed on a solution fully. Do we hold a CodeCorps user locally, to assign to the Task? Do we store the `user_github_id` as a Task field, to be joined later? Do we create a user, which can later be merged with a newly registered/connected user?

We create a user which can later be merged with a new user _or_ a newly connected-to-GitHub user. There are some tasks hanging around related to this, including the one to remove the database-level reliance on a user's email being present when signing up (GitHub does not necessarily provide us with this information; more often than not, it will not).

lib/code_corps/adapter/map_transformer.ex
```diff
@@ -0,0 +1,86 @@
+defmodule CodeCorps.Adapter.MapTransformer do
```
begedin, line 1: We were using this module as a Stripe util module, but the approach it uses is generic and can be reused for the purposes of GitHub adapters, so I moved it to outside of Stripe's scope.

I also did some renaming and added documentation to make it clearer what the functions in it actually do.

lib/code_corps/github/event/installation/repos.ex
```diff
@@ -141,18 +134,28 @@ defmodule CodeCorps.GitHub.Event.Installation.Repos do
...
   end
 
+  @spec changeset(GithubRepo.t, map) :: Changeset.t
+  defp changeset(%GithubRepo{} = github_repo, %{} = repo_attributes) do
+    github_repo
+    |> Changeset.change(repo_attributes)
+    |> Changeset.validate_required([
+      :github_id, :name, :github_account_id,
+      :github_account_avatar_url, :github_account_login, :github_account_type
+    ])
```
begedin, line 156: Instead of having the adapter implicitly doing validation of the repo payload, I switched to simply having the process (and the transaction) erroring out if there are any validation errors on any of the repos.

lib/code_corps/github/event/issues.ex
```diff
@@ -1,17 +1,132 @@
...
+      |> Multi.run(:repo, fn _ -> find_repo(payload) end)
+      |> Multi.run(:user, fn _ -> UserLinker.find_or_create_user(payload) end)
+      |> Multi.run(:tasks, &sync_all(&1, payload))
+
+    case Repo.transaction(multi) do
+      {:ok, %{tasks: tasks}} -> {:ok, tasks}
+      {:error, :repo, :unmatched_project, _steps} -> {:ok, []}
+      {:error, _errored_step, error_response, _steps} -> {:error, error_response}
+    end
+  end
```
begedin, line 79: Adding support for other subtypes of the `Issues` event, should be relatively simple from here. Most likely, it will be a matter of extending the behavior of the `ChangesetBuilder` module below.

lib/code_corps/github/event/issues.ex
```diff
@@ -1,17 +1,132 @@
...
+  defp do_handle(%{} = payload) do
+    multi =
+      Multi.new
+      |> Multi.run(:repo, fn _ -> find_repo(payload) end)
+      |> Multi.run(:user, fn _ -> UserLinker.find_or_create_user(payload) end)
+      |> Multi.run(:tasks, &sync_all(&1, payload))
+
+    case Repo.transaction(multi) do
+      {:ok, %{tasks: tasks}} -> {:ok, tasks}
+      {:error, :repo, :unmatched_project, _steps} -> {:ok, []}
```
begedin, line 76: We want the transaction to fail if there are no tasks to create/update due to unmatched projects. If it does not fail, we will end up creating a user for no reason. 

However, this transaction failure should not be interpreted as a failure to handle the webhook, so we translate it into a success with no data updated.

lib/code_corps/github/event/issues/changeset_builder.ex
```diff
@@ -0,0 +1,37 @@
...
+
+    task
+    |> Changeset.change(issue_attrs |> TaskAdapter.from_issue())
+    |> Changeset.put_change(:state, payload |> StateMapper.get_state())
+    |> MarkdownRendererService.render_markdown_to_html(:markdown, :body)
+    |> Changeset.put_change(:project_id, project_id)
+    |> Changeset.put_change(:user_id, user_id)
+    |> Changeset.validate_required([:project_id, :user_id, :markdown, :body, :title])
+    |> Changeset.assoc_constraint(:project)
+    |> Changeset.assoc_constraint(:user)
```
begedin, line 35: Likely, as we add new event subtypes for `Issues`, the actual changes will happen here. We may be creating `TaskUser` records alongside the main record, etc.

lib/code_corps/github/event/issues/user_linker.ex
```diff
@@ -0,0 +1,37 @@
...
+  defp find_user(%{"id" => github_id}) do
+    User |> Repo.get_by(github_id: github_id)
+  end
+
+  @spec create_user(map) :: {:ok, User.t}
+  defp create_user(%{} = user_attrs) do
+    %User{}
+    |> Changeset.change(user_attrs |> UserAdapter.from_github_user())
+    |> Repo.insert
+  end
```
begedin, line 36: If we want to introduce the concept of something like an `IdentityProvider` record linked to a user record, which is something we've mentioned in the past, that would likely affect the `UserLinker` module here.

lib/code_corps/github/event/issues/validator.ex
```diff
@@ -0,0 +1,20 @@
...
+  Returns `true` if all keys required to properly handle an Issue webhook are
+  present in the provided payload.
+  """
+  @spec valid?(map) :: boolean
+  def valid?(%{
+    "issue" => %{
+      "id" => _, "title" => _, "body" => _, "state" => _,
+      "user" => %{"id" => _}
+    },
+    "repository" => %{"id" => _}}), do: true
```
begedin, line 18: This function returns true if the payload possesses the minimum required data to handle the `Issues` webhook.

Basically this is our "first line of defense", to make sure nothing has changed with the Github API spec and the structure of the payload is as we expect it to be.

priv/repo/migrations/20170717092127_allow_user_email_username_null.exs
```diff
@@ -0,0 +1,17 @@
+defmodule CodeCorps.Repo.Migrations.AllowUserEmailUsernameNull do
+  use Ecto.Migration
+
+  def up do
+    alter table(:users) do
+      modify :email, :string, null: true
+      modify :username, :string, null: true
```
begedin, line 7: I could technically leave this one out and set the `username` to be the same as `github_username`, but I feel it just confuses things as we later merge a new/newly-connected user with this one.

begedin:
@joshsmith This should be read for a review and, hopefully, merge

# Final process:

- We recieve an issues payload. 
- If the subtype is any of `~w(opened closed edited reopened)` we handle it. The event will eventually be processed
- If the subtype is any of `~w(assigned unassigned labeled unlabeled milestoned demilestoned)` we do not handle it. The event will error out
- If the subtype is any other, we do not handle it, but the return value is different from previous. Not really important right now, but might be. Wanted to make it explicit. The event will error out.
- If the payload data does not contain the minimum needed to handle the event, the event will error out.

## When handling

- we try and find a github repo. If none is found, we error out, the event will error out
- we also find or create a user
- for the found github repo, if there are any `ProjectGithubRepo` children, for each of those we create or update a `Task` assigned to the found `User` and the associated `Project`
- the whole process is in a transaction, so if anything goes wrong, the event errors out and none of the records are created/updated

## Mapping of issue -> task

- Issue `state` is `open` or `closed`. This becomes task `status`
- Task `state` can be `published or `edited`. `Issue opened` becomes `published`. `closed`, `reopened`, `edited` all become `edited`
- Issue body is task markdown
- Task `body` gets generated using the same facilities we use when creating the task directly on cc

## Mapping of github user --> user

- `id` -> `github_id`
- `avatar_url` -> `github_avatar_url`
- `login` -> `github_username`

Not other mapping is performed. User is created without `username` or `email`.

lib/code_corps/adapter/map_transformer.ex
joshsmith: Have you tested this still works with Stripe locally?

lib/code_corps/github/adapters/github_repo.ex
```diff
@@ -1,22 +1,16 @@
...
-      github_account_type: github_account_type
-    }
+
+  @mapping [
+    {:github_id, ["id"]},
+    {:name, ["name"]},
+    {:github_account_id, ["owner", "id"]},
+    {:github_account_avatar_url, ["owner", "avatar_url"]},
+    {:github_account_login, ["owner", "login"]},
+    {:github_account_type, ["owner", "type"]}
```
joshsmith: Just for consistency's sake, would clean up alpha ordering here (and wherever possible).

lib/code_corps/github/event/issues.ex
```diff
@@ -1,17 +1,132 @@
...
+  alias Ecto.{Changeset, Multi}
+
+  @typep outcome :: {:ok, list(Task.t)} |
+                    {:error, :not_fully_implemented} |
+                    {:error, :unexpected_payload} |
+                    {:error, :unexpected_action} |
+                    {:error, :unmatched_repository}
+
+  @unimplemented_actions ~w(assigned unassigned milestoned demilestoned labeled unlabeled)
+  @implemented_actions ~w(opened closed edited reopened)
```
joshsmith: Might change ordering of these.

joshsmith (commented):
Forgot to submit this review until way late in the evening. Everything looks generally fine. Some small comments/questions here, but nothing preventing merge if addressed.

We do need to loop back again in #849, and I'm sure I just need to see this in action to understand all the possible edge cases here, but it looks like much of the main path is figured out here.

lib/code_corps/adapter/map_transformer.ex
begedin: I did some testing, but generally, the stripe adapter layer is already well-tested.

I went through it one more time and the only thing that was missing was a `from_params_test` for the `StripeConnectAccount` adapter, so I added that one.

PR #850, opened 2017-07-19
Issue comment webhook

### LINKED ISSUE #827: Handle issue_comment webhooks from GitHub
# Problem

We need to write code that handles `issue_comment` events in the `CodeCorps.GitHub.Events.IssueComment` module.

This should:

- Have handlers for the `created`, `edited`, and `deleted` actions.
- Fetches the comment from GitHub.
- Checks whether the `CodeCorps.Comment` exists for the GitHub comment.
- Checks whether the `CodeCorps.Task` exists for the GitHub issue. If not, we may need to handle this event later (once the issue is processed).
- Checks whether this is the last write (i.e. another event from some source did not happen since this event transpired). If this is the last write, this write wins and we should create or update the `CodeCorps.Comment`.

begedin:
Closes #827 

Implements handling for issue_comment "created", "edited" and "deleted" webhooks

# Handling of "created" and "edited" webhooks

This one is identical and works as follows

- Tries to match with a `GithubRepo` record. If it fails, event errors out
- Tries to match with the issue user, creates one if necessary
- Tries to match with the comment user, creates one if necessary 
- For each project associated with found `GithubRepo` record
  - creates or updates `Task`
  - creates or updates `Comment`

This is all done as part of a transaction, so if any part fails, nothing gets created/updated and the event errors out

# Handling of "deleted" webhook

Simply, in a single database action, deletes all comments with the specified github id.

# Task state problem

With the `StateMapper`, I've tried to find a general way of inferring a `Task` `:state` ("edited" or "published") from the github payload. 

I found it hard, almost impossible to find a mapping rule to follow that would make sense and actually be useful to us.

Instead, I just opted to mark newly created tasks as "published" and "edited" tasks as "edited".

With that being said, the field it self is not used on ember at all and I'm finding it hard to find a useful use case for it in general.

- We can always compare inserted_at/updated_at to see if the record is in it's original state, or if it has been edited
- From github, it's difficult to do anything either way, but if we want, we could always take te equivalent fields over from github
- Comments can also be edited, but there is no such field there
- We discussed adding "actions/events" to tasks, similar to how github has it. Those would reduce the usefulness of a "published"/"edited" state further

With that said, I would recommend we outright remove the state field. It complicates things and, as it is right now, provides no benefit whatsoever, since we do not use it in our code base at all, currently.

@JoshSmith , what do you think?


joshsmith:
👍 on task state.

begedin:
The coverage drop reported here is incorrect. It's listing drops in files untouched by this PR

PR #855, opened 2017-07-25
Support for retrieving GithubAppInstallations available to a User

### LINKED ISSUE #853: Add available GithubAppInstallations to a user
# Problem

We need to find a way to add available `GithubAppInstallations` for a `User` that the user has access to for the GitHub App.

We'll have to refresh this data by hitting [GitHub's API endpoint](https://developer.github.com/v3/apps/#list-installations-for-user) for listing a user's installations.

We may want an endpoint for this at our own `/user/github-app-installations`. We could also consider having this simply be available to a `User` on the `UserView`, but this may confuse things with an existing `GithubAppInstallations` relationship. We could perhaps rename the key to something like `available-github-app-installations` but keep the type `github-app-installation`.

Alternatively, we could also rename the `user` relationship in `GithubAppInstallation` to `sender`, which is more in-line with GitHub's own API and can reduce some confusion there.

begedin:
Closes #853 when done

This PR should implement the feature described in #853.

However, we haven't really agreed on how we want to approach it exactly. Will it be a special endpoint, or a new `User` has_many relationship, etc?

In any case, the parts implemented should stay about as is. The new adapter function should sit alongside the one defined in #854, and I've already written down and commented out the fields added in that PR, so likely, we will have to merge that one first.


joshsmith:
@begedin what's the status of this given changes we've implemented since?

begedin:
@JoshSmith I believe parts of the code here can still be used, but parts need to be dropped. Instead of the current approach of fetching and syncing, we should make the index action on github app installation controller be scoped to the user, meaning

1. We fetch a list of installations for the user from github
2. We load and render those records from our database, whose github ids match the ids of the installations we got from github

That means we likely need to new github api endpoint and there's still potential use for the adapter (in that we can refactor parts of our event handling code to make use of it), but we do not need the installation syncer.

joshsmith:
Is it still much of a priority right now, or is it more for the use of our still open Ember PR with the more complex behavior we intend to add in the future?

begedin:
The Ember PR will work without it. Presumably, the index route will simply eventually return less records than it does now.

Issue #518, opened 2016-12-03
Create an OrganizationInvite model


joshsmith:
# Problem

We need to have an `OrganizationInvite` model for inviting organizations.

## Subtasks
- [ ] Create a migration
- [ ] Generate a short, unique code that can be used to fulfill the invite
- [ ] Email for inviting the person
- [ ] Organization title for use in sending an invite email and pre-populating the create form when the invite is consumed
- [ ] A`fulfilled` boolean that can only move from `false` to `true`
- [ ] Index on the `code` and `email`
- [ ] Write unit tests

snewcomer:
@JoshSmith looks like this is needed first?

PR #856, opened 2017-07-27
Added model Organisation invites along with unit tests

vishaldeepak:
# Solves issue #518 - Create OrganizationInvite Model

Created OrganisationInvite model. All fields as requested in the issue. 
Two parts of the code that should be looked on is 
 - Method `check_fulfilled_changes_to_true` which makes sure that field `fulfilled` only changes from false to true
 - Method `generate_code` - Which generates a 10 digit unique code using `:crypto`

### DIFF
--- a/web/models/organization_invite.ex
+++ b/web/models/organization_invite.ex
@@ -1,0 +1,54 @@
+defmodule CodeCorps.OrganizationInvite do
+  @moduledoc """
+  Handles inviting organizations via email
+  """
+
+  use CodeCorps.Web, :model
+
+  schema "organization_invites" do
+    field :code, :string
+    field :email, :string
+    field :title, :string
+    field :fulfilled, :boolean, default: false
+
+    timestamps()
+  end
+
+  @doc """
+  Builds a changeset based on the `struct` and `params`.
+  """
+  def changeset(struct, params \\ %{}) do
+    struct
+    |> cast(params, [:email, :title, :fulfilled])
+    |> validate_required([:email, :title])
+    |> validate_format(:email, ~r/@/)
+    |> validate_change(:fulfilled, &check_fulfilled_changes_to_true/2)
+  end
+
+  @doc """
+  Builds a changeset for creating an organization invite.
+  """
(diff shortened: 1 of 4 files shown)


web/models/organization_invite.ex
```diff
@@ -0,0 +1,54 @@
...
+    struct
+    |> changeset(params)
+    |> generate_code
+  end
+
+  defp generate_code(changeset) do
+    case changeset do
+      %Ecto.Changeset{valid?: true}  ->
+        length = 10
+        code = :crypto.strong_rand_bytes(length) |> Base.encode64 |> binary_part(0, length)
```
vishaldeepak, line 41: The issue mentioned a short key. I took the short length to be 10. 
Also I assumed the `:crypto.strong_rand_bytes(length)` would be random enough that collisions won't occur with other `code` values in the database. Field `code` has a unique index on it.  In case it needs to checked, would it be better  to keep generating code until a unique string is found, that does not clash with the database value of field?

begedin:
This is nicely self-contained and looks good. I think it's good to merge, so as soon as you resolve the conflict, I can merge it 👍 

Feel free to resolve simply by using your version of `structure.sql` as the correct one. The file in no way affects the rest of the code, so that's the quickest way to resolve.

begedin (changes requested):
Just resolve the conflict and this is good to merge. Thank you for your work!

vishaldeepak:
Changed as requested.

begedin (approved):

hl:
hi @vishaldeepak, great work!

I do have some small suggestions, mainly to do with pattern matching.

I've create https://gist.github.com/hl/e8da9adeeb54c9d05f4057b2f357e747 for you to check.

vishaldeepak:
@hl that makes the code even cleaner. Maybe you can raise a pull request, or would you want me make the changes.
I definitely need to go back to elixir style guide once more 😃

hl:
@vishaldeepak go ahead and change it, make sure you tested, I've only made a gist and am not sure if it will compile

PR #859, opened 2017-07-31
Associate connected user with installations

### LINKED ISSUE #858: When a CodeCorps user connects with github, associate it with any existing GithubAppInstallatons
# Problem

When handling an `Installation::created` webhook, there are cases where our app could be installed from github, without the user getting there from codecorps. Such an installation might match with a user, if there's a github connected user on CodeCorps with that github id. 

However, it also might not match with a user at all. 

In that case, the GithubAppInstallation record will be created without a User association, but with a `sender_github_id` field.

Later on, when a CodeCorps `User` connects with github, they will be assigned a `github_id`. We should, at that point, associate all orphaned `GithubAppInstallations` with that `sender_github_id` with the newly connected user.

begedin:
Closes #858 

This is a first step in getting to an MVP for the project integration tab.


web/models/user.ex
```diff
@@ -103,12 +103,6 @@ defmodule CodeCorps.User do
     |> apply_state_transition(struct)
   end
 
-  def github_association_changeset(struct, params) do
-    struct
-    |> cast(params, [:github_auth_token, :github_avatar_url, :github_email, :github_id, :github_username])
-    |> validate_required([:github_auth_token, :github_avatar_url, :github_id, :github_username])
-  end
```
begedin, line 110: No need to expose it separately, since it's a simple thing. Should be enough to test the module calling this behaviour.

PR #862, opened 2017-08-08
Upgrade to phoenix 1 3 final

### LINKED ISSUE #749: Upgrade to Phoenix 1.3
# Problem

[Phoenix 1.3.0-rc.1 is out](https://github.com/phoenixframework/phoenix/blob/master/CHANGELOG.md#130-rc1-2017-03-14) and we should consider starting a branch to track these changes.

Chris McCord has written up some [upgrade instructions here](https://gist.github.com/chrismccord/71ab10d433c98b714b75c886eff17357) that we can use as a basis for figuring all of this out.

begedin:
Makes obsolete and closes #782, closes #749 

This PR performs a full update to phoenix 1.3, but does not apply any major code refactorings which phoenix 1.3 would encourage.

Primarily, it does not fully eliminate our models, nor does it attempt to make a switch to the bounded context approach. I believe that part would be extremely out of scope of this PR and should be discussed thoroughly, before it is attempted.

To that purpose, I will create an issue and link it here as well as in #749 - Issue created #863 

# What this PR does

- updates phoenix to 1.3, as vell as several other dependencies to the latest versions
- moves `/web` to `/lib/code_corps_web` and introduces the `CodeCorpsWeb` namespace
- moves `/web/models` into `lib/code_corps/model` as the first step to introducing bounded context. Note that the namespace for models has *not* been changed from `CodeCorps.X` to `CodeCorps.Model.Ex`. That would serve little purpose and would just increase the complexity of this PR. Our aim is to get rid of the folder entirely and organize our schemas into contexts.
- moves `/web/policies` to `lib/code_corps/policy`, again, as an initial step towards bounded contexts
- moves several other minor modules to better-fitting locations
- cleans up `mix.exs`
- other minor changes and fixes to ensure everything works as intended


paulsullivanjr:
Ship it!

lib/code_corps_web/controllers/project_category_controller.ex
```diff
@@ -11,6 +11,9 @@ defmodule CodeCorps.ProjectCategoryController do
   plug :load_and_authorize_resource, model: ProjectCategory, only: [:delete]
   plug JaResource
 
+  @spec model :: module
+  def model, do: CodeCorps.ProjectCategory
```
snewcomer, line 15: Just curious...what are these changes for?

snewcomer (commented):
This looks g2g!

lib/code_corps_web/controllers/project_category_controller.ex
begedin: Good question. Was gonna comment on that myself to elaborate.

We use the `ja_resource` package in our controllers, to auto-load resources. 

It's been getting less and less useful for us, though, and this is just another case where more pain and no benefit is added.

`ja_resource` auto loads by assuming an `id` in the params. That part works in most of our cases, except in the case where we use slugs.

It also assumes/infers the model name from the controller name, so in this specific case, it will assume 

`CodeCorpsWeb.ProjectCategoryController` -> `CodeCorpsWeb.ProjectCategory`.

When the only namespace was `CodeCorps`, this worked. Now, it does not anymore, so we have to override the `model` function provided to us by `ja_resource` to get it working again.

With this one more thing we have to write in our controller and the fact that phoenix 1.3 wants to get away from models and into contexts, as well as the addition of the `FallbackController`, plus several other problems `ja_resource` is creating for us, I'm going to recommend we ditch the package completely.

Issue #864, opened 2017-08-09
Remove ja_resource


begedin:
# Problem

`ja_resource` has been giving us trouble before and will likely continue to give us more trouble.

Our initial reasoning for introducing it to our project was to reduce our controller boiler plate. However, with the way our other tools developed, this benefit has been gradually reducing and we now barely have any less boilerplate than before, and the cost has been reduced readability of code, as well as performance, in some cases.

- we have to override several hooks per controller, due to using custom, non-basic behavior
- it clashes heavily with our authorization layer, in some places causing us to load the same resource twice, before authorizing and performing the action
- it makes it hard to switch to a "better" authorization framework. We've been discussing switching away from canary in the past and `ja_resource` was the main blocker in doing so.
- the switch to Phoenix 1.3 means several things that further reduce the usefulness of `ja_resource`
  - our controllers are now in the `CodeCorpsWeb` namespace, while models are in `CodeCorps`, so each controller needs to override the `ja_resource` `model/0` function to get it working
  - the aim is to move away from the model concept and into contexts and `ja_resource` depends heavily on the model concept to generally work
  - with the introduction of `FallbackController` and action fallbacks, we can reduce boilerplate that way
- the library itself is not updated often, so we're bound to encounter more issues as time goes on

# Conclusion

With all that in mind, my recommendation is to get rid of `ja_resource` completely. This would involve...

# Requirements

- rewriting our controller code to the old action system
- make use of fallback actions and `FallbackController` to reduce the boilerplate that way
- make use of our own, custom plugs to reduce boilerplate further
- slight adjustment of our authorization plugs

# Notes

The advantages are

- less issues down the line
- increased flexibility
- improved code clarity and explicitness

*On a personal note*, it's my opinion that we introduced `ja_resource` early in our elixir experience when we still tried to make it work sort of like ruby. Now that I have more experience with the language and the phoenix framework, the negatives, to me, seem to outweigh the positives.

With a simple REST api, it might still work, but with the things we do here, I don't think it does.

begedin:
To add, we can do this gradually. Switch one or two major controllers to it as examples, then volunteers can take it over bit by bit. Once everything is switched, we remove the package from our dependencies.

snewcomer:
I'd be interested in helping out here.

begedin:
@snewcomer This is still at a "suggestion" level, but if we go through with it, your help is much appreciated.

joshsmith:
[...]

PR #867, opened 2017-08-18
Switched comment controller away from ja_resource

begedin:
Progress on #864 

# What's in this PR?

This PR contains a mergeable example of a complete switch away from `ja_resource` towards using the facilities provided by `action_fallback` available in Phoenix 1.3

The general philosophy is explicitness. In each controller action we

- convert params to attributes
- load resources (falls back to rendering a 404 if resource is not found)
- authorize action on resource (falls back to rendering a 403 if not authorized)
- perform action (falls back to 422 if there are validation errors)

Note that in some cases, we instead authorize an action on a changeset, so the changeset is initialized first. This was troublesome for us before as well, but seems less so this time around.

The changes here remove all reliance on `ja_resource` from the comment controller, as well as reliance on the canary library for authorization, since all that library really provides is it's own set of `load_resource`, `authorize_resource`, etc. plugs. It still keeps policies around, but they are moved outside the scope of a `Canada.Can` implementation.

A big advantage is that, since `ja_resource` isn't there to do it's own loading anymore, we effectively almost halved the amount of querying the do. 

Sadly, the `action_fallback` does have a limitation - it does not fallback out of plugs, only from the controller action itself. That means that, while full reliance on plugs would reduce controller boilerplate further, it would also effectively put us in the same rigid system we have now.

I feel the changes here, while increasing the number of lines of code slightly
- boost performance (no double-queries)
- improve code readability and explicitness

# Some progressive enhancements we can do

- a general plug which performs the `JaSerializer.Params.to_attributes(data)`step. We could call it a `json_api` plug
- a fallback for rendering the success json itself. The create action is problematic there, since we also need to add a `201` status to the conn

### DIFF
--- a/lib/code_corps/helpers/query.ex
+++ b/lib/code_corps/helpers/query.ex
@@ -2,7 +2,12 @@ defmodule CodeCorps.Helpers.Query do
   import CodeCorps.Helpers.String, only: [coalesce_id_string: 1]
   import Ecto.Query, only: [where: 3, limit: 2, order_by: 2]
 
-  def id_filter(query, id_list) do
+  @spec id_filter(Ecto.Queryable.t, map | String.t) :: Ecto.Queryable.t
+  def id_filter(query, %{"filter" => %{"id" => id_csv}}) do
+    query |> id_filter(id_csv)
+  end
+  def id_filter(query, %{}), do: query
+  def id_filter(query, id_list) when is_binary(id_list) do
     ids = id_list |> coalesce_id_string
     query |> where([object], object.id in ^ids)
   end
--- a/lib/code_corps/policy/policy.ex
+++ b/lib/code_corps/policy/policy.ex
@@ -8,6 +8,25 @@ defmodule CodeCorps.Policy do
   alias CodeCorps.Policy
   alias Ecto.Changeset
 
+  @doc ~S"""
+  Determines if the specified user can perform the specified action on the
+  specified resource.
+
+  The resource can be a record, when performing an action on an existing record,
+  or it can be a map of parameters, when creating a new record.
+  """
+  @spec authorize(User.t, atom, map | struct) :: {:ok, :authorized} | {:error, :not_authorized}
+  def authorize(%User{} = user, action, resource) do
+    case user |> can?(action, resource) do
+      true -> {:ok, :authorized}
+      false -> {:error, :not_authorized}
+    end
+  end
+
+  @spec can?(User.t, atom, map | struct) :: boolean
+  defp can?(%User{} = user, :update, %Comment{} = comment), do: Policy.Comment.update?(user, comment)
+  defp can?(%User{} = user, :create, %{} = params), do: Policy.Comment.create?(user, params)
+
   defimpl Canada.Can, for: User do
     # NOTE: Canary sets an :unauthorized and a :not_found handler on a config level
     # The problem is, it will still go through the authorization process first and only call the
@@ -22,9 +41,6 @@ defmodule CodeCorps.Policy do
 
     def can?(%User{} = user, :create, Category), do: Policy.Category.create?(user)
     def can?(%User{} = user, :update, %Category{}), do: Policy.Category.update?(user)
-
-    def can?(%User{} = user, :create, %Changeset{data: %Comment{}} = changeset), do: Policy.Comment.create?(user, changeset)
-    def can?(%User{} = user, :update, %Comment{} = comment), do: Policy.Comment.update?(user, comment)
 
     def can?(%User{} = user, :create, %Changeset{data: %DonationGoal{}} = changeset), do: Policy.DonationGoal.create?(user, changeset)
     def can?(%User{} = user, :update, %DonationGoal{} = comment), do: Policy.DonationGoal.update?(user, comment)
--- a/lib/code_corps_web.ex
+++ b/lib/code_corps_web.ex
@@ -32,7 +32,6 @@ defmodule CodeCorpsWeb do
     quote do
       use Phoenix.Controller, namespace: CodeCorpsWeb
 
-      alias CodeCorps.Repo
       import Ecto
       import Ecto.Query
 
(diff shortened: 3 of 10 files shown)


snewcomer:
@begedin this looks great and definitely like the clarity this brings!  Where was the double querying happening?

begedin:
> @begedin this looks great and definitely like the clarity this brings! Where was the double querying happening?

Basically hidden from plain sight

`load_resource` provided by canary would perform loading. Then `ja_resource` would also do automatic loading of the same resource.

snewcomer:
@begedin I'm sure we are waiting on @JoshSmith but I can help once this is merged!

lib/code_corps/policy/policy.ex
```diff
@@ -8,6 +8,17 @@ defmodule CodeCorps.Policy do
...
+  @spec authorize(User.t, atom, struct) :: {:ok, :authorized} | {:error, :not_authorized}
+  def authorize(%User{} = user, action, resource) do
+    case user |> Policy.can?(action, resource) do
+      true -> {:ok, :authorized}
+      false -> {:error, :not_authorized}
+    end
+  end
+
+  def can?(%User{} = user, :update, %Comment{} = comment), do: Policy.Comment.update?(user, comment)
+  def can?(%User{} = user, :create, %Changeset{data: %Comment{}} = changeset), do: Policy.Comment.create?(user, changeset)
```
begedin: We do not rely on canary here, so we do not scope it within `defimpl`

lib/code_corps/helpers/query.ex
```diff
@@ -2,7 +2,12 @@ defmodule CodeCorps.Helpers.Query do
   import CodeCorps.Helpers.String, only: [coalesce_id_string: 1]
   import Ecto.Query, only: [where: 3, limit: 2, order_by: 2]
 
-  def id_filter(query, id_list) do
+  @spec id_filter(Ecto.Queryable.t, map | String.t) :: Ecto.Queryable.t
+  def id_filter(query, %{"filter" => %{"id" => id_csv}}) do
+    query |> id_filter(id_csv)
+  end
+  def id_filter(query, %{}), do: query
+  def id_filter(query, id_list) when is_binary(id_list) do
```
begedin, line 10: This whole thing has potential to be rewritten in a recursive fashion, where we accept a map, then pass the queryable through a pipe of clauses, each supporting a single key, applying a filter from that key, then calling the next clause with a map where that key was removed. We could even split our queries apport into context-specific modules, for example, `Common.Query`, `Comment.Query`, etc.

lib/code_corps/policy/policy.ex
```diff
@@ -8,6 +8,25 @@ defmodule CodeCorps.Policy do
...
+  def authorize(%User{} = user, action, resource) do
+    case user |> can?(action, resource) do
+      true -> {:ok, :authorized}
+      false -> {:error, :not_authorized}
+    end
+  end
+
+  @spec can?(User.t, atom, map | struct) :: boolean
+  defp can?(%User{} = user, :update, %Comment{} = comment), do: Policy.Comment.update?(user, comment)
+  defp can?(%User{} = user, :create, %{} = params), do: Policy.Comment.create?(user, params)
```
begedin, line 28: As we switch controllers away from ja_resource, we also switch policies away from `Canary`. They also should no longer accept changesets and instead simply accept params maps.

lib/code_corps_web.ex
```diff
@@ -41,6 +40,9 @@ defmodule CodeCorpsWeb do
 
       import Canary.Plugs
       import CodeCorpsWeb.AuthenticationHelpers, only: [load_and_authorize_changeset: 2]
+
+      alias CodeCorps.{Repo, Policy}
+      alias Plug.Conn
```
begedin, line 45: These will presumably be used in ally controllers, once we make the full switch so no reason not to include them here.

lib/code_corps_web/controllers/comment_controller.ex
```diff
@@ -1,30 +1,41 @@
 defmodule CodeCorpsWeb.CommentController do
   use CodeCorpsWeb, :controller
-  use JaResource
 
-  import CodeCorps.Helpers.Query, only: [id_filter: 2]
+  alias CodeCorps.{Comment, User, Helpers.Query}
 
-  alias CodeCorps.Comment
+  action_fallback CodeCorpsWeb.FallbackController
+  plug CodeCorpsWeb.Plug.DataToAttributes
```
begedin, line 7: We can move this  one into the router `:api` pipeline once everything is fully switched.

lib/code_corps_web/controllers/comment_controller.ex
```diff
@@ -1,30 +1,41 @@
 defmodule CodeCorpsWeb.CommentController do
   use CodeCorpsWeb, :controller
-  use JaResource
 
-  import CodeCorps.Helpers.Query, only: [id_filter: 2]
+  alias CodeCorps.{Comment, User, Helpers.Query}
 
-  alias CodeCorps.Comment
+  action_fallback CodeCorpsWeb.FallbackController
```
begedin, line 6: This could go into the `code_corps_web.ex` `using` block once we make the full switch.

Any additional fallbacks can be put into the `FallbackController` module if they are used across controllers, or into the current controller module as a function, if specific to this controller, since this is just a plug anyway.

begedin:
I feel this has been sufficiently reviewed and well documented here as well as in #864, so I'm merging this.

joshsmith:
@snewcomer just bringing to your attention that this is merged!

lib/code_corps_web/plugs/data_to_attributes.ex
```diff
@@ -0,0 +1,21 @@
+defmodule CodeCorpsWeb.Plug.DataToAttributes do
+  @moduledoc """
+  Puts authenticated Guardian user into conn.assigns[:current_user]
```
snewcomer, line 3: @begedin Is this moduledoc correct?

begedin: @snewcomer It was absolutely not. I'll do a quick correction PR

begedin: Fixed in #884

PR #871, opened 2017-08-30
Fix github-app-installations relationship to be available on user view

joshsmith:
# What's in this PR?

This fixes the `UserView` to have `github-app-installations` relationships available on it, otherwise the Ember app does not have any unconnected installations to display for the user.

Also adds a test for this behavior.


begedin (approved):

begedin:
Good catch. Figured I'd miss something.

PR #875, opened 2017-08-30
Refactored GitHub api layer for easier and more flexible testing

begedin:
# What's in this PR?

This PR ought to perform the refactor of the github layer for easier and more flexible mocking.

It cleans up the `CodeCorps.Github` namespace, adds a mocking macro and infrastructure and refactors code so the part that needs to be mocked is minimal.

It also starts converting tests from bypass to using the mocking macro, with some examples already in, and a bunch left to do.

Ideally, what we'll end up with is most test using the default mock module, some specific tests using custom modules and the only place bypass will be used is in `tests/lib/codecorps/github/api_test.exs`

I'll comment on the code as soon as I can.


lib/code_corps/github/api.ex
```diff
@@ -0,0 +1,34 @@
...
+  defp marshall_response({:ok, status, _headers, body}) when status in 400..599 do
+    case body |> Poison.decode do
+      {:ok, json} -> {:error, APIError.new({status, json})}
+      {:error, error} -> {:error, error}
+    end
+  end
+  defp marshall_response({:error, reason}) do
+    {:error, HTTPClientError.new(reason: reason)}
+  end
+end
```
begedin, line 41: This should be the only place where requests to github are actually made, and even then, we should keep it as flat as simple as possible. 

We need to write a test module for this file to, and it should be the only one to use bypass.

lib/code_corps/github/bypass.ex
```diff
@@ -1,12 +0,0 @@
...
-  def setup(bypass) do
-    Application.put_env(:code_corps, :github_oauth_url, "http://localhost:#{bypass.port}")
-    Application.put_env(:code_corps, :github_base_url, "http://localhost:#{bypass.port}/")
-  end
-
-  def teardown do
-    Application.delete_env(:code_corps, :github_base_url)
-    Application.delete_env(:code_corps, :github_oauth_url)
-  end
-end
```
begedin, line 12: Since it's only used from tests, I figured it makes more sense to place it inside the tests folder

lib/code_corps/github/github.ex
```diff
@@ -1,6 +1,5 @@
 defmodule CodeCorps.GitHub do
```
begedin: This file is split out a bit. Parts have been moved to `github/api.ex` and parts to `github/headers.ex`. Both these files probably need some form or testing, api more so.

lib/code_corps/github/installation.ex
```diff
@@ -61,7 +61,7 @@ defmodule CodeCorps.GitHub.Installation do
   def refresh_token(%GithubAppInstallation{github_id: installation_id} = installation) do
     endpoint = "installations/#{installation_id}/access_tokens"
     with {:ok, %{"token" => token, "expires_at" => expires_at}} <-
-           GitHub.authenticated_integration_request(%{}, :post, endpoint, %{}, []),
+           GitHub.integration_request(:post, endpoint, %{}, %{}, []),
```
begedin, line 64: I went with the `method, url, headers, body, options` order of arguments for all request functions where it applies. My reasoning is that `:hackney` uses the same ordering.

test/lib/code_corps/github/event/installation/repos_test.exs
```diff
@@ -16,20 +16,38 @@ defmodule CodeCorps.GitHub.Event.Installation.ReposTest do
...
+      payload =
+        "installation_repositories"
+        |> GitHubCase.Helpers.load_endpoint_fixture
+        |> Map.put("repositories", [%{}])
+      {:ok, payload}
+    end
+    def request(method, endpoint, headers, body, options) do
+      GitHubCase.SuccessAPI.request(method, endpoint, headers, body, options)
+    end
+  end
```
begedin: This is an example of using mock modules to replace the api layer temporarily for specific tests.

The code being tested here makes two requests and the tests specify what happens when one of them fails in a specific way. 

We've defined modules for those two failure cases where the method signature which ought to fail, fails in that specific way, while all other method signatures _delegate_ to the default `GitHubCase.SuccessAPI` module.

test/lib/code_corps/github/event/installation/repos_test.exs
```diff
@@ -93,25 +107,23 @@ defmodule CodeCorps.GitHub.Event.Installation.ReposTest do
...
-      "/installation/repositories" => {403, @forbidden},
-      "/installations/#{@app_github_id}/access_tokens" => {200, @access_token_create_response}
-    }
     test "returns installation as errored if api error" do
       installation = insert(:github_app_installation, github_id: @app_github_id, state: "initiated_on_code_corps")
-      {:error, %GithubAppInstallation{state: state}, %CodeCorps.GitHub.APIError{}} = Repos.process(installation)
+
+      with_mock_api(BadRepoRequest) do
+        {:error, %GithubAppInstallation{state: state}, %CodeCorps.GitHub.APIError{}} = Repos.process(installation)
+      end
```
begedin, line 109: This is how we use the macro to mock out an api request.

With that done, the bypass tag is no longer needed. Our goal in this PR is to remove all existing bypass tags and switch to `with_mock_api`. 

While doing that, we will fill out the default `SuccessAPI` module with default responses.

Something to consider is that it may also make sense to define a default `FailureAPI` if we see enough code being repeated.

test/lib/code_corps_web/controllers/github_events_controller_test.exs
```diff
@@ -26,17 +26,6 @@ defmodule CodeCorpsWeb.GitHubEventsControllerTest do
...
-  @access_token_create_response %{"token" => @access_token, "expires_at" => @expires_at}
-
-  @installation_created load_event_fixture("installation_created")
-  @installation_repositories load_endpoint_fixture("installation_repositories")
-
-  @tag bypass: %{
-    "/installation/repositories" => {200, @installation_repositories},
-    "/installations/#{@installation_created["installation"]["id"]}/access_tokens" => {200, @access_token_create_response}
-  }
   test "responds with 200 for a supported event", %{conn: conn} do
```
begedin, line 26: This is an example of simply taking away the `bypass` tag. The default `SuccessAPI` is enough to keep the tests passing.

test/support/github_case/success_api.ex
```diff
@@ -0,0 +1,55 @@
...
+    send(self(), {method, url, headers, body, options})
+
+    {:ok, mock_response(method, url, headers, body, options)}
+  end
+
+  defp mock_response(:post, "https://github.com/login/oauth/access_token", _, _, _) do
+    %{"access_token" => "foo_auth_token"}
+  end
+  defp mock_response(method, "https://api.github.com/" <> endpoint, headers, body, options) do
+    mock_api_response(method, endpoint |> String.split("/"), headers, body, options)
```
begedin: There can be dynamic parts to a request url, so this is how we deal with those. We can't pattern match a string when a dynamic part is not the last part of the string, so we split it into a list of parts and pattern match the list in a _subfunction_

test/support/github_case/success_api.ex
```diff
@@ -0,0 +1,55 @@
...
+  end
+  defp mock_api_response(_method, ["installation", "repositories"], _, _, _) do
+    load_endpoint_fixture("installation_repositories")
+  end
+  defp mock_api_response(:post, ["installations", _id, "access_tokens"], _, _, _) do
+    %{
+      "token" => "v1.1f699f1069f60xxx",
+      "expires_at" => Timex.now() |> Timex.shift(hours: 1) |> DateTime.to_iso8601
+    }
+  end
```
begedin: Further requests can be added here. We should not have a catch-all request, since an umatched clause is the quickest way to detect that our `SuccessAPI` isn't _complete_

begedin:
# TODO LIST

- [x] Fix existing tests by switching them away from `tag bypass:` to `with_mock_api`
- [x] Add tests for `CodeCorps.GitHub.API` and `CodeCorps.GitHub.Headers`. The first is the complex one and should be the only one that still uses bypass (or we could even consider ExVCR for it. The second is far less complex in comparison.
- [x] Add any tests that may still be required for `CodeCorps.GitHub`

joshsmith (approved):
Looks great! Merge at will!

PR #879, opened 2017-08-30
Switches category controller away from ja_resource

crodriguez1a:
# What's in this PR?

- [x] Updates `Category` policy
      - can
      - update
      - create

- [x] Updates `Category` controller
       - index
       - show
       - update
       - create
 
Progress on: #864


crodriguez1a:
@snewcomer Not sure if the `structure.sql` should be committed?

snewcomer:
@crodriguez1a na I don't think so.

PR #880, opened 2017-08-30
Switches category controller away from ja_resource

crodriguez1a:
# What's in this PR?

- [x] Updates `Category` policy methods
      - can (`policy.ex`)
      - update (`category/policy.ex`)
      - create (`category/policy.ex`)

- [x] Updates `Category` controller methods
       - index
       - show
       - update
       - create
 
Progress on: #864


lib/code_corps/policy/category.ex
```diff
@@ -1,9 +1,7 @@
...
+  def create?(%{admin: true}), do: true
+  def create?(%{admin: false}), do: false
 
-  def create?(%User{admin: true}), do: true
-  def create?(%User{admin: false}), do: false
-
-  def update?(%User{admin: true}), do: true
-  def update?(%User{admin: false}), do: false
+  def update?(%{admin: true}), do: true
+  def update?(%{admin: false}), do: false
```
begedin: I understand your goal here is reduction of code, but we tend to prefer using stronger type constraints unless we have to loosen them for some reason. 

In fact, what we'd prefer here is to rewrite these as 

```Elixir
@spec create?(User.t) :: boolean
def create?(%User{admin: true}), do: true
def create?(%User{}), do: false

@spec update?(User.t) :: boolean
def update?(%User{admin: true}), do: true
def update?(%User{}), do: false
```

lib/code_corps/policy/policy.ex
```diff
@@ -24,6 +24,10 @@ defmodule CodeCorps.Policy do
   end
 
   @spec can?(User.t, atom, struct, map) :: boolean
+  defp can?(%User{} = user, :create, %Category{}, %{}), do: Policy.Category.create?(user)
+  defp can?(%User{} = user, :update, %Category{}, %{}), do: Policy.Category.update?(user)
+
+  @spec can?(User.t, atom, struct, map) :: boolean
```
begedin: These all have the same "prototype" (same amount of arguments, same types of arguments in the same order), so there is no need for an extra `@spec`, nor for keeping a blank line between them. Elixir, and Credo prefer no blank lines between the same function's (in the context of arity) different clauses.

This is something that wasn't clear since there was only a single example in the code previously, but should be clearer once we merge your PR in.

lib/code_corps_web/controllers/category_controller.ex
```diff
@@ -1,17 +1,43 @@
...
-  plug :load_and_authorize_resource, model: Category, only: [:create, :update]
-  plug JaResource
+  action_fallback CodeCorpsWeb.FallbackController
+  plug CodeCorpsWeb.Plug.DataToAttributes
 
-  @spec model :: module
-  def model, do: CodeCorps.Category
+  @spec index(Conn.t, map) :: Conn.t
+  def index(%Conn{} = conn, %{} = params) do
+    with category <- Category |> Query.id_filter(params) |> Repo.all do
```
begedin: `categories` instead of `category` here. We don't want to imply that index loads a single record. I'm guessing a copy-paste lapse :)

lib/code_corps_web/controllers/category_controller.ex
```diff
@@ -1,17 +1,43 @@
...
+  def update(%Conn{} = conn, %{"id" => id} = params) do
+    with %Category{} = category <- Category |> Repo.get(id),
+      %User{} = current_user <- conn |> Guardian.Plug.current_resource,
+      {:ok, :authorized} <- current_user |> Policy.authorize(:update, category),
+      {:ok, %Category{} = category} <- category |> Category.changeset(params) |> Repo.update
+    do
+      conn |> render("show.json-api", data: category)
+    end
   end
 end
```
begedin, line 43: Textbook quality controller conversion here 👍

begedin (changes requested):
Just some minor code style changes I'd like before we get this merged. Otherwise, done exactly as it should have been.

Awesome job and thank you!

crodriguez1a:
@begedin Having a little trouble when rebasing. Is there a wiki somewhere with your git conventions?

crodriguez1a:
@begedin Actually, I think I may have rebased from master(instead of develop). I may need to close this PR and open another.

crodriguez1a:
@begedin Closing in favor of #881 (since I accidentally rebased from master)

PR #882, opened 2017-08-31
Remove ja_resource and canary reliance from UserController

snewcomer:
# What's in this PR?
[WIP]
Remove ja_resource from User Controller

## References
Fixes #882

Progress on: #864

### DIFF
--- a/lib/code_corps/policy/policy.ex
+++ b/lib/code_corps/policy/policy.ex
@@ -42,8 +42,9 @@ defmodule CodeCorps.Policy do
     # will never do anything
     #
     # The only solution is to have a catch_all match for the resource being nil, which returns true
+    
+    # NOTE: other tests are using the User policy for the time being.
     def can?(%User{}, _action, nil), do: true
-
     def can?(%User{} = current_user, :update, %User{} = user), do: Policy.User.update?(user, current_user)
 
     def can?(%User{} = user, :create, %Changeset{data: %DonationGoal{}} = changeset), do: Policy.DonationGoal.create?(user, changeset)
--- a/lib/code_corps/policy/policy.ex
+++ b/lib/code_corps/policy/policy.ex
@@ -31,7 +31,6 @@ defmodule CodeCorps.Policy do
   defp can?(%User{} = user, :create, %Organization{}, %{}), do: Policy.Organization.create?(user)
   defp can?(%User{} = user, :update, %Organization{} = organization, %{}), do: Policy.Organization.update?(user, organization)  
   defp can?(%User{} = current_user, :update, %User{} = user, %{}), do: Policy.User.update?(user, current_user)
-  defp can?(%User{}, _action, %User{} = _user, %{}), do: true
 
   defimpl Canada.Can, for: User do
     # NOTE: Canary sets an :unauthorized and a :not_found handler on a config level
--- a/lib/code_corps/policy/policy.ex
+++ b/lib/code_corps/policy/policy.ex
@@ -44,7 +44,6 @@ defmodule CodeCorps.Policy do
     
     # NOTE: other tests are using the User policy for the time being.
     def can?(%User{}, _action, nil), do: true
-    def can?(%User{} = current_user, :update, %User{} = user), do: Policy.User.update?(user, current_user)
 
     def can?(%User{} = user, :create, %Changeset{data: %DonationGoal{}} = changeset), do: Policy.DonationGoal.create?(user, changeset)
     def can?(%User{} = user, :update, %DonationGoal{} = comment), do: Policy.DonationGoal.update?(user, comment)
--- a/lib/code_corps/policy/policy.ex
+++ b/lib/code_corps/policy/policy.ex
@@ -30,10 +30,8 @@ defmodule CodeCorps.Policy do
   defp can?(%User{} = user, :update, %Comment{} = comment, %{}), do: Policy.Comment.update?(user, comment)
   defp can?(%User{} = user, :create, %Organization{}, %{}), do: Policy.Organization.create?(user)
   defp can?(%User{} = user, :update, %Organization{} = organization, %{}), do: Policy.Organization.update?(user, organization)  
-
-  @spec can?(User.t, atom, struct) :: boolean
-  defp can?(%User{}, _action, nil), do: true
-  defp can?(%User{} = current_user, :update, %User{} = user), do: Policy.User.update?(user, current_user)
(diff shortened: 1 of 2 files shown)


lib/code_corps/policy/policy.ex
```diff
@@ -35,8 +38,9 @@ defmodule CodeCorps.Policy do
     # will never do anything
     #
     # The only solution is to have a catch_all match for the resource being nil, which returns true
+    
+    # NOTE: other tests are using the User policy for the time being.
     def can?(%User{}, _action, nil), do: true
-
     def can?(%User{} = current_user, :update, %User{} = user), do: Policy.User.update?(user, current_user)
```
snewcomer: NOTE: had to leave these in here b/c other tests not in the `user_controller_test` were depending on this

lib/code_corps_web/controllers/user_controller.ex
```diff
@@ -68,9 +71,9 @@ defmodule CodeCorpsWeb.UserController do
     conn |> json(hash)
   end
 
-  defp login(conn, _opts) do
-    Plug.Conn.register_before_send(conn, &do_login(&1))
-  end
+  # defp login(conn, _opts) do
```
snewcomer: @begedin any idea why commenting these lines (and the plug :login) above out allows for a passing test suite?  Otherwise I get this [line](https://github.com/code-corps/code-corps-api/blob/develop/test/lib/code_corps/github/installation_test.exs#L46) and this [line](https://github.com/code-corps/code-corps-api/blob/develop/test/lib/code_corps/github/installation_test.exs#L75) failing

begedin: I'm not sure, but I would expect the plug to cause things to fail either way. It was added mainly to get around some ja_resource quirks.

What I would do is remove `plug :login when action in [:create]`, as well as `defp login` and `defp do_login` completely.

Then, in `create/2`, I would add a `|> Guardian.Plug.sign_in(user)` between `conn` and `|> put_status(:created)`, or really, anywhere in that chain.

That would be the most explicit and easiest to read approach for dealing with it anyway.

lib/code_corps/policy/policy.ex
begedin: I understand why line 44 needs to stay, but are you sure about 45? Which tests fail if you remove it? Assuming you add the "fixes" i described below.

begedin (changes requested):
This generally looks good. Ping me for another review when you try the advice I gave.

lib/code_corps_web/controllers/user_controller.ex
snewcomer: @begedin is this b/c the user is signed in automatically after creating a user?

lib/code_corps_web/controllers/user_controller.ex
```diff
@@ -1,40 +1,42 @@
...
+    end
   end
 
-  def handle_create(_conn, attributes) do
-    %User{} |> User.registration_changeset(attributes)
+  @spec create(Conn.t, map) :: Conn.t
+  def create(%Conn{} = conn, %{} = params) do
+    with {:ok, %User{} = user} <- %User{} |> User.registration_changeset(params) |> Repo.insert
+    do
+      conn |> Plug.Conn.fetch_session |> Guardian.Plug.sign_in(user) |> put_status(:created) |> render("show.json-api", data: user)
```
snewcomer: @begedin so this worked.

snewcomer: Although a bit lost on why this is needed as compared to before these changes.

begedin: The `login` function was a plug. It automatically ran before performing the `create` action and told the conn that, after performing `create`, it needs to run the `do_login` function.

The `do_login` function then signed the `user` into the `conn` if the `create` action was successful.

Basically, from a user's perspective, "i want to be signed in when I register".

We had to do it via a plug, due to some quirks with `canary`.

Now that `canary` is gone, we can do the more direct route, so we just do it as part of the action itself.

begedin: If you mean the `fetch_session` part, that should not be needed. What happens if you remove it?

snewcomer: @begedin screen shot 2017-09-05 at 7 46 21 am

begedin: @snewcomer Strange, but it makes sense.

I think you actually found a bug in the old login function. We were just calling `Conn.assign` instead of actually formally signing the user in. 

Additionally, it looks like our client is not expecting to sign the user in automatically upon registration and instead does it manually.

Considering these two, I think the simplest thing we can do is outright remove both `fetch_session` and `sign_in`.

Simply put, the code should read `conn |> put_status(:created) |> render("show.json-api", data: user)`and we can leave it that that. 

Sorry to make go through all of this, but hey, we discovered a bug and a piece of code that does nothing along the way!

begedin (changes requested):

lib/code_corps/policy/policy.ex
```diff
@@ -30,6 +30,8 @@ defmodule CodeCorps.Policy do
   defp can?(%User{} = user, :update, %Comment{} = comment, %{}), do: Policy.Comment.update?(user, comment)
   defp can?(%User{} = user, :create, %Organization{}, %{}), do: Policy.Organization.create?(user)
   defp can?(%User{} = user, :update, %Organization{} = organization, %{}), do: Policy.Organization.update?(user, organization)  
+  defp can?(%User{} = current_user, :update, %User{} = user, %{}), do: Policy.User.update?(user, current_user)
+  defp can?(%User{}, _action, %User{} = _user, %{}), do: true
```
begedin: Remind me again, why is this one needed? It seems unsafe, as it would technically allow any user to perform  any action on any other user. 

If you meant to copy-paste the one from line 46 below, then the third argument would be nil, but that means you'd also have to update the spec to allow nil, not just `User.t` as the third argument.

However, I don't think the catch all from below is strictly necessary when not using canary, so could you double check?

snewcomer: Yep my bad.  Good call!

begedin (approved):

begedin:
Hey @snewcomer. I merged! Excellent work.

joshsmith:
🙌  I love seeing things like this happening when I'm not even looking. Great work @snewcomer!

Issue #883, opened 2017-09-01
Remove ja_resource from Organisation Controller


vishaldeepak:
# Problem
Progress on #864. 

Use steps defined in https://github.com/code-corps/code-corps-api/issues/864#issuecomment-325633865, to remove ja_resource from `organisation_controller.ex`

begedin:
Hey @vishaldeepak 

I've created individual issues for all controllers. Since they are uniform in format, I'll close this one in favor of #892

PR #886, opened 2017-09-01
Remove ja resource from organization controller

vishaldeepak:
# What's in this PR?
Remove `ja_resource` from organization controller. Code as per instructions

Targets issue #883

### DIFF
--- a/lib/code_corps/policy/policy.ex
+++ b/lib/code_corps/policy/policy.ex
@@ -28,6 +28,8 @@ defmodule CodeCorps.Policy do
   defp can?(%User{} = user, :update, %Category{}, %{}), do: Policy.Category.update?(user)
   defp can?(%User{} = user, :create, %Comment{}, %{} = params), do: Policy.Comment.create?(user, params)
   defp can?(%User{} = user, :update, %Comment{} = comment, %{}), do: Policy.Comment.update?(user, comment)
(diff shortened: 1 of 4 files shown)


lib/code_corps/policy/policy.ex
```diff
@@ -27,6 +27,9 @@ defmodule CodeCorps.Policy do
   defp can?(%User{} = user, :create, %Comment{}, %{} = params), do: Policy.Comment.create?(user, params)
   defp can?(%User{} = user, :update, %Comment{} = comment, %{}), do: Policy.Comment.update?(user, comment)
 
+  defp can?(%User{} = user, :create, %Organization{}, _), do: Policy.Organization.create?(user)
```
begedin: We prefer keeping functions with identical arity grouped up. Credo gives out a warning when that's not the case to, so I'd remove this blank line. Also, I know the last argument doesn't matter in this case, but the spec at the top says it's a map, so `%{}` instead of `_` is still prefered.

begedin (changes requested):
Branch needs updating so I can merge and there's a minor style detail. Other than that, great job. 👍

vishaldeepak:
Thanks, made required changes.

begedin (approved):

begedin:
@vishaldeepak Glad to merge this. Thank you for your work!

joshsmith:
🙌

PR #919, opened 2017-09-08
Remove reliance on ja_resource and canary from UserTaskController

### LINKED ISSUE #917: Switch UserTaskController away from ja_resource and canary
Progress on #864 

# Problem

# Procedure

- pick a controller to switch
- for each action

```Elixir
  end
end

  end
end

  end
end

  end
end
```

### `index` action

- render data

### `show` action

- load resource
- render resource

### `create` action

- perform insert action
- render resulting resource

- load resource
- perform update/delete

### Examples in code

## How to rewrite policies

```Elixir
```

becomes

```Elixir
```

### Examples in code

snewcomer:
## References
Fixes #917 

Progress on: #864

### DIFF
--- a/lib/code_corps/policy/helpers.ex
+++ b/lib/code_corps/policy/helpers.ex
@@ -86,9 +86,10 @@ defmodule CodeCorps.Policy.Helpers do
   @doc """
   Retrieves task from associated record
   """
-  @spec get_task(Changeset.t | TaskSkill.t | UserTask.t) :: Task.t
+  @spec get_task(Changeset.t | TaskSkill.t | UserTask.t | map) :: Task.t
   def get_task(%TaskSkill{task_id: task_id}), do: Repo.get(Task, task_id)
   def get_task(%UserTask{task_id: task_id}), do: Repo.get(Task, task_id)
+  def get_task(%{"task_id" => task_id}), do: Repo.get(Task, task_id)
   def get_task(%Changeset{changes: %{task_id: task_id}}), do: Repo.get(Task, task_id)
 
   @doc """
--- a/lib/code_corps/policy/policy.ex
+++ b/lib/code_corps/policy/policy.ex
@@ -31,6 +31,9 @@ defmodule CodeCorps.Policy do
   defp can?(%User{} = user, :create, %Organization{}, %{}), do: Policy.Organization.create?(user)
   defp can?(%User{} = user, :update, %Organization{} = organization, %{}), do: Policy.Organization.update?(user, organization)  
   defp can?(%User{} = current_user, :update, %User{} = user, %{}), do: Policy.User.update?(user, current_user)
+  defp can?(%User{} = user, :create, %UserTask{}, %{} = params), do: Policy.UserTask.create?(user, params)
+  defp can?(%User{} = user, :update, %UserTask{} = user_task, %{}), do: Policy.UserTask.update?(user, user_task)
+  defp can?(%User{} = user, :delete, %UserTask{} = user_task, %{}), do: Policy.UserTask.delete?(user, user_task)
 
   defimpl Canada.Can, for: User do
     # NOTE: Canary sets an :unauthorized and a :not_found handler on a config level
@@ -113,8 +116,5 @@ defmodule CodeCorps.Policy do
     def can?(%User{} = user, :create, %Changeset{data: %UserSkill{}} = changeset), do: Policy.UserSkill.create?(user, changeset)
     def can?(%User{} = user, :delete, %UserSkill{} = user_skill), do: Policy.UserSkill.delete?(user, user_skill)
 
-    def can?(%User{} = user, :create, %Changeset{data: %UserTask{}} = changeset), do: Policy.UserTask.create?(user, changeset)
-    def can?(%User{} = user, :update, %UserTask{} = user_task), do: Policy.UserTask.update?(user, user_task)
-    def can?(%User{} = user, :delete, %UserTask{} = user_task), do: Policy.UserTask.delete?(user, user_task)
   end
 end
--- a/lib/code_corps/policy/user_task.ex
+++ b/lib/code_corps/policy/user_task.ex
@@ -15,10 +15,10 @@ defmodule CodeCorps.Policy.UserTask do
   alias Ecto.Changeset
 
   @spec create?(User.t, Changeset.t) :: boolean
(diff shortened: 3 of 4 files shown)


lib/code_corps/policy/user_task.ex
```diff
@@ -15,10 +15,10 @@ defmodule CodeCorps.Policy.UserTask do
   alias Ecto.Changeset
 
   @spec create?(User.t, Changeset.t) :: boolean
-  def create?(%User{} = user, %Changeset{} = changeset) do
+  def create?(%User{} = user, params) do
     cond do
-      changeset |> get_task |> task_authored_by?(user) -> true
-      changeset |> get_task |> get_project |> contributed_by?(user) -> true
+      params |> get_task |> task_authored_by?(user) -> true
+      params |> get_task |> get_project |> contributed_by?(user) -> true
```
joshsmith, line 21: I know this isn't quite relevant to your PR at hand, but I'm not sure now that I'm seeing it that I like how `contributed_by?/1` reads. Maybe we should consider an issue to rewrite to something like `has_contributor?/1`?

lib/code_corps/policy/policy.ex
```diff
@@ -31,6 +31,9 @@ defmodule CodeCorps.Policy do
   defp can?(%User{} = user, :create, %Organization{}, %{}), do: Policy.Organization.create?(user)
   defp can?(%User{} = user, :update, %Organization{} = organization, %{}), do: Policy.Organization.update?(user, organization)  
   defp can?(%User{} = current_user, :update, %User{} = user, %{}), do: Policy.User.update?(user, current_user)
```
joshsmith, line 33: Also not in scope of this PR but it's odd to me seeing `current_user` as the first argument only in this case. That seems backwards, no?

lib/code_corps/policy/helpers.ex
```diff
@@ -86,9 +86,10 @@ defmodule CodeCorps.Policy.Helpers do
   @doc """
   Retrieves task from associated record
   """
-  @spec get_task(Changeset.t | TaskSkill.t | UserTask.t) :: Task.t
+  @spec get_task(Changeset.t | TaskSkill.t | UserTask.t | map) :: Task.t
   def get_task(%TaskSkill{task_id: task_id}), do: Repo.get(Task, task_id)
   def get_task(%UserTask{task_id: task_id}), do: Repo.get(Task, task_id)
+  def get_task(%{"task_id" => task_id}), do: Repo.get(Task, task_id)
```
joshsmith, line 92: Why do we need a map here when we didn't before?

joshsmith (commented):
Looks good, one relevant question and a couple questions outside of scope.

lib/code_corps/policy/policy.ex
snewcomer: oh...this was b/c it was for the `user` update action...so to differentiate the variable names.  Would you like it renamed?  Maybe `user` as first arg, then `user_struct` as second arg?

lib/code_corps/policy/helpers.ex
snewcomer: Yeah good call out.  So this was a result of the `create` action sending through a map instead of the struct....after removing the use of ja_resource, the policy accepts a `map` for the `create` action

lib/code_corps/policy/policy.ex
joshsmith: Yeah, what I had meant is that the first param in the policy is `user` and the second is `current_user`. That actually seems wrong. And, in fact, in the other policies the `user` _should be_ `current_user`! At least, this is how I currently understand it. Is that the right reading of the situation?

lib/code_corps/policy/helpers.ex
joshsmith: Ah, okay. That's a little unfortunate in the sense that we lose the pattern matching against `%Task{}`, for example, unless I'm misunderstanding the change.

joshsmith: I'm not sure there's much we can or _should_ do to change that, but just something I noticed.

lib/code_corps/policy/policy.ex
snewcomer: Yeah that might be a good idea.  Change user to `current_user` as the pattern matched first arg.

lib/code_corps/policy/helpers.ex
begedin: >That's a little unfortunate in the sense that we lose the pattern matching against `%Task{}`, for example, unless I'm misunderstanding the change.

We don't, not really.  We never pattern matched against `%Task{}` when authorizing an action coming from a `UserTaskController`. We pattern-matched on a `%UserTask{}` or a `%Changeset{}`, containing a `:task_id` change. Instead, we now pattern-match on a `%UserTask{}` or a `param` map  containing a `"task_id"` key.

We were, and still are, pattern-matching on a  `%Task{}` on a change coming from the `TaskController.`

I'm actually thinking I initially went a bit too far with trying to dry up the code here. We have this `get_task` method which is called from everywhere, implying there's a shared approach in authorizing. 

Really,  when authorizing actions on a `UserTask`, it's pretty simple.

- for `:update` and `:delete`, get the `user_task` record and associated `task` record, decide if the `current_user` can do those things
- for `:create`, look at the `params` sent from the client and decide if the `current_user` can create such a record 

Maybe we should make it a bit less dry to improve clarity. Not something that's the concern of this PR, though.

I do like the move away from changesets, though. They're intended for validation and we were forcing them into authorization.

Previously, it was

- get resource/params
- init changeset with resource/params
- authorize action on resource/changeset
- init same changeset again to validate an perform action

Not it's 

- get resource/params
- authorize action on resource/params
- perform action

@snewcomer Again, this is a conversation about what we could potentially do. Your changes are in line with the approach we're using to move away in other places.

lib/code_corps/policy/policy.ex
begedin: @snewcomer, @JoshSmith  We can actually do that right now, considering how the policy is implemented.

```Elixir
defp can?(%User{} = current_user, :update, %User{} = user, %{}), do: Policy.User.update?(current_user, user)
```

If you look at the policy itself, it just checks if the two records have the same ids, so we can reorder them without consequences at the moment.

begedin (approved):
I'm not actually seeing any issues with this PR, so it looks good to merge. Our conversation is something we can move into separate issues.

PR #920, opened 2017-09-10
Update dependencies, add Timber logging

joshsmith:
# What's in this PR?

- Updates a bunch of dependencies
- Update Elixir, Erlang/OTP versions
- Fix tests to work with new Poison
- Install new Timber version
- Add user context for Timber

Poison needs to be overridden to work with Sentry.


config/dev.exs
```diff
@@ -67,7 +67,7 @@ config :code_corps,
 # If the dev environment has no CLOUDEX_API_KEY set, we want the app
 # to still run, with cloudex in test API mode
 if System.get_env("CLOUDEX_API_KEY") == nil do
-  IO.puts("NOTE: No Cloudex configuration found. Cloudex is runnning in test mode.")
+  IO.puts("NOTE: No Cloudex configuration found. Cloudex is running in test mode.")
```
begedin, line 70: 🥇

lib/code_corps_web/guardian_serializer.ex
```diff
@@ -10,6 +10,24 @@ defmodule CodeCorpsWeb.GuardianSerializer do
...
+
+    if user do
+      name = full_name(user)
+      %Timber.Contexts.UserContext{id: user.id, email: user.email, name: name}
+      |> Timber.add_context()
+    end
+
+    {:ok, user}
+  end
   def from_token(_), do: {:error, "Unknown resource type"}
```
begedin, line 24: This is sort of adding a side-effect to the serializer.

I would be more comfortable specifying a separate, more explicit plug which adds this context.

Something like the combination of `CodeCorpsWeb.Plug.Segment` and `CodeCorps.Analytics.SegmentPlugTracker`, except in a single module.

begedin (dismissed):
I like the addition, but I also have a couple of suggestions about making it more explicit.

lib/code_corps_web/guardian_serializer.ex
joshsmith: In the interest of getting this merged, could we open an issue to circle back and address it? This, for me, is lower priority than other things on the totem pole.

begedin: #932

begedin (dismissed):

begedin:
Created issue. @JoshSmith Feel free to merge when ready.

PR #924, opened 2017-09-11
Clean up Policy.User, change order of arguments

### LINKED ISSUE #922: Change order of arguments in CodeCorps.Policy.User
# Problem

All other policies take in the current user as the first argument, and the record the action is performed as the second, but the user policy switches this order around.

We can, however, reverse the order to the correct one without consequences, due to the way the policy works internally. It returns true if the ids of the two records match and false otherwise.

begedin:
Closes #922 

I also took the liberty of renaming the first argument from all policy calls `user -> current_user` to make it as clear as possible.


joshsmith (approved):
Looks great!

Issue #802, opened 2017-05-02
When editing comment on cc, if comment has a github_id, update associated github comment


begedin:
# Problem

A CodeCorps project might be connected to one or more github repositories 

In such a project, all Tasks should also be connected to a GitHub issue belonging to one of those repositories, so the comments need to be connected to their respective comments as well.

When such a comment is updated, the same update should propagate to the related Github comment as well.

If the author is not connected, the system should provide a user to update as. This is provided by #800.

We should also extract the comment update behavior into a module, since it's getting more complex now.

## Subtasks
- [ ] If the module does not yet exist, add `/lib/codecorps/task/comment.ex`, namespaced `CodeCorps.Task.Comment`
- [ ] Add `update/3` function to module, which accepts a comment struct, an attributes map and a current user
  - updates comment
  - the comment should have a `github_id`. It's task parent should be related to a `GithubRepo`. This information should be used to update the associated github issue
   - this should be done on behalf of the comment user, or, if the comment user is not connected with github, on behalf of the identity provided in #800
- [ ] Modify Comment controller update action to call this new module
- [ ] Write tests for new module update function to ensure the github sync part works. The controller action itself is already tested well enough.

joshsmith:
Closed by #925.

PR #925, opened 2017-09-11
Added comment syncing with github during create and update

### LINKED ISSUE #801: When posting comment on cc, if task is tied to github issue, also create github comment
# Problem

A CodeCorps project might be connected to a github repository. In such a project, all Tasks should also be connected to a GitHub issue.

When a comment is created for such a task, a comment should also be created on github for the related issue.

We should also extract the comment update behavior into a module, since it's getting more complex now.

## Subtasks
- [ ] If the module does not yet exist, add `/lib/codecorps/task/comment.ex`, namespaced `CodeCorps.Task.Comment`
- [ ] Add a `create` function to the module. Iit should accept the current user and an attributes map
  - creates comment
  - if the associated `Task` is related to a `GithubRepo`, use information from the task (`GithubRepo`, `github_issue_number`, to create a github issue comment on github
   - this should be done on behalf of the comment user, or, if the comment user is not connected with github, on behalf of the identity provided in #800
- [ ] Modify Comment controller update action to call this new module
- [ ] Write tests for new module update function to ensure the github sync part works. The controller action itself is already tested well enough.

begedin:
Closes #801, #802

This PR implements github syncing of comments during the `:update` and `:create` actions.

It uses #805 as a template, so the implementation is similar.

#805 should be merged and this rebased before it's reviewed, since this PR builds on top of that one. The base of this PR should be changed to develop once #805 is merged.


joshsmith:
Are these merge conflicts due to needing a rebase or what? I know I need to review #805 first.

begedin:
@JoshSmith I'm not seeing any merge conflicts here. Did you fix them? EIther way, no point in reviewing this until we agree #805 is good to merge, since any changes to the process there will likely have to be applied to the process here as well.

joshsmith:
There are conflicts again.

begedin:
And there will be as long as #805 keeps getting modified. My vote is to leave this alone until that one is done. I've spent close to a day rebasing and resolving here, when it can't be merged anyway until we deal with #805

joshsmith (changes requested):
When the comment is created from Code Corps but done by the `code-corps-local[bot]` account, then when the webhook comes back the comment gets _overwritten_ with that (bot) user, overwriting entirely the original user on Code Corps who created it.

There's some work to do here to resolve this.

joshsmith:
Also, while I don't have time right this moment to write a failing test, posting a comment gets a `403` error because `plug CodeCorpsWeb.Plug.IdsToIntegers` is not being used in the `CommentController` (and wasn't clear to me that it needed to be used). This suggests something is wrong with our testing, or even just the way we're processing IDs generally.

joshsmith:
Unrelated – but it's only started happening here – when I run `mix phx.server` I now get tests running automatically every time the code changes, as though I were running `mix test.watch` or something. I see that `mix_test_watch` is set to run only in `:dev`, which may be part of the problem. But it makes it impossible to do debugging because the tests are just clouding the log output.

begedin:
@JoshSmith Regarding id's to integers, I should have written about this, but simply put, I forgot, so it's a mistake on my part. Sorry.

I encountered this about two weeks back. The phoenix API does not convert numbers in received request bodies into integers. However, elixir does not auto-cast strings to integers in any way, so comparison in those cases is troublesome.

The problem with that is that ember's json api adapter, casts all ids to strings, because, while our ids are all integers, some apis might have uuids, slugs, etc, for ids, so that's the safest approach from ember's POV.

Even for us `Repo.get(User, 1)` and `Repo.get(User, "1")` is interchangeable. It's just this specific case of comparing a record id against a param value, where the param value will always be string by default is when it fails. 

That's why I concluded that the better place to handle the conversion is in the API, even though it may make sense to do it in our ember app's serializers to.

As you can see, the issue has been identified and resolved in the task pr, but I didn't get a chance to propagate it to this comment PR yet.

All that being said, even though there are arguments for implementing the fix for the issue this way, through the plug, I'm still not completely sure this is the best approach, so I'm open for discussion and suggestions.

joshsmith:
@begedin I decided to merge this now so we can circle back to the user issues as a separate PR.

PR #926, opened 2017-09-13
[USERSKILL]: refactor away ja_resource

### LINKED ISSUE #916: Switch UserSkillController away from ja_resource and canary
Progress on #864 

# Problem

# Procedure

- pick a controller to switch
- for each action

```Elixir
  end
end

  end
end

  end
end

  end
end
```

### `index` action

- render data

### `show` action

- load resource
- render resource

### `create` action

- perform insert action
- render resulting resource

- load resource
- perform update/delete

### Examples in code

## How to rewrite policies

```Elixir
```

becomes

```Elixir
```

### Examples in code

snewcomer:
Fixes #916

### DIFF
--- a/lib/code_corps/policy/user_skill.ex
+++ b/lib/code_corps/policy/user_skill.ex
@@ -1,11 +1,9 @@
 defmodule CodeCorps.Policy.UserSkill do
-  alias CodeCorps.UserSkill
-  alias CodeCorps.User
-  alias Ecto.Changeset
+  alias CodeCorps.{UserSkill, User}
 
-  def create?(%User{admin: true}, %Changeset{}), do: true
-  def create?(%User{id: id}, %Changeset{changes: %{user_id: user_id}}), do: id == user_id
-  def create?(%User{}, %Changeset{}), do: false
+  def create?(%User{admin: true}, %{}), do: true
+  def create?(%User{id: id}, %{user_id: user_id}), do: id == user_id
+  def create?(%User{}, %{}), do: false
 
   def delete?(%User{admin: true}, %UserSkill{}), do: true
   def delete?(%User{id: id}, %UserSkill{user_id: user_id}), do: id == user_id
--- a/lib/code_corps/policy/user_skill.ex
+++ b/lib/code_corps/policy/user_skill.ex
@@ -2,7 +2,7 @@ defmodule CodeCorps.Policy.UserSkill do
   alias CodeCorps.{UserSkill, User}
 
   def create?(%User{admin: true}, %{}), do: true
-  def create?(%User{id: id}, %{user_id: user_id}), do: id == user_id
+  def create?(%User{id: id}, %{"user_id" => user_id}), do: id == user_id
   def create?(%User{}, %{}), do: false
 
   def delete?(%User{admin: true}, %UserSkill{}), do: true
--- a/lib/code_corps_web/controllers/user_skill_controller.ex
+++ b/lib/code_corps_web/controllers/user_skill_controller.ex
@@ -1,26 +1,43 @@
 defmodule CodeCorpsWeb.UserSkillController do
   use CodeCorpsWeb, :controller
-  use JaResource
 
-  import CodeCorps.Helpers.Query, only: [id_filter: 2]
+  alias CodeCorps.{UserSkill, User, Helpers.Query}
 
-  alias CodeCorps.UserSkill
+  action_fallback CodeCorpsWeb.FallbackController
+  plug CodeCorpsWeb.Plug.DataToAttributes
 
-  plug :load_resource, model: UserSkill, only: [:show], preload: [:user, :skill]
-  plug :load_and_authorize_changeset, model: UserSkill, only: [:create]
-  plug :load_and_authorize_resource, model: UserSkill, only: [:delete]
-  plug JaResource
-
-  @spec model :: module
-  def model, do: CodeCorps.UserSkill
-
-  @spec filter(Plug.Conn.t, Ecto.Query.t, String.t, String.t) :: Ecto.Query.t
-  def filter(_conn, query, "id", id_list) do
-    query |> id_filter(id_list)
+  @spec index(Conn.t, map) :: Conn.t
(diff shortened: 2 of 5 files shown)


lib/code_corps_web/controllers/user_skill_controller.ex
```diff
@@ -1,26 +1,43 @@
...
 
-  @spec handle_create(Plug.Conn.t, map) :: Ecto.Changeset.t
-  def handle_create(_conn, attributes) do
-    %UserSkill{} |> UserSkill.create_changeset(attributes)
+  @spec delete(Conn.t, map) :: Conn.t
+  def delete(%Conn{} = conn, %{"id" => id} = _params) do
+    with %UserSkill{} = user_skill <- UserSkill |> Repo.get(id),
+      %User{} = current_user <- conn |> Guardian.Plug.current_resource,
+      {:ok, :authorized} <- current_user |> Policy.authorize(:delete, user_skill),
+      {:ok, %UserSkill{} = _user_skill} <- user_skill |> Repo.delete
```
snewcomer, line 39: @begedin getting a test failure [here](https://github.com/code-corps/code-corps-api/blob/develop/test/lib/code_corps_web/controllers/user_skill_controller_test.exs#L85)

Not really familiar with this `:track` code.  Lmk what you think here.

begedin: You've discovered a bug we've been creating on some controllers with this switch

We have a plug, `SegmentPlugTracker`, which registers an action right before the conn sends a response.

This action figures out what happened (in this case, that a `UserSkill` was deleted) and calls tracking code for that action.

However, the "figuring out" also relied on ja_resource and canary, so in this case, it got broken and the controller test found it. Unfortunately it seems not all our controller tests have this `assert_received`, so we may have turned off tracking in some cases. I'll make sure to check.

We probably want to make that explicit as well, to avoid further issues, so I'll create an issue to do this. 

For know, to keep tracking working here, you'll need to do a `conn |> Conn.assign(:user_skill, user_skill) |> conn |> send_resp(:no_content, "")`

begedin (changes requested):
Good work. The failed test is actually a good thing and you may have stopped us from causing bugs down the line.

joshsmith:
@snewcomer perhaps we need some higher-level docs somewhere about the tracking?

begedin:
@JoshSmith Higher level docs are good, but I've also created #927 to avoid unclear cases like this one. I think we made far too many things implicit in the early days of this project.

snewcomer:
@begedin is this g2g?

begedin:
@snewcomer Squash the two commits together and then you can consider it merged. However, I'm hoping #805 gets merged first, since rebasing this one would be much easier, so don't worry, if the PR stays up for a bit. It's just waiting on that one.

joshsmith:
#805 is possibly not going to get merged quickly, pending what @begedin says in reply to review comments from today.

lib/code_corps/policy/user_skill.ex
```diff
@@ -1,11 +1,9 @@
...
-  alias CodeCorps.UserSkill
-  alias CodeCorps.User
-  alias Ecto.Changeset
+  alias CodeCorps.{UserSkill, User}
 
-  def create?(%User{admin: true}, %Changeset{}), do: true
-  def create?(%User{id: id}, %Changeset{changes: %{user_id: user_id}}), do: id == user_id
-  def create?(%User{}, %Changeset{}), do: false
+  def create?(%User{admin: true}, %{}), do: true
+  def create?(%User{id: id}, %{user_id: user_id}), do: id == user_id
```
begedin: @snewcomer The params we send here use strings as keys. Additionally, there's an issue with id values in the params map all being strings (this is what ember sends), so to fix the problem here you need to 

- replace this line with `def create?(%User{id: id}, %{"user_id" => user_id}), do: id == user_id`
- update the policy test to match
- add `plug CodeCorpsWeb.Plug.IdsToIntegers` into the controller, right after the `DataToAttributes` plug.

test/lib/code_corps/policy/user_skill_test.exs
```diff
@@ -16,9 +16,9 @@ defmodule CodeCorps.Policy.UserSkillTest do
 
     test "returns true if user is creating their own record" do
       user = insert(:user)
-      changeset = %UserSkill{} |> create_changeset(%{user_id: user.id})
+      user_skill = %UserSkill{user_id: user.id}
```
begedin: As I explained above, this needs to be a `params` map isntead, looking like

```Elixir
params = %{"user_id" => user.id}

assert create?(user, params)
```

begedin (changes requested):
We merged one of the two PRs, so rebasing the other became simple enough where we can get this merged to. 

However, on a second review, I found a few things that need to be fixed.

lib/code_corps/policy/user_skill.ex
snewcomer: @begedin thats a great catch.  Fixed.

test/lib/code_corps/policy/user_skill_test.exs
```diff
@@ -11,21 +11,19 @@ defmodule CodeCorps.Policy.UserSkillTest do
       user = build(:user, admin: true)
       changeset = %UserSkill{} |> create_changeset(%{})
 
-      assert create?(user, changeset) 
+      assert create?(user, %{"user_id" => user.id})
```
begedin: You're not using the changeset here anymore. You can likely also remove `import CodeCorps.UserSkill, only: [create_changeset: 2]` from the top of the file.

Sorry I didn't catch this on my previous pass.

begedin (changes requested):

test/lib/code_corps/policy/user_skill_test.exs
snewcomer: oops.  fixed.

begedin (approved):

begedin:
Thank you @snewcomer. Great work 👍

PR #930, opened 2017-09-13
Added missing tests for user controller github connect

### LINKED ISSUE #790: Add handling of error scenarios for github connect action
# Problem

The action defined in #788 calls the function defined in #789, but only deals with the successful response, which is an `{:ok, user}`.

Thanks to `ja_resource`, it will also successfully deal with an `{:error, changeset}` response, but we should still test for this scenario on the controller level.

Lastly, it does not know how to handle an `{:error, HTTPoison.ErrorResponse}` response, so we should handle that by rendering a 500, and write a test for that.

## Subtasks

- ~~Write user controller test for the `{:error, changeset}` case~~
- [x] Write handling for the `{:error, HTTPoison.ErrorResponse}` case
- [x] Write test for the `{:error, HTTPoison.ErrorResponse}` case

## References

Blocked by: #788, #789

begedin:
Closes #790 

Makes use of the same `FailureAPI` mock module used in #805 and #925, so depending on merge order, there may be an easily resolvable conflict.

I did not add a test for the changeset case because it doesn't seem to be possible for it to happen, since the only validations are those for presence of github keys, and these cannot be nil.

### DIFF
--- a/test/lib/code_corps_web/controllers/user_controller_test.exs
+++ b/test/lib/code_corps_web/controllers/user_controller_test.exs
@@ -3,6 +3,8 @@ defmodule CodeCorpsWeb.UserControllerTest do
 
   use CodeCorpsWeb.ApiCase, resource_name: :user
 
+  import CodeCorps.GitHub.TestHelpers
+
   alias CodeCorps.{User, Repo}
 
   @valid_attrs %{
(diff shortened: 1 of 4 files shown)


test/lib/code_corps_web/controllers/user_controller_test.exs
```diff
@@ -243,26 +245,33 @@ defmodule CodeCorpsWeb.UserControllerTest do
...
+      json = conn |> post(path, attrs) |> json_response(200)
 
-      json = conn |> authenticate(user) |> post(path) |> json_response(200)
+      assert json["data"]["id"] |> String.to_integer == current_user.id
+      assert json["data"]["attributes"]["github-id"]
+    end
 
-      assert json["data"]["id"] |> String.to_integer == user.id
+    test "requires authentication", %{conn: conn} do
+      path = user_path(conn, :github_oauth, %{"code" => "foo", "state" => "bar"})
```
joshsmith: Unless I'm misreading this entirely, I think what's confusing me about this test is that it implies that `valid_code` and `valid_state` somehow work whereas `foo` and `bar` would not, when in reality it simply tests that the user is authenticated with our system regardless of the params that make it to GitHub.

Can you clarify @begedin?

test/support/github/failure_api.ex
```diff
@@ -0,0 +1,37 @@
...
+  access token request, that one is set to succeed here as well.
+  """
+  import CodeCorps.GitHub.TestHelpers
+
+  alias CodeCorps.GitHub.SuccessAPI
+
+  def request(method, url, headers, body, options) do
+    case {method, url} |> for_access_token?() do
+      true -> SuccessAPI.request(method, url, headers, body, options)
+      false ->
```
joshsmith, line 21: I wonder if there's a way to metaprogram this and use an `@impl` or something?

test/lib/code_corps_web/controllers/user_controller_test.exs
begedin: It should work regardless, I'll update.

begedin: @JoshSmith To clarify, the controller action was implemented with the `code` and `state` in the signature. I've moved `attrs` into a module variable, so it doesn't imply that specific values are important. 

However, the action will only work with attributes containing `code` and `state` keys.

Not sure if I should loosen that requirement, or write a test for it. Either way, this PR is about closing #790. If we want to loosen the requirement, then that requires a rewrite of `GitHub.User` as well, so I would do it separately. Adding a test could go here, but this would be the first time we're writing a test for such a case.

joshsmith: This addresses totally what I was taking issue with, thanks!

joshsmith (approved):

Issue #885, opened 2017-09-01
Bug in Organization Policy Test


vishaldeepak:
# Problem
This is following test in  the policy `/policy/organization_test.ex`

```elixir
    test "returns true when user is the organization owner" do
      user = insert(:user, admin: true)
      organization = build(:organization, owner_id: user.id)
      assert update?(user, organization)
    end
```
I feel this test case should check if  the user is not an admin but owns the organisation he should be allowed to update.The issue here is that we need to make `admin` is false.

The `admin:true` case is already covered by test case _"returns true when user is an admin"_ for _update_.

This is the corresponding policy in `/policy/organization.ex` is as follows 
```elixir
  def update?(%User{admin: true}, %Organization{}), do: true
  def update?(%User{} = user, %Organization{} = organization), do: organization |> owned_by?(user)
```

begedin:
Good job finding this. Feel free to correct, either as part of your PR to switch away from ja_resource, or separately. If not, someone else can take it. 👍

vishaldeepak:
Il raise a PR separately.

PR #931, opened 2017-09-14
Fixes a bug in organization_test.exs where a user who is not an admin…

vishaldeepak:
Fixes bug #885

# What's in this PR?
Simply removed `admin:true` from the test when user is created.


begedin (approved):

begedin:
Thank you for this @vishaldeepak 👍

joshsmith:
🙌

PR #933, opened 2017-09-14
Moved Project Controller away from ja resource.

vishaldeepak:
# What's in this PR?
Everything is as given in the instructions. I had to add a new helper method `get_organization`in `/policy/helpers`  which takes string as key. 

## References
Fixes #
#896


begedin (approved):
Good work on this @vishaldeepak 👍 

Good work in general!

PR #943, opened 2017-09-15
Move Project User Controller away from ja_resource and canary

### LINKED ISSUE #900: Switch ProjectUserController away from ja_resource and canary
Progress on #864 

# Problem

# Procedure

- pick a controller to switch
- for each action

```Elixir
  end
end

  end
end

  end
end

  end
end
```

### `index` action

- render data

### `show` action

- load resource
- render resource

### `create` action

- perform insert action
- render resulting resource

- load resource
- perform update/delete

### Examples in code

## How to rewrite policies

```Elixir
```

becomes

```Elixir
```

### Examples in code

vishaldeepak:
Closes #900 

# What's in this PR?

This move is definitely more complex than the other controllers, require a few clarifictions. More emphasis needs to be had on the change in `policy/project_user.ex`.


lib/code_corps_web/controllers/project_user_controller.ex
```diff
@@ -1,35 +1,62 @@
...
+      conn |> render("index.json-api", data: project_users)
+    end
+  end
 
-  @spec filter(Plug.Conn.t, Ecto.Query.t, String.t, String.t) :: Ecto.Query.t
-  def filter(_conn, query, "id", id_list) do
-    query |> id_filter(id_list)
+  @spec show(Conn.t, map) :: Conn.t
+  def show(%Conn{} = conn, %{"id" => id}) do
+    with %ProjectUser{} = project_user <- ProjectUser |> Repo.get(id) |> Repo.preload(@preloads) do
```
vishaldeepak: Preloading here may or may not be required.

lib/code_corps_web/controllers/project_user_controller.ex
```diff
@@ -1,35 +1,62 @@
...
   end
 
-  @spec handle_create(Plug.Conn.t, map) :: Ecto.Changeset.t
-  def handle_create(_conn, attributes) do
-    %ProjectUser{} |> ProjectUser.create_changeset(attributes)
+  @spec create(Plug.Conn.t, map) :: Conn.t
+  def create(%Conn{} = conn, %{} = params) do
+    with %User{} = current_user <- conn |> Guardian.Plug.current_resource,
+         {:ok, :authorized} <- current_user |> Policy.authorize(:create, %ProjectUser{}, params),
+         {:ok, %ProjectUser{} = project_user} <- %ProjectUser{} |> ProjectUser.create_changeset(params) |> Repo.insert
```
vishaldeepak, line 30: The old code had `load_and_authorize_changeset` which included preload. I don't quite understand where the preloading should be happening here.
`  plug :load_and_authorize_changeset, model: ProjectUser, only: [:create, :update], preload: @preloads`

lib/code_corps_web/controllers/project_user_controller.ex
```diff
@@ -1,35 +1,62 @@
...
+      conn |> put_status(:created) |> render("show.json-api", data: project_user)
+    end
   end
 
-  @spec handle_update(Plug.Conn.t, ProjectUser.t, map) :: Ecto.Changeset.t
-  def handle_update(_conn, model, attributes) do
-    model |> ProjectUser.update_changeset(attributes)
+  @spec update(Conn.t, map) :: Conn.t
+  def update(%Conn{} = conn, %{"id" => id} = params) do
+    with %ProjectUser{} = project_user <- ProjectUser |> Repo.get(id) |> Repo.preload(@preloads),
```
vishaldeepak: Preload here may or may not be required let me know

lib/code_corps/policy/project_user.ex
```diff
@@ -7,17 +7,18 @@ defmodule CodeCorps.Policy.ProjectUser do
...
+  @spec create?(User.t, map) :: boolean
+  def create?(%User{id: user_id}, %{"user_id" => author_id})
+    when user_id == author_id and not is_nil(user_id), do: true
+  def create?(%User{}, %{}), do: false
 
-  @spec update?(User.t, Ecto.Changeset.t) :: boolean
-  def update?(%User{} = user, %Changeset{data: %ProjectUser{} = record} = changeset) do
-    user_membership = record |> get_project_membership(user)
+  @spec update?(User.t, ProjectUser.t, map) :: boolean
+  def update?(%User{} = user, %ProjectUser{} = existing_record, params) do
```
vishaldeepak, line 15: The change requires to take 3 params. This part of code caused a lot of change in the corresponding, `project_user_test.ex` file

vishaldeepak:
There is one test that is failing. Cant seem to figure out the cause. Would love some feedback on this. 
```
  1) test update updates and renders resource when data is valid (CodeCorpsWeb.ProjectUserControllerTest)
     test/lib/code_corps_web/controllers/project_user_controller_test.exs:81
     There were 0 emails delivered to this process.

     If you expected an email to be sent, try these ideas:

     1) Make sure you call deliver_now/1 or deliver_later/1 to deliver the email
     2) Make sure you are using the Bamboo.TestAdapter
     3) Use shared mode with Bamboo.Test. This will allow Bamboo.Test
     to work across processes: use Bamboo.Test, shared: :true
     4) If you are writing an acceptance test through a headless browser, use
     shared mode as described in option 3.

     code: assert_delivered_email email
     stacktrace:
       test/lib/code_corps_web/controllers/project_user_controller_test.exs:102: (test)
```

test/lib/code_corps_web/controllers/project_user_controller_test.exs
```diff
@@ -99,6 +99,8 @@ defmodule CodeCorpsWeb.ProjectUserControllerTest do
         |> CodeCorps.Repo.preload([:project, :user])
         |> CodeCorps.Emails.ProjectUserAcceptanceEmail.create()
 
+      email |> CodeCorps.Mailer.deliver_now()
+
```
vishaldeepak: Test is not passing without the explicit call to `deliver_now`. It used to pass earlier without this. Any ideas why?

lib/code_corps/policy/project_user.ex
begedin: I understand why and it was expected. Good call on moving on with the changes. A lot of contributors would stop here, since most people are uneasy about making bigger alterations.

lib/code_corps/policy/project_user.ex
```diff
@@ -5,19 +5,19 @@ defmodule CodeCorps.Policy.ProjectUser do
...
 
   alias CodeCorps.{ProjectUser, Repo, User}
-  alias Ecto.Changeset
 
-  @spec create?(User.t, Ecto.Changeset.t) :: boolean
-  def create?(%User{id: id}, %Changeset{changes: %{user_id: user_id}}), do:  id == user_id
-  def create?(%User{}, %Changeset{}), do: false
+  @spec create?(User.t, map) :: boolean
+  def create?(%User{id: user_id}, %{"user_id" => author_id})
+    when user_id == author_id and not is_nil(user_id), do: true
```
begedin, line 11: We have a bug that isn't being tested for right now. Parameters received from the client will all come as strings, including ids. That means that, as you pass in the params into the policy, `author_id` will be a string, while `user_id` will be an integer.

There are no tests right now which would reveal this, but PR #950 deals with modifying our JsonAPI test helpers to create failing tests.

To fix the issue, what you need to do is add `plug CodeCorpsWeb.Plug.IdsToIntegers` to the `ProjectUserController`.

Once we merge #950, if this still stays open, you should be getting test failures if you don't add the plug.

lib/code_corps_web/controllers/project_user_controller.ex
begedin: It doesn't look like the preloads are needed for any of the two (`:create, :update`) actions. Maybe they were needed at some point, but we may have changed it. If the tests pass without preloads, feel free to ditch them.

lib/code_corps_web/controllers/project_user_controller.ex
begedin: See comment on `create` action

test/lib/code_corps_web/controllers/project_user_controller_test.exs
begedin: The email should have been delivered from the controller, so you should definitely remove this line. Instead, if you look at the `schedule_email` plug, it calls `maybe_send_email`, which then uses the changeset to determine if the `project_user` was promoted from "pending", to "contributor" and sends an email in that case.

What we should do instead is, within the action method of the controller, figure out if there was a role change from "pending" to "contributor" and explicitly send the email from the `with` clause in that case.

My suggestion here would be to rename the second `project_user` into `updated_project_user` and then compare the role of the one you initially loaded from the database against the updated one and send an email if there was the change we need.

begedin (changes requested):
Good work so far, @vishaldeepak 

I added some comments which will hopefully help further.

joshsmith:
@vishaldeepak let me know if you want any help on the changes requested here. Sorry this wasn't clear initially.

@begedin wondering if we should make some sort of guide or docs for these tests since we're familiar with the general approach here.

begedin (approved):

begedin:
@vishaldeepak As usual, great work 👍

PR #945, opened 2017-09-16
779 Scheme is missing when images are placed into email dynamically f…

paulsullivanjr:
779 Scheme is missing when images are placed into email dynamically from Cloudinary

# What's in this PR?

Added a function to prepend cloudinary url's with https://

## References
Fixes #

Progress on: #

### DIFF
--- a/lib/code_corps/cloudex/cloudinary_url.ex
+++ b/lib/code_corps/cloudex/cloudinary_url.ex
@@ -7,5 +7,15 @@ defmodule CodeCorps.Cloudex.CloudinaryUrl do
   end
   def for(public_id, options, _version, _default_color, _type) do
     @cloudex.Url.for(public_id, options)
+    |> add_uri_scheme
   end
+
+  defp add_uri_scheme(generated_url) do
+    base_url =  String.split(generated_url, "//")
+    add_https(base_url)
+  end
+
+  defp add_https(base_url) when is_list(base_url) and length(base_url) > 0, do: "https://" <> List.last(base_url)
+  defp add_https(url), do: url
+
 end
--- a/test/lib/code_corps/cloudex/cloudinary_url_test.exs
+++ b/test/lib/code_corps/cloudex/cloudinary_url_test.exs
@@ -8,6 +8,12 @@ defmodule CodeCorps.Cloudex.CloudinaryUrlTest do
     assert expected_url == url
   end
 
+  test "call Cloudex.Url.for insert https://" do
+    expected_url = "https://placehold.it/100x100"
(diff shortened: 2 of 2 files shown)


test/lib/code_corps/cloudex/cloudinary_url_test.exs
```diff
@@ -8,6 +8,12 @@ defmodule CodeCorps.Cloudex.CloudinaryUrlTest do
     assert expected_url == url
   end
 
+  test "call Cloudex.Url.for insert https://" do
+    expected_url = "https://placehold.it/100x100"
+    url = CloudinaryUrl.for("res-x://placehold.it/100x100", %{height: 100, width: 100}, nil, nil, nil)
```
joshsmith: It appears Cloudinary actually sends back a URL like `//res.cloudinary.com`

joshsmith:
@begedin could you give this a second set of eyes beyond my own?

lib/code_corps/cloudex/cloudinary_url.ex
```diff
@@ -7,5 +7,11 @@ defmodule CodeCorps.Cloudex.CloudinaryUrl do
   end
   def for(public_id, options, _version, _default_color, _type) do
     @cloudex.Url.for(public_id, options)
+    |> add_html_scheme
+  end
+
+  def add_html_scheme(generated_url) do
+    base_url =  String.split(generated_url, "//")
+    if (length(base_url) > 0), do: "https://" <> List.last(base_url), else: generated_url
   end
```
begedin: Could be rewritten without an iff using something like

```Elixir
defp add_html_scheme("//" <> base_url), do: "https://" <> base_url
defp add_html_scheme(url), do: url
```

Note that I also made the functions private, since they're not being called externally.

begedin (changes requested):

lib/code_corps/cloudex/cloudinary_url.ex
```diff
@@ -7,5 +7,15 @@ defmodule CodeCorps.Cloudex.CloudinaryUrl do
...
   def for(public_id, options, _version, _default_color, _type) do
     @cloudex.Url.for(public_id, options)
+    |> add_html_scheme
   end
+
+  defp add_html_scheme(generated_url) do
```
joshsmith: So sorry for all the back and forth here, but I do think that `html` is not quite right here. I think it should be `add_uri_scheme`. Thoughts?

paulsullivanjr: its all good I meant to change that and forgot. I just pushed the update

joshsmith (approved):

begedin (approved):

begedin:
Good work @paulsullivanjr

PR #955, opened 2017-09-19
Added "github" context to users created during github sync

### LINKED ISSUE #948: When creating a user from GitHub, set their signup context
# Problem

When a user is created from GitHub webhooks, their sign up context is `default`.

We could instead do `github-api`? `github-webook`? Something else?

begedin:
Closes #948 

`Issues.UserLinker` and `IssueComment.UserLinker` ended up needing the exact same change and are quite similar to begin with. Instead of trying to force this into a `Github.Event.Common` module, I figured this is a good opportunity to start a `CodeCorps.Account` bounded context, with its first function, `Account.create_from_github/1`

- added `Account` bounded context
- added `Account.create_from_github/1`
- added `AccountTest`
- added `fixtures/github/endpoint/user.json`

The context is simply `"github"` since the whole idea is that the user being created originates from github.

### DIFF
--- a/lib/code_corps/accounts/accounts.ex
+++ b/lib/code_corps/accounts/accounts.ex
@@ -1,0 +1,26 @@
+defmodule CodeCorps.Accounts do
+  @moduledoc ~S"""
+  Main entry-point for managing accounts.
+
+  All actions to acounts should go through here.
+  """
+
+  alias CodeCorps.{
+    GitHub.Adapters,
+    User,
+    Repo
+  }
+  alias Ecto.Changeset
+
+  @doc ~S"""
+  Creates a user record using attributes from a GitHub payload.
+  """
+  @spec create_from_github(map) :: {:ok, User.t} | {:error, Changeset.t}
+  def create_from_github(%{} = attrs) do
+    %User{}
+    |> Changeset.change(attrs |> Adapters.User.from_github_user())
+    |> Changeset.put_change(:context, "github")
+    |> Changeset.unique_constraint(:email)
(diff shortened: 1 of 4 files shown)


joshsmith:
The Phoenix docs on contexts seem to indicate that a pluralized form – `Accounts` – might make more sense here. How do you feel about this?

lib/code_corps/account/account.ex
```diff
@@ -0,0 +1,26 @@
...
+
+  @doc ~S"""
+  Creates a user record using attributes from a GitHub payload.
+  """
+  @spec create_from_github(map) :: {:ok, User.t} | {:error, Changeset.t}
+  def create_from_github(%{} = attrs) do
+    %User{}
+    |> Changeset.change(attrs |> Adapters.User.from_github_user())
+    |> Changeset.put_change(:context, "github")
+    |> Changeset.unique_constraint(:email)
```
joshsmith: What happens if the email is `nil`? GitHub is not guaranteed to return one.

begedin: We already merged in a PR where a user can be inserted without an email, so we should be fine. Really, what's important for us here is the github id, so we can associate this "stub" with an actual user later, when they connect with github.

begedin:
> The Phoenix docs on contexts seem to indicate that a pluralized form – Accounts – might make more sense here. How do you feel about this?

I don't feel strongly about it one way or another. I can make the quick change

joshsmith (dismissed):

joshsmith (approved):

PR #956, opened 2017-09-19
When connecting github user, associate with tasks and comments

### LINKED ISSUE #941: When connecting user to github, associate tasks
# Problem

At the moment a code corps user is connected to github, there is a good chance there will already be task records on codecorps, associated to github issues associated to the user.

During the connection process, just like we do with installations, we should identify tasks which need to be associated to the newly connected user and associate them.

The tasks that need to be associated, will have already been associated with premade user records containing the newly connected user's github id.

That means, as the user connects, we need to

- find all users with the same `github_id`
- find their tasks
- reassociate those tasks with the newly connected user

Tagging as question because it likely needs more discussion

### LINKED ISSUE #942: When connecting user to github, associate comments
# Problem

At the moment a code corps user is connected to github, there is a good chance there will already be comment records on codecorps, associated to github comments associated to the user.

During the connection process, just like we do with installations, we should identify comments which need to be associated to the newly connected user and associate them.

The comments that need to be associated, will have already been associated with premade user records containing the newly connected user's github id.

That means, as the user connects, we need to

- find all users with the same `github_id`
- find their comments
- reassociate those comments with the newly connected user

Tagging as question because it likely needs more discussion

begedin:
Closes #941, closes #942

This PR adds 2 additional steps to the connect process. One associates existing tasks with the newly connected user, while the other does the same with comments.

There is a decision to be made about the user previously associated with these. Should the record be deleted, or should something else be done with it?

That's something that can be done as a separate issue, however.


lib/code_corps/github/user.ex
```diff
@@ -50,6 +53,8 @@ defmodule CodeCorps.GitHub.User do
       Multi.new
       |> Multi.update(:user, changeset)
       |> Multi.run(:installations, fn %{user: %User{} = user} -> user |> associate_installations() end)
+      |> Multi.run(:tasks, fn %{user: %User{} = user} -> user |> associate_tasks() end)
+      |> Multi.run(:comments, fn %{user: %User{} = user} -> user |> associate_comments() end)
```
joshsmith, line 57: I get why you chained more `Multi.run`s on here but I'm deeply concerned about the number of tasks/comments that we may be associating here. I had originally envisioned this process being fairly asynchronous and displaying to the user something to the effect of "your account is syncing with GitHub", even going so far as to be blocking.

This sync problem is going to be even larger when we sync large repos from GitHub to projects.

I know this adds some non-trivial complexity here. We could probably get away with this as a first pass, MVP and all. But this is something we should really circle back to, at the very least.

joshsmith: Reading that these are `Repo.update_all` the operation may not be too slow. We should still write down an issue to benchmark certainly.

Syncing projects does seem like a different beast. That we need an issue for.

joshsmith:
My review above can be considered a 👍  if my questions are a non-issue right now. This can be merged if so.

joshsmith:
Do you want to create an issue for project sync async?

begedin:
Created #959

Issue #878, opened 2017-08-30
Timestamps on issues/comments do not match GitHub's


joshsmith:
# Problem

When an issue is synced from GitHub, the `inserted_at` timestamp is used as the timestamp in the UI for when the task was created/written. This is not correct and can lead to some confusing differences between GitHub's timestamps and our own.

We should correct this by specifying the timestamp when the sync happens, using the data returned by GitHub's payload.

@begedin question for you is whether we should consider adding a different timestamp here to distinguish between `inserted_at` (which has a very particular meaning) and when the issue was _actually_ posted at.

begedin:
@JoshSmith I did not consider that, but yes, it would make sense to keep "inserted_at" as the timestamp for then the record was created in our own database.

As for tracking `posted_at` in relation to github, I'm thinking we either

1. Leave it for now - we don't have an actual use yet, other than UI
2. Name it `inserted_on_github_at` (To keep ecto conventions). If we're storing it just for the sake of it, then an `edited_on_github_at` makes sense to.

I'd actually vote on leaving it be right now. When we find we have a use for it, we can easily add it in.  Sure, we'd then have to use a script to update those fields across our tasks, but that's not difficult to do.

joshsmith:
I raised the issue exactly because it's confusing in the UI right now. I'd like to see a `post_created_at` and `post_edited_at` (and `comment_` versions, too) for these that takes the time from either us or GitHub depending on source. The UI would then use the `post_` and `comment_` for timestamp display instead of the default as it currently does.

begedin:
This would then involve

## For comments - #887

- add `comment_created_at`, `comment_updated_at` timestamp fields
- modify `CodeCorps.Comment.create_changeset/2` to update `comment_created_at` to the current timestamp
- add `CodeCorps.Comment.update_changeset/2`, which calls `changeset/2` and additionally sets the value of `comment_updated_at` to the current timestamp
- modify `CodeCorpsWeb.CommentController.update/2` so it calls `Comment.update_changeset/2` instead of `Comment.changeset/2`
- probably make `Comment.changeset/2` private, if no other piece of the code is using it
- modify `Comment.changeset/2` tests so they are actually `Comment.update_changeset/2` tests and add an assertion for the `comment_updated_at` field being set, **if the changeset was made private**
- if the changeset **was not made private**, add `Comment.update_changeset/2` tests in addition to the existing tests
[...]

PR #960, opened 2017-09-20
Modify Task to accommodate Github sync timestamps

### LINKED ISSUE #888: Modify Task timestamps to accommodate github sync
- Tasks can be multi-sourced. They can either by created or modified from Code Corps or GitHub (and in the future from other integrations, too).
- We will add a `created_at` and `modified_at` timestamp to `Task`.

landongrindheim:
# What's in this PR?
Timestamps (`task_created_at` and `task_updated_at`) have been _added_ to Task. When `Task.update_changeset/2` is called, `task_updated_at` is updated. Likewise, when `Task.create_changeset/2` is called, `task_created_at` is updated.

#888 specified adding the new timestamp fields, which I understood to mean in addition to the default fields. If the defaults should be renamed, please let me know. Likewise, there was some inconsistency with a name. In one instance one of the new fields was referred to as `task_created_at`, while in another place it was referred to as `task_inserted_at`. I opted for the former, but I can do either one.

## References
Fixes #888 (first portion)

Progress on: #888, #878

### DIFF
--- a/lib/code_corps/model/task.ex
+++ b/lib/code_corps/model/task.ex
@@ -15,6 +15,8 @@ defmodule CodeCorps.Task do
     field :status, :string, default: "open"
     field :title, :string
     field :github_issue_number, :integer
+    field :task_created_at, :naive_datetime
+    field :task_updated_at, :naive_datetime
 
     field :position, :integer, virtual: true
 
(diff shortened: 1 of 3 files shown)


joshsmith:
@landongrindheim did you see the comments I posted over in #888? Could use your thoughts on that.

landongrindheim:
@JoshSmith I'm closing this as per conversation in #888

joshsmith:
Thanks! Apologies again for the flux on this one. Did you want to tackle the new one? If not, I can prioritize it myself.

landongrindheim:
I won't claim it as I don't have a block of time this week to explore the issue in depth

joshsmith:
Okay, thanks for the heads up!

Issue #940, opened 2017-09-15
TaskSyncer.sync_all can return error changesets that get swallowed by our GitHub issues event handler without error


begedin:
# Problem

Quoting https://github.com/code-corps/code-corps-api/pull/805#issuecomment-329641688

>I got an `opened` event for `issues` (created by a GH user that had no corresponding user on CC) that resulted in no issue being created, a 200 returned, a `processed` status, and no user created on CC. That means the flow is broken and it's not clear where it's breaking down.

We need to go through such a process and verify that it goes through correctly and a user is actually created.

joshsmith:
This is due to a bug in `GitHub.Event.Issues`. Inside the `do_handle/1` we do the following:

```elixir
    case Repo.transaction(multi) do
      {:ok, %{tasks: tasks}} -> {:ok, tasks} # This is our bug
      {:error, :repo, :unmatched_project, _steps} -> {:ok, []}
      {:error, _errored_step, error_response, _steps} -> {:error, error_response}
    end
```

The `{:ok, %{tasks: tasks}}` also happily pattern matches against tasks that `TaskSyncer.sync_all` returns regardless of whether `tasks` might be a `List` of, for example, `Ecto.Changeset`s with errors.

> A side note: I think `sync_all` implies things it does not, i.e. that we're syncing all the tasks for a given user and project's GitHub repos.

In this case, the bug is triggered _specifically_ by the markdown from GitHub being blank (it can be! so we found _two_ bugs here). Thus `do_handle` returns:

```elixir
{:ok,
 [#Ecto.Changeset<action: :insert,
   changes: %{github_issue_number: 13, github_repo_id: 17, project_id: 1,
     task_list_id: 1, title: "Another test account without a Code Corps user",
     user_id: 68},
   errors: [markdown: {"can't be blank", [validation: :required]},
    body: {"can't be blank", [validation: :required]}], data: #CodeCorps.Task<>,
   valid?: false>]}
```

My best guess is that we do not, in fact, want `sync_all` to return `{:ok, list(Task.t)}` in the event that one of these tasks has an error changeset. Probably the responsibility lies squarely there to:

- perhaps return an `error` tuple of some sort with some reasoning
- perhaps perform some logging (with some context)
- perhaps explicitly raise an error (with some context)

...or some combination of the above.

[...]

PR #963, opened 2017-09-21
Loosen markdown restrictions on Comment and Task

### LINKED ISSUE #962: Remove markdown restriction because GitHub Issues and Comments may have blank markdown
# Problem

GitHub issues and comments can have `markdown` (in terms of our `Task` and `Comment` models) that are blank/empty/nil etc.

We validate `markdown` coming from a user on Code Corps because we don't want to allow for blank task descriptions. This is a contributor experience concern.

I think, however, that we may want to remove the restriction on the API side altogether, and open an issue on the frontend to do that validation only in certain circumstances (from the `/tasks/new` route, for example). This would fix the bug raised in #940.

> NB: this would also prevent future bugs from occurring in new contexts, like the possibility of us adding a task list UI within some future project milestones UI. You may just want to spike out what's needed for a feature without writing the descriptions just yet.

We should also probably see if there are other fields our changesets are validating for other GitHub-related data and verify that GitHub will always return them the way we expect, else open new issues for those.

landongrindheim:
`NOT NULL` constraint from the comments table. Requirement validations for `markdown` on the `Task` and `Comment` models have been moved. Tests that were based on those requirements were removed.

## References
Fixes #962 

Progress on: #940

### DIFF
--- a/lib/code_corps/model/comment.ex
+++ b/lib/code_corps/model/comment.ex
@@ -22,7 +22,6 @@ defmodule CodeCorps.Comment do
   def changeset(struct, params \\ %{}) do
     struct
     |> cast(params, [:markdown])
-    |> validate_required([:markdown])
     |> MarkdownRendererService.render_markdown_to_html(:markdown, :body)
   end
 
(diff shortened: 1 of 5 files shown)


joshsmith (dismissed):

joshsmith (approved):

joshsmith:
Excellent work on this @landongrindheim. Looks great 👍

PR #967, opened 2017-09-25
Fix user linking of github issues and comments to account for bot user

### LINKED ISSUE #939: Handling bot user when handling github webhook
# Problem

If a CodeCorps user who is not connected to github creates a task on codecorps which is connected to github this triggers the following process.

1. Task is created
2. Github issue is created on selected repo, authored by the github application bot user
3. Github webhook for `issue_created` is triggered and handled from our end
4. When handling the webhook, the bot user record is created on codecorps, since there's no matching actual user.

Step 4 should not happen, or should be reconsidered, at least.

The problem is in `github/event/issues.ex`

The current implementation is 

```Elixir
Multi.new
|> Multi.run(:repo, fn _ -> RepoFinder.find_repo(payload) end)
|> Multi.run(:user, fn _ -> UserLinker.find_or_create_user(payload) end)
|> Multi.run(:tasks, fn %{repo: github_repo, user: user} -> TaskSyncer.sync_all(github_repo, user, payload) end)
```

This works then the task creator is not the bot. It however, does not work when it is the bot, since in that case, we will already have a matching task record which is associated to an actual, github unconnected user.

## Scenarios

The most complicated scenario is

- `org_1/project` and `org_2/project` are both connected to `repo_1`
- `user_1` is member of `org_1`, but not `org_2`, they are not github connected
. `user_1` creates `task_1` on `org_1/project`
- this causes `issue_1` to be created on `repo_1`
- this causes an `issues` webhook to be fired, by `app[bot] user`
- we handle the webhook
    - `org_1/project/task_1` already exist, so we should find it and "update" it, the user should remain `user_1`
    - `org_2/project/task_1` does not exist yet, so it should be created. `user_1` is not a member of `org_2`, so the user assigned to the task should be the bot user? (OPTION A) Or should we still assign `user_1` and allow him to edit that specific task on code corps to, even though they aren't a member of `org_2`? (OPTION B)

Other scenarios are 

- there's just one project linked to the repo (easy, find and update the task)
- there are multiple projects linked, but the user is a member of all organizations (find and update initial task, create other ones, associate with user)

#### if issue user is the bot and we went with option A

- get task identifier (`github_repo_name`, `github_repo_owner`, `github_issue_number`)
- find task for that identifier, there needs to be one, "update" it (probably no changes)
- for each other project linked to that repo, create or update task and assign to project, repo and the bot user

#### if issue user is the bot and we went with option B

- get task identifier (`github_repo_name`, `github_repo_owner`, `github_issue_number`)
- find task for that identifier, there needs to be one, "update" it (probably no changes)
- get user for that task
- for each other project linked to that repo, create or update task and assign to project, repo and the initial task user
 
### if issue user is not the bot

Basically, use the current flow

- get repo
- find or create user
- for each project associated with repo
  - create task, associate with project, repo and user

### Notes 

@JoshSmith  I'm marking this as "question", since it probably needs further discussion or at least a review of the process I'm proposing.

begedin:
Closes #939

The changes are constrained to the `UserLinker` module and I've added detailed documentation on how it works. I tried to make it as intuitive as possible.


joshsmith:
After some offline conversation about this, it looks like we want to allow bot users to be created, but prevent bot comments from being created when a task likely already exists for that user. This PR does the latter but not the former.

To fix this, we'll need to:

- [x] Add a `type` to `User` that can be `user` or `bot`
- [x] Add to the `Accounts.create_from_github/1` to cast the type from GitHub and validate the changeset
- [x] Add tests for the `Accounts.create_from_github/1` cases
- [x] Match against the `type` in the user linker
- [x] Remove the `:user_not_found` case

joshsmith:
@begedin this should be good for a "review" by you, although I know I need to be the one to approve.

begedin:
@JoshSmith I looked at the changes, added some tests and comments and cleaned up a few things I missed on my first go. 

Should be good for review/merge now

joshsmith (dismissed):
Looks good pending that migration getting into `structure.sql`.

joshsmith (dismissed):

joshsmith (approved):

PR #968, opened 2017-09-25
Switch StripeConnectAccountController away from ja_resource and canary

### LINKED ISSUE #905: Switch StripeConnectAccountController away from ja_resource and canary
Progress on #864 

# Problem

# Procedure

- pick a controller to switch
- for each action

```Elixir
  end
end

  end
end

  end
end

  end
end
```

### `index` action

- render data

### `show` action

- load resource
- render resource

### `create` action

- perform insert action
- render resulting resource

- load resource
- perform update/delete

### Examples in code

## How to rewrite policies

```Elixir
```

becomes

```Elixir
```

### Examples in code

treble37:
Closes #905

Removes reliance and ja_resources and canary 

## References
Fixes # 905

Progress on: # 865

### DIFF
--- a/lib/code_corps/policy/policy.ex
+++ b/lib/code_corps/policy/policy.ex
@@ -55,6 +55,9 @@ defmodule CodeCorps.Policy do
   defp can?(%User{} = current_user, :delete, %UserSkill{} = user_skill, %{}), do: Policy.UserSkill.delete?(current_user, user_skill)
   defp can?(%User{} = current_user, :create, %UserRole{} = user_role, %{}), do: Policy.UserRole.create?(current_user, user_role)
   defp can?(%User{} = current_user, :delete, %UserRole{} = user_role, %{}), do: Policy.UserRole.delete?(current_user, user_role)
+  defp can?(%User{} = user, :show, %StripeConnectAccount{} = stripe_connect_account, %{}), do: Policy.StripeConnectAccount.show?(user, stripe_connect_account)
+  defp can?(%User{} = user, :create, %StripeConnectAccount{}, %{} = params), do: Policy.StripeConnectAccount.create?(user, params)
+  defp can?(%User{} = user, :update, %StripeConnectAccount{} = stripe_connect_account, %{}), do: Policy.StripeConnectAccount.update?(user, stripe_connect_account)
 
   defimpl Canada.Can, for: User do
     # NOTE: Canary sets an :unauthorized and a :not_found handler on a config level
@@ -91,10 +94,6 @@ defmodule CodeCorps.Policy do
     def can?(%User{} = user, :delete, %RoleSkill{}), do: Policy.RoleSkill.delete?(user)
 
     def can?(%User{} = user, :create, Skill), do: Policy.Skill.create?(user)
-
-    def can?(%User{} = user, :show, %StripeConnectAccount{} = stripe_connect_account), do: Policy.StripeConnectAccount.show?(user, stripe_connect_account)
-    def can?(%User{} = user, :create, %Changeset{ data: %StripeConnectAccount{}} = changeset), do: Policy.StripeConnectAccount.create?(user, changeset)
-    def can?(%User{} = user, :update, %StripeConnectAccount{} = stripe_connect_account), do: Policy.StripeConnectAccount.update?(user, stripe_connect_account)
 
     def can?(%User{} = user, :show, %StripeConnectPlan{} = stripe_connect_plan), do: Policy.StripeConnectPlan.show?(user, stripe_connect_plan)
     def can?(%User{} = user, :create, %Changeset{ data: %StripeConnectPlan{}} = changeset), do: Policy.StripeConnectPlan.create?(user, changeset)
--- a/lib/code_corps/policy/stripe_connect_account.ex
+++ b/lib/code_corps/policy/stripe_connect_account.ex
@@ -6,8 +6,8 @@ defmodule CodeCorps.Policy.StripeConnectAccount do
   def show?(%User{} = user, %StripeConnectAccount{} = stripe_connect_account),
     do: stripe_connect_account |> get_organization() |> owned_by?(user)
 
-  def create?(%User{} = user, %Ecto.Changeset{} = changeset),
-    do: changeset |> get_organization() |> owned_by?(user)
+  def create?(%User{} = user, %{} = params),
+    do: params |> get_organization() |> owned_by?(user)
 
   def update?(%User{} = user, %StripeConnectAccount{} = stripe_connect_account),
     do: stripe_connect_account |> get_organization() |> owned_by?(user)
--- a/lib/code_corps_web/controllers/stripe_connect_account_controller.ex
+++ b/lib/code_corps_web/controllers/stripe_connect_account_controller.ex
@@ -1,27 +1,49 @@
 defmodule CodeCorpsWeb.StripeConnectAccountController do
   use CodeCorpsWeb, :controller
-  use JaResource
 
   alias CodeCorps.ConnUtils
   alias CodeCorps.StripeConnectAccount
   alias CodeCorps.StripeService.StripeConnectAccountService
+  alias CodeCorps.User
 
-  plug :load_and_authorize_changeset, model: StripeConnectAccount, only: [:create]
-  plug :load_and_authorize_resource, model: StripeConnectAccount, only: [:show, :update]
-  plug JaResource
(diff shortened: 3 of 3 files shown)


treble37:
I think I need some help for the 2 tests that are failing locally...

lib/code_corps_web/controllers/stripe_connect_account_controller.ex
```diff
@@ -1,27 +1,47 @@
...
+  plug CodeCorpsWeb.Plug.DataToAttributes
 
   @spec model :: module
   def model, do: CodeCorps.StripeConnectAccount
 
-  def handle_create(conn, attributes) do
-    attributes
-    |> Map.put("tos_acceptance_ip", conn |> ConnUtils.extract_ip)
-    |> Map.put("tos_acceptance_user_agent", conn |> ConnUtils.extract_user_agent)
-    |> Map.put("managed", true)
```
begedin: You need to do the same to `params` below, to get this passing correctly.

lib/code_corps/policy/policy.ex
```diff
@@ -51,6 +51,9 @@ defmodule CodeCorps.Policy do
   defp can?(%User{} = current_user, :delete, %UserSkill{} = user_skill, %{}), do: Policy.UserSkill.delete?(current_user, user_skill)
   defp can?(%User{} = current_user, :create, %UserRole{} = user_role, %{}), do: Policy.UserRole.create?(current_user, user_role)
   defp can?(%User{} = current_user, :delete, %UserRole{} = user_role, %{}), do: Policy.UserRole.delete?(current_user, user_role)
+  defp can?(%User{} = user, :show, %StripeConnectAccount{} = stripe_connect_account, params), do: Policy.StripeConnectAccount.show?(user, params)
```
begedin: This should be 

```Elixir
defp can?(%User{} = user, :show, %StripeConnectAccount{} = stripe_connect_account, %{}), do: Policy.StripeConnectAccount.show?(user, stripe_connect_account)
```

Once you make that change, you can remove the equivalent from line 97

lib/code_corps/policy/policy.ex
```diff
@@ -51,6 +51,9 @@ defmodule CodeCorps.Policy do
...
+  defp can?(%User{} = user, :create, %StripeConnectAccount{}, params), do: Policy.StripeConnectAccount.create?(user, params)
```
begedin: `%{} = params`, so we typecheck against a map.

lib/code_corps/policy/policy.ex
```diff
@@ -51,6 +51,9 @@ defmodule CodeCorps.Policy do
...
+  defp can?(%User{} = user, :update, %StripeConnectAccount{} = stripe_connect_account, params), do: Policy.StripeConnectAccount.update?(user, stripe_connect_account)
```
begedin: `%{}` here, since we aren't using the params, but still want to typecheck.

You should be getting compiler warnings for these, so be sure to get those eliminated.

lib/code_corps/policy/policy.ex
```diff
@@ -92,7 +95,7 @@ defmodule CodeCorps.Policy do
     def can?(%User{} = user, :create, Skill), do: Policy.Skill.create?(user)
 
     def can?(%User{} = user, :show, %StripeConnectAccount{} = stripe_connect_account), do: Policy.StripeConnectAccount.show?(user, stripe_connect_account)
-    def can?(%User{} = user, :create, %Changeset{ data: %StripeConnectAccount{}} = changeset), do: Policy.StripeConnectAccount.create?(user, changeset)
+    def can?(%User{} = user, :create, %StripeConnectAccount{}, params), do: Policy.StripeConnectAccount.create?(user, params)
     def can?(%User{} = user, :update, %StripeConnectAccount{} = stripe_connect_account), do: Policy.StripeConnectAccount.update?(user, stripe_connect_account)
```
begedin: The end result is that you want to remove lines 97, 98 and 99 in favor of lines 54, 55 and 56

lib/code_corps/policy/stripe_connect_account.ex
```diff
@@ -3,11 +3,11 @@ defmodule CodeCorps.Policy.StripeConnectAccount do
 
   alias CodeCorps.{StripeConnectAccount, User}
 
-  def show?(%User{} = user, %StripeConnectAccount{} = stripe_connect_account),
-    do: stripe_connect_account |> get_organization() |> owned_by?(user)
+  def show?(%User{} = user, params),
+    do: params |> get_organization() |> owned_by?(user)
```
begedin: This was an unnecessary change here. `show?` should still accept an account as a second argument.

lib/code_corps/policy/stripe_connect_account.ex
```diff
@@ -3,11 +3,11 @@ defmodule CodeCorps.Policy.StripeConnectAccount do
...
 
-  def create?(%User{} = user, %Ecto.Changeset{} = changeset),
-    do: changeset |> get_organization() |> owned_by?(user)
+  def create?(%User{} = user, params),
```
begedin: `%{} = params` for the same reason as above.

lib/code_corps_web/controllers/stripe_connect_account_controller.ex
```diff
@@ -1,27 +1,47 @@
...
   alias CodeCorps.ConnUtils
   alias CodeCorps.StripeConnectAccount
   alias CodeCorps.StripeService.StripeConnectAccountService
+  alias CodeCorps.User
 
-  plug :load_and_authorize_changeset, model: StripeConnectAccount, only: [:create]
-  plug :load_and_authorize_resource, model: StripeConnectAccount, only: [:show, :update]
-  plug JaResource
+  action_fallback CodeCorpsWeb.FallbackController
+  plug CodeCorpsWeb.Plug.DataToAttributes
```
begedin, line 10: don't forget `plug CodeCorpsWeb.IdsToIntegers` to

begedin (changes requested):

lib/code_corps_web/controllers/stripe_connect_account_controller.ex
treble37: I think I fixed all the issues now, thanks @begedin

lib/code_corps/policy/policy.ex
```diff
@@ -51,6 +51,9 @@ defmodule CodeCorps.Policy do
   defp can?(%User{} = current_user, :delete, %UserSkill{} = user_skill, %{}), do: Policy.UserSkill.delete?(current_user, user_skill)
   defp can?(%User{} = current_user, :create, %UserRole{} = user_role, %{}), do: Policy.UserRole.create?(current_user, user_role)
   defp can?(%User{} = current_user, :delete, %UserRole{} = user_role, %{}), do: Policy.UserRole.delete?(current_user, user_role)
+  defp can?(%User{} = user, :show, %StripeConnectAccount{} = stripe_connect_account, %{} = params), do: Policy.StripeConnectAccount.show?(user, stripe_connect_account)
```
begedin: `%{} = params` should throw out a compile warning when you run the tests, because `params` is not being used. Is it not happening for you. `%{}` should be correct.

lib/code_corps/policy/policy.ex
```diff
@@ -51,6 +51,9 @@ defmodule CodeCorps.Policy do
...
+  defp can?(%User{} = user, :create, %StripeConnectAccount{}, %{} = params), do: Policy.StripeConnectAccount.create?(user, params)
+  defp can?(%User{} = user, :update, %StripeConnectAccount{} = stripe_connect_account, %{} = params), do: Policy.StripeConnectAccount.update?(user, stripe_connect_account)
```
begedin: Same here with `%{} = params` -> `%{}`

begedin (dismissed):
Just two more very minor details and this is good to go. Good work @treble37

begedin:
Another reminder to rebase and squash as you add the fixes @treble37, not that you haven't been doing that correctly up to this point. I just find it's easier to remind, than to wait another day (due to time difference) to merge something that's otherwise 👍

lib/code_corps/policy/policy.ex
treble37: @begedin - I think I missed it because it throws out warnings only when I open a terminal and do `mix test` the first time....subsequent runs of mix test don't show the `params is unused warning`.....

treble37:
Thanks for the kind reminder, should be rebased and squashed, thanks!

joshsmith (approved):

PR #974, opened 2017-09-26
Make task and comment syncers fail on validation errors

### LINKED ISSUE #940: TaskSyncer.sync_all can return error changesets that get swallowed by our GitHub issues event handler without error

begedin:
Closes #940 

The changes here are restricted to `TaskSyncer` and `CommentSyncer` and I used a recursive list approach, as discussed in the issue.


joshsmith (dismissed):
Looks good. Is there anywhere else you think we should add this?

joshsmith (approved):

PR #977, opened 2017-09-26
Propagate comment changes to GitHub comments

joshsmith:
# What's in this PR?

This somehow was *not* merged properly before into anything. I'm not quite sure what happened.


joshsmith:
@begedin I have since had to address bugs that cropped up in this and were not caught in tests. We need to test this out against actual API endpoints before things go up for code review. Mocks only go so far and this was a case where the mocks were lying to us about what the endpoints actually looked like.

This should now work and be ready for review.

joshsmith:
There's an argument to be made here that we should be doing some actual integration tests.

joshsmith:
If this looks good, please do squash into a single commit and then merge.

begedin (approved):

PR #982, opened 2017-09-27
Fix minor issues with seeds

joshsmith:
# What's in this PR?

Fixes some minor issues with the way seeds worked previously.

### DIFF
--- a/priv/repo/seeds.exs
+++ b/priv/repo/seeds.exs
@@ -1,6 +1,6 @@
 alias CodeCorps.{
   Category, Organization, ProjectUser, Project, ProjectCategory, ProjectSkill,
-  Repo, Role, Skill, SluggedRoute, Task, TaskList, User, UserCategory, UserRole,
+  Repo, Role, Skill, SluggedRoute, Task, User, UserCategory, UserRole,
   UserSkill
 }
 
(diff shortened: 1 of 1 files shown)


priv/repo/seeds.exs
```diff
@@ -50,37 +50,42 @@ case Repo.all(User) do
...
 case Repo.all(Organization) do
   [] ->
     organizations
-    |> Enum.map(&Repo.insert!/1)
-    |> Enum.each(fn %{id: id, slug: slug} ->
-      %SluggedRoute{organization_id: id, slug: slug} |> Repo.insert!()
+    |> Enum.each(fn params ->
+      %Organization{}
+      |> Organization.create_changeset(params)
+      |> Repo.insert!()
```
begedin, line 66: At some point a couple of months back, I intentionally changed the seeds so they do not rely on our changesets, reason being, they were often breaking as we introduced changes to the codebase, and our tests do not cover seed failures, so we ended up merging them in in a broken state.

I basically followed the same reasoning as we do with migrations, except there, it's problematic even to rely on schemas.

I'm not strongly against using changesets in seeds, but I thought I should point out this potential issue.

joshsmith: Sure, I understand what you mean here but there are some setup things happening in changesets that we would have to reinvent the wheel with in seeds to replicate. That the application undergoes churn does not lead to the conclusion, in my opinion, that the seeds can be isolated from that churn in some way. Instead, we should consider ways to ensure the seeds are working, perhaps by adding some smoke tests. I'll look into ways we can handle this.

Do you mind opening an issue to address this separately?

PR #984, opened 2017-09-27
Added field closed_at for Tasks and setup changeset accordingly

### LINKED ISSUE #979: Add a `closed_at` to `Task` to track when task is closed
# Problem

Add a `closed_at` to `Task` so we can track when a task was closed (i.e. completed).  Could use this in a project's update emails to their donors about what's been done in the past month.

vishaldeepak:
# What's in this PR?
Also found a bug in a test case that I fixed

## References
Fixes #979.

### DIFF
--- a/lib/code_corps/model/task.ex
+++ b/lib/code_corps/model/task.ex
@@ -4,10 +4,12 @@ defmodule CodeCorps.Task do
   import EctoOrdered
 
   alias CodeCorps.Services.MarkdownRendererService
+  alias Ecto.Changeset
 
   @type t :: %__MODULE__{}
 
   schema "tasks" do
+    field :closed_at, :utc_datetime
     field :body, :string
     field :markdown, :string
     field :number, :integer, read_after_writes: true
@@ -58,6 +60,7 @@ defmodule CodeCorps.Task do
     |> changeset(params)
     |> cast(params, [:status])
     |> validate_inclusion(:status, statuses())
+    |> set_closed_at()
   end
 
   def apply_position(changeset) do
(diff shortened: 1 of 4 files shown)


test/lib/code_corps/model/task_test.exs
```diff
@@ -34,8 +34,24 @@ defmodule CodeCorps.TaskTest do
   describe "update_changeset/2" do
     test "only allows specific values for status" do
       changes = Map.put(@valid_attrs, :status, "nonexistent")
-      changeset = Task.update_changeset(%Task{}, changes)
+      changeset = Task.update_changeset(%Task{task_list_id: 1}, changes)
```
vishaldeepak, line 37: Not setting a `task_list_id` would refute the changeset anyway, making the test invalid

lib/code_corps/model/task.ex
```diff
@@ -4,10 +4,12 @@ defmodule CodeCorps.Task do
   import EctoOrdered
 
   alias CodeCorps.Services.MarkdownRendererService
+  alias Ecto.Changeset
 
   @type t :: %__MODULE__{}
 
   schema "tasks" do
+    field :closed_at, :utc_datetime
```
joshsmith, line 12: I'm not sure I've seen utc_datetime before. Is this new in ecto?

lib/code_corps/model/task.ex
```diff
@@ -58,6 +60,7 @@ defmodule CodeCorps.Task do
     |> changeset(params)
     |> cast(params, [:status])
     |> validate_inclusion(:status, statuses())
+    |> set_closed_at()
```
joshsmith, line 63: I understand why you placed this in update but I do think since it's possible we'll need to write another issue to address this with GitHub sync. Can you open such an issue? We'd need to handle the closed event on GitHub and propagate that close to here. This could happen on create so I want to be sure we're not using the create changeset here.

joshsmith (dismissed):

joshsmith (approved):

lib/code_corps/model/task.ex
vishaldeepak: The old `:datetime` field now shows deprecation warning on migration.  We need to use either `naive_datetime` or `utc_datetime`.

lib/code_corps/model/task.ex
vishaldeepak: Sure raised an issue. But I am unaware of GithubSync feel free to edit the issue -https://github.com/code-corps/code-corps-api/issues/992

PR #987, opened 2017-09-27
Fix a number of warnings and change log levels

joshsmith:
# What's in this PR?

- Fixes warnings about migrations relying on `:datetime`. I saw no harm in changing them to `:utc_datetime` instead.
- Changes `Logger.warn` to `Logger.info` in a few places
- Removes a few compiler warnings and does some quick reformatting where needed

### DIFF
--- a/priv/repo/migrations/20161019090945_add_stripe_customers_cards_tables.exs
+++ b/priv/repo/migrations/20161019090945_add_stripe_customers_cards_tables.exs
@@ -3,7 +3,7 @@ defmodule CodeCorps.Repo.Migrations.AddStripeCustomersCardsTables do
 
   def change do
     create table(:stripe_customers) do
-      add :created, :datetime
+      add :created, :utc_datetime
       add :currency, :string
       add :delinquent, :boolean
       add :email, :string
--- a/lib/code_corps/comment/service.ex
+++ b/lib/code_corps/comment/service.ex
@@ -42,8 +42,8 @@ defmodule CodeCorps.Comment.Service do
   defp marshall_result({:ok, %{github: %Comment{} = comment}}), do: {:ok, comment}
   defp marshall_result({:error, :comment, %Changeset{} = changeset, _steps}), do: {:error, changeset}
   defp marshall_result({:error, :github, result, _steps}) do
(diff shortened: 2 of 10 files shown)


joshsmith:
It works! Only warning (compiler or otherwise) comes from upstream dep:

Terminal output

priv/repo/migrations/20161019090945_add_stripe_customers_cards_tables.exs
```diff
@@ -3,7 +3,7 @@ defmodule CodeCorps.Repo.Migrations.AddStripeCustomersCardsTables do
 
   def change do
     create table(:stripe_customers) do
-      add :created, :datetime
+      add :created, :utc_datetime
```
snewcomer, line 6: @JoshSmith no need to replace `field :created, :utc_datetime` as well in the model?

snewcomer: One thing I have seen done is for other models
```
@timestamps_opts [type: :utc_datetime] in the model file.
```

snewcomer (commented):
Nice @JoshSmith!

Just for reference, this is a really good article if you haven't seen it.

https://www.amberbit.com/blog/2017/8/3/time-zones-in-postgresql-elixir-and-phoenix/

priv/repo/migrations/20161019090945_add_stripe_customers_cards_tables.exs
joshsmith: Would you mind opening an issue for this? I'd like to migrate everything over to it specifically. We're lucky in the sense that the servers are all on UTC time by default so it shouldn't be a pain to migrate. But this is something that needs done ASAP.

joshsmith: For now, I'm just trying to fix compiler warnings that will _always_ be there unless we fix old migrations. That one was from Oct 2016!

snewcomer: Cool.  I'll open an issue.

snewcomer (approved):

PR #990, opened 2017-09-28
Fix bug introduced into StripeConnectAccountController that does not update on Stripe

joshsmith:
# What's in this PR?

A PR accidentally introduced a bug that ensures our Stripe Connect accounts don't get updated on Stripe itself.

The following PR fixes this.


joshsmith:
This still needs tests. I didn't have time before I wrapped up.

joshsmith:
I actually do not have any way to test this that I can think of. Just going to merge to fix this for now.

PR #997, opened 2017-09-28
Fix email overwriting by GitHub

### LINKED ISSUE #980: GitHub email overwrites user's email upon connecting
# Problem

When the GitHub account connection process occurs (via `/oauth/github`), the user's email gets overwritten by the `GitHub.Adapters.User.from_github_user/1` function.

We should not be overwriting a user-provided email with an automatically provided one.

We need to identify if there are other places the GitHub user adapter is overwriting the user's email unexpectedly.

joshsmith:
# What's in this PR?

This PR:

- fixes email overwriting by GitHub when adding the OAuth app
- moves processing of user account merging out of `GitHub.User` and into `CodeCorps.Accounts`
- adds some tests

## References
Fixes #980

### DIFF
--- a/lib/code_corps/accounts/accounts.ex
+++ b/lib/code_corps/accounts/accounts.ex
@@ -2,15 +2,21 @@ defmodule CodeCorps.Accounts do
   @moduledoc ~S"""
   Main entry-point for managing accounts.
 
-  All actions to acounts should go through here.
+  All actions to accounts should go through here.
   """
 
   alias CodeCorps.{
+    Accounts.Changesets,
+    Comment,
     GitHub.Adapters,
+    GithubAppInstallation,
     User,
-    Repo
+    Repo,
+    Task
   }
-  alias Ecto.Changeset
+  alias Ecto.{Changeset, Multi}
+
+  import Ecto.Query
 
   @doc ~S"""
   Creates a user record using attributes from a GitHub payload.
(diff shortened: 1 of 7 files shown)


priv/repo/structure.sql
```diff
@@ -1524,7 +1524,7 @@ ALTER SEQUENCE user_tasks_id_seq OWNED BY user_tasks.id;
 --
 
 CREATE TABLE users (
-    id bigint NOT NULL,
+    id integer NOT NULL,
```
begedin: You should probably do a `mix ecto.reset`. This is reverting the ecto change pushed through one of my recent PRs.

lib/code_corps/accounts/accounts.ex
```diff
@@ -29,8 +34,92 @@ defmodule CodeCorps.Accounts do
...
+  end
+
+  @doc ~S"""
+  Casts a changeset used for creating a user account from a github user payload
+  """
+  @spec update_from_github_oauth_changeset(struct, map) :: Changeset.t
+  def update_from_github_oauth_changeset(struct, %{} = params) do
+    struct
+    |> Changeset.cast(params, [:github_auth_token, :github_avatar_url, :github_id, :github_username, :type])
+    |> Changeset.validate_required([:github_auth_token, :github_avatar_url, :github_id, :github_username, :type])
```
begedin: Are we not setting the email at all then? 

My understanding was, we set the user email from github if there's no email already.

I can see the reasoning behind not wanting to set it at all, but I want to make sure.

begedin (changes requested):
A minor change and a minor question, but otherwise, looks good.

lib/code_corps/accounts/accounts.ex
joshsmith: In this case, where the user is connecting their GitHub account via OAuth, the assumption is that there already is an email. I can see how maybe we would want to not assume that and pull it and set it if it's not yet set. I'm not quite sure how to do that, though.

priv/repo/structure.sql
joshsmith: I have done one a number of times. This change seems to be happening at random.

lib/code_corps/accounts/accounts.ex
begedin:
```Elixir
def ensure_email(%Changeset{} = changeset, %{"email" => email} = _params do
  case changeset |> Changeset.get_field(:email) do
    nil -> changeset |> Changeset.put_change(:email, email)
    _email -> changeset
  end
end
```

Something like that. `get_field/2` returns the value of the `get_change/2` if there is a change, or of the underlying struct field, if there is no change to it.

Basically, we set the email from github, only if the current email is nil and is not being changed to something else already. This covers both the case of using this changeset directly, as well as potentially using it as part of another changeset.

joshsmith: @begedin I'm not sure I see how this works if someone has their email set in the database though. This only pulls from the changeset and params, no?

begedin: It pulls from the changes, then from the database field if nothing is found in changes. 

```Elixir
iex> %Changeset{data: %User{email: "foo"}, changes: %{email: "bar"}} |> get_field(:email)

"bar"

iex> %Changeset{data: %User{email: "foo"}, changes: %{}} |> get_field(:email)

"foo"

iex> %Changeset{data: %User{email: nil}, changes: %{email: "bar"}} |> get_field(:email)

"bar"

iex> %Changeset{data: %User{email: nil}, changes: %{}} |> get_field(:email)

nil
```

Example 3 won't really happen here, but example 4 is when we want to set the email from params in our case - the database field is currently `nil`.

lib/code_corps/accounts/accounts.ex
```diff
@@ -29,8 +34,107 @@ defmodule CodeCorps.Accounts do
   def create_from_github_changeset(struct, %{} = params) do
```
begedin: This is not really part of this PR, but it is something that we've discussed before and I feel should be tackled.

`create_from_github_changeset/2` is public. The only place it's being called from publically is another public method in this module (`create_from_github/1`) and tests.

That means, it is in fact private and we made it public in order to test it.

The conventional approach if a private method is complex enough to warrant testing would be to extract the method's behavior into a submodule.

`Accounts.Changesets` would sound like a good candidate for that.

lib/code_corps/accounts/accounts.ex
```diff
@@ -29,8 +34,107 @@ defmodule CodeCorps.Accounts do
...
+    |> Changeset.put_change(:sign_up_context, "github")
     |> Changeset.unique_constraint(:email)
     |> Changeset.validate_inclusion(:type, ["bot", "user"])
   end
+
+  @doc ~S"""
+  Ensures an email is set without overwriting an existing email.
+  """
+  @spec ensure_email(Changeset.t, map) :: Changeset.t
+  def ensure_email(%Changeset{} = changeset, %{"email" => new_email} = _params) do
```
begedin: The reason for the comment above is because we do the same here

The only place we call `ensure_email/2` from is `update_from_github_changeset/2`. The only reason it's public is so we can test it directly. 

`Accounts.Changesets` would, again, be a good candidate to do this instead.

lib/code_corps/accounts/accounts.ex
```diff
@@ -29,8 +34,107 @@ defmodule CodeCorps.Accounts do
...
+      {:error, :user, %Changeset{} = changeset, _actions_done} ->
+        {:error, changeset}
+    end
+  end
+
+  @doc ~S"""
+  Casts a changeset used for creating a user account from a github user payload
+  """
+  @spec update_from_github_oauth_changeset(struct, map) :: Changeset.t
+  def update_from_github_oauth_changeset(struct, %{} = params) do
```
begedin: Again, if we put `update_from_github_oauth_changeset/2` is public, even though, outside of tests, it's only ever called internally from this module.

Moving it into `Accounts.Changesets`, we can move `ensure_email/2` there as well, make it private and then simply test `update_from_github_changeset/2` instead, since we really only care about the outcome of that one, not the "subhelper" `ensure_email`.

begedin (dismissed):
I think some of the code needs to be moved around a bit, since as it is right now, it's being constructed to conform tests, instead of actual use cases.

joshsmith:
I don't like merging without another look but I think I've addressed everything here @begedin. If you'd like to give it a look again after the fact as just a safety measure, then that would be nice. But I'm going to go ahead and merge and move onto something else.

begedin (commented):
All looks good to me

PR #998, opened 2017-09-29
Modify Task timestamps for GitHub sync

### LINKED ISSUE #888: Modify Task timestamps to accommodate github sync

joshsmith:
# What's in this PR?

Currently a work in progress. Could use feedback on it so far, though.

## References
Fixes #888


lib/code_corps/github/adapters/task.ex
```diff
@@ -26,6 +28,8 @@ defmodule CodeCorps.GitHub.Adapters.Task do
     task
     |> Map.from_struct
     |> MapTransformer.transform_inverse(@mapping)
+    |> Map.delete("created_at")
     |> Map.delete("number")
+    |> Map.delete("updated_at")
```
begedin: Maybe a function `clear_autogenerated_github_keys`, which removes all of these keys, for added explicitness and self-documenting?

begedin (dismissed):
A minor suggestion for better self-documentation of the adapter code, but past that point, I think this is good to go.

joshsmith:
I'm not sure this is fully ready but not sure what's missing. Doesn't look like a test for changeset builder and not sure what else.

begedin:
I added a couple of missing comments and cleaned up a few minor details. I think this is good to merge.

begedin (dismissed):

PR #1010, opened 2017-10-03
Modify Comment timestamps for GitHub sync

### LINKED ISSUE #887: Modify Comment timestamps to accommodate github sync
- Comments can be multi-sourced. They can either by created or modified from Code Corps or GitHub (and in the future from other integrations, too).
- We will add a `created_at` and `modified_at` timestamp to `Comment`.

joshsmith:
# What's in this PR?

Modifies comment timestamps to be able to accept GitHub sync.

This also addresses `Task` actually getting the `created_from` and `modified_from` values set to `github` when the task is created and modified from there.

## References
Fixes #887


lib/code_corps/github/event/issues/changeset_builder.ex
```diff
@@ -42,13 +42,16 @@ defmodule CodeCorps.GitHub.Event.Issues.ChangesetBuilder do
     %ProjectGithubRepo{project_id: project_id, github_repo_id: github_repo_id},
     %User{id: user_id}) do
 
+      IO.inspect("hello")
```
begedin: Don't forget to remove

begedin (dismissed):
You missed a debugger statement. Otherwise looks good.

PR #1015, opened 2017-10-04
Set GitHub user's avatar

### LINKED ISSUE #949: When creating a user from GitHub, set their avatar
# Problem

When the user gets created from the GitHub webhooks, we need to make sure their avatar is set (otherwise we get a broken one).

We should actually do two things:

- Create an async process to pull the image from GitHub and upload it to Cloudinary
- Set a fallback to the GitHub avatar (when present) in the user serializer

joshsmith:
# What's in this PR?

In the async process (after a successful creation of the GitHub user):

- Ensures we have the largest (or original) size of the GitHub avatar in the avatar URL (this is the default we get back)
- Uploads to Cloudinary with the GitHub avatar URL `Cloudex.upload("http://example.org/test.jpg")`
- Takes the `public_id` that's returned and sets it to `cloudinary_public_id` for the `User`
- Adapts `CloudexTest` to allow us to use `Cloudex.upload` in tests
- Tests that the `cloudinary_public_id` is correctly set

Also changes `ImagePresenter` `@logo_options` to `@large_options`.

## References
Fixes #949

### DIFF
--- a/lib/code_corps/accounts/accounts.ex
+++ b/lib/code_corps/accounts/accounts.ex
@@ -5,14 +5,15 @@ defmodule CodeCorps.Accounts do
   All actions to accounts should go through here.
   """
 
+  alias Task.Supervisor, as: TaskSupervisor
   alias CodeCorps.{
     Accounts.Changesets,
     Comment,
     GitHub.Adapters,
     GithubAppInstallation,
+    Task,
     User,
-    Repo,
-    Task
+    Repo
   }
   alias Ecto.{Changeset, Multi}
 
@@ -23,6 +24,16 @@ defmodule CodeCorps.Accounts do
   """
   @spec create_from_github(map) :: {:ok, User.t} | {:error, Changeset.t}
   def create_from_github(%{} = attrs) do
+    with {:ok, user} <- do_create_from_github(attrs) do
+      user |> upload_github_photo_async
+      {:ok, user}
+    else
+      error -> error
+    end
+  end
+
+  @spec do_create_from_github(map) :: {:ok, User.t} | {:error, Changeset.t}
+  defp do_create_from_github(%{} = attrs) do
     %User{}
     |> Changesets.create_from_github_changeset(attrs)
     |> Repo.insert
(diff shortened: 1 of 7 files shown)


joshsmith:
I decided to skip the intermediate fallback step here and put it into polish.

We may also want to pull in the user's image from GitHub if they're connecting from GitHub with an existing account but don't have a `cloudinary_id` yet. The only downside is that they may _also_ not have an actual avatar on GitHub: it may just be an auto-generated one. That still might be better than nothing.

lib/code_corps/accounts/accounts.ex
```diff
@@ -56,6 +65,20 @@ defmodule CodeCorps.Accounts do
     end
   end
 
+  def upload_github_photo_async(%User{} = user) do
+    Task.Supervisor.start_child(:background_processor, fn -> upload_github_photo(user) end)
```
joshsmith: I'm not sure if this needs to be explicitly async since Cloudex itself uses an async process to do the upload. Thoughts here?

lib/code_corps/accounts/accounts.ex
```diff
@@ -79,7 +102,7 @@ defmodule CodeCorps.Accounts do
       |> select([u], u.id)
       |> Repo.all
 
-    Task
+    CodeCorps.Task
```
joshsmith: This is a little annoying, but may not be a problem depending on answer to the above.

test/lib/code_corps/cloudex/uploader_test.exs
```diff
@@ -0,0 +1,12 @@
+defmodule CodeCorps.Cloudex.UploaderTest do
+  alias CodeCorps.Cloudex.Uploader
+  use ExUnit.Case, async: true
+
+  test "returns the public_id" do
+    %Cloudex.UploadedImage{public_id: public_id} =
+      "https://placehold.it/500x500"
+      |> Uploader.upload
+
+    assert public_id
```
joshsmith: Can't think of a better test here.

lib/code_corps/accounts/accounts.ex
begedin: This is one of those things where I lack the experience to be sure. We should look into Cloudex internals.

lib/code_corps/accounts/accounts.ex
begedin: You can always do `alias X as: Y` to get around it. 

I do think we already have other places in code where we had to go by full module name due to this, though, so not a new thing here.

test/lib/code_corps/cloudex/uploader_test.exs
begedin:
```Elixir
assert "https://placehold.it/500x500 |> Upload.upload |> Map.get(:public_id)
```

is an option

begedin (dismissed):
It generally looks good. Consider it approved, but I can't really offer a decent opinion on the async part.

PR #1016, opened 2017-10-04
Add unique constraints for GitHub users

### LINKED ISSUE #981: Duplicate GitHub account connections can happen for users
# Problem

Right now it's possible to connect multiple GitHub accounts to a user. This is due to the fact that an account is connected via a `github_id` field that has no unique index and no validation against duplicates.

We may want to introduce a `GitHubUser` relationship prior to the making this fix, but it should be straightforward to migrate either way.

begedin:
# What's in this PR?

The simplest "solution" to this problem right now is to add an index to `github_id`, add a validation, and add a decent custom error message explaining what went wrong:

> This GitHub account is already connected to another user on Code Corps.

## References

Fixes #981


joshsmith:
@begedin would you be able to update your version of postgres, or should we find some way to indicate a specific version we're looking for (like a log warning)?

begedin:
I can update, and should/will, but if different minor versions of postgres are going to be a problem for volunteers, that could become annoying.

Ideally, we could have something like disable outputing the version into `structure.sql` and instead add some sort of behavior where the version is checked unrelated to the file, and a warning is printed out if major versions mismatch.

joshsmith:
@begedin I can't ask for a review from you but I do need a review on this.

begedin (commented):
Can't really approve since I started the PR but 👍

PR #1022, opened 2017-10-07
When updating user from Github, add avatar if none exists

joshsmith:
# What's in this PR?

Similar to another recent PR but only sets the user's avatar to a GitHub one if none exists, including when someone connects to GitHub as an existing user.


begedin (dismissed):
Looks pretty straightforward to me 👍

PR #1024, opened 2017-10-08
Remove duplicate :metadata key in test

landongrindheim:
This commit removes a duplicate key in `StripePlatformCardTest`, which is the last bit of noise left when running the test suite. All green dots now :)

Before:
image

After:
image


joshsmith (approved):

joshsmith:
Good catch! Been wondering about that.

begedin:
@landongrindheim Thank you for this, was annoying me. 👍

PR #1029, opened 2017-10-09
Added filter to Tasks using Archived field

### LINKED ISSUE #1007: Filter tasks by archived status
# Problem

- Add a filter to return tasks based on whether they are archived
- Default for returning tasks is _not_ archived

vishaldeepak:
# What's in this PR?
Retrieve tasks with respect to `archived`. Add test cases accordingly. 

Fixes #1007


begedin (approved):
Looks exactly as it should. Great work @vishaldeepak

PR #1037, opened 2017-10-09
Use modified_at in Task to ensure the last write wins

### LINKED ISSUE #1035: Use modified_at in Task to ensure the last write wins
# Problem

We need to be able to use the `modified_at` timestamp in `Task` to ensure that new requests coming in to modify a `Task` do not actually modify it unless the current timestamp is after the `modified_at` timestamp.

### LINKED ISSUE #1036: Use modified_at in Comment to ensure the last write wins
# Problem

We need to be able to use the `modified_at` timestamp in `Comment` to ensure that new requests coming in to modify a `Comment` do not actually modify it unless the current timestamp is after the `modified_at` timestamp.

joshsmith:
# What's in this PR?

This simply adds a custom validation to the `Issues.ChangesetBuilder` that checks whether the `modified_at` is after the `modified_at` we're trying to set, and if so, then errors.

## References
Fixes #1035 
Fixes #1036


joshsmith:
One thing I don't like about this is that I had to change e.g. in all the JSON `2015` to `2075` as a far-future enough date for us to have the `modified_at` changeset validation not be an issue. This is pretty hidden behavior and causes the errors to be rather difficult to parse sometimes.

begedin (dismissed):
Not much to say about this one. Clean and straightforward 👍

PR #1053, opened 2017-10-10
Standardize error output for GitHub.Event.Installation

### LINKED ISSUE #1042: Standardize output for CodeCorps.GitHub.Event.Installation
# Problem

Part of #983

All github webhook event handlers should adhere to the same output conventions. 

The `handle` function should `{:ok, any}` where `any` is the relevant output data such as affected records, etc., if the handling process was successful.

Likewise, if the handling process failed, it should result in an `{:erro, atom}` where `atom` is a failure reason.

joshsmith:
# What's in this PR?

Not 100% comfortable with the changes I've made here, so could really use a second set of eyes on this.

## References
Fixes #1042

### DIFF
--- a/lib/code_corps/github/event/installation.ex
+++ b/lib/code_corps/github/event/installation.ex
@@ -1,80 +1,114 @@
 defmodule CodeCorps.GitHub.Event.Installation do
   @moduledoc """
-  In charge of dealing with "Installation" GitHub Webhook events
+  In charge of handling a GitHub Webhook payload for the Installation event type
+  [https://developer.github.com/v3/activity/events/types/#installationevent](https://developer.github.com/v3/activity/events/types/#installationevent)
   """
+
+  @behaviour CodeCorps.GitHub.Event.Handler
 
   alias CodeCorps.{
     GithubAppInstallation,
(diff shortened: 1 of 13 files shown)


lib/code_corps/github/event/installation.ex
```diff
@@ -1,80 +1,103 @@
...
     User
   }
 
-  @typep outcome :: {:ok, GithubAppInstallation.t, Task.t} | {:error, any}
+  alias Ecto.Multi
+
+  @typep outcome :: {:ok, GithubAppInstallation.t} |
+                    {:error, :unexpected_action} |
+                    {:error, :unexpected_payload} |
+                    {:error, any}
```
joshsmith: Looks like I missed some error types here.

begedin (dismissed):
Approved your stuff from my end. As for reviewing and approving my stuff, not sure if you'll be able to do that.

Either way, this is ready for another review and hopefully merge.

Note that I've added a new module - `ResultAggregator`, since I've noticed a lot of code repetition in places where we're doing commits on collections of records. This is a sort of attempt to standardize.

PR #1055, opened 2017-10-11
Switch GithubAppInstallationController away from ja_resource and canary

### LINKED ISSUE #890: Switch GithubAppInstallationController away from ja_resource and canary
Progress on #864 

# Problem

# Procedure

- pick a controller to switch
- for each action

```Elixir
  end
end

  end
end

  end
end

  end
end
```

### `index` action

- render data

### `show` action

- load resource
- render resource

### `create` action

- perform insert action
- render resulting resource

- load resource
- perform update/delete

### Examples in code

## How to rewrite policies

```Elixir
```

becomes

```Elixir
```

### Examples in code

treble37:
# What's in this PR?

Helps support #864

## References
Fixes #890

Progress on: #864

### DIFF
--- a/lib/code_corps/policy/github_app_installation.ex
+++ b/lib/code_corps/policy/github_app_installation.ex
@@ -6,9 +6,6 @@ defmodule CodeCorps.Policy.GithubAppInstallation do
 
   alias CodeCorps.{GithubAppInstallation, User}
 
-  def create?(%User{} = user, %Ecto.Changeset{} = changeset),
-    do: changeset |> get_project |> owned_by?(user)
+  def create?(%User{} = user, params), do: params |> get_project |> owned_by?(user)
 
-  def update?(%User{} = user, %GithubAppInstallation{} = github_app_installation),
-    do: github_app_installation |> get_project |> owned_by?(user)
 end
--- a/lib/code_corps_web/controllers/github_app_installation_controller.ex
+++ b/lib/code_corps_web/controllers/github_app_installation_controller.ex
@@ -1,29 +1,34 @@
 defmodule CodeCorpsWeb.GithubAppInstallationController do
   use CodeCorpsWeb, :controller
-  use JaResource
 
   import CodeCorps.Helpers.Query, only: [id_filter: 2]
 
-  alias CodeCorps.{GithubAppInstallation}
+  alias CodeCorps.{GithubAppInstallation, User}
 
-  @preloads [:project, :user]
+  action_fallback CodeCorpsWeb.FallbackController
+  plug CodeCorpsWeb.Plug.DataToAttributes
+  plug CodeCorpsWeb.Plug.IdsToIntegers
 
-  plug :load_resource, model: GithubAppInstallation, only: [:show], preload: @preloads
-  plug :load_and_authorize_changeset, model: GithubAppInstallation, only: [:create], preload: @preloads
-
-  plug JaResource
-
-  @spec model :: module
-  def model, do: CodeCorps.GithubAppInstallation
-
-  @spec filter(Plug.Conn.t, Ecto.Query.t, String.t, String.t) :: Ecto.Query.t
-  def filter(_conn, query, "id", id_list) do
-    query |> id_filter(id_list)
+  @spec index(Conn.t, map) :: Conn.t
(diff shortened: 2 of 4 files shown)


lib/code_corps_web/controllers/github_app_installation_controller.ex
```diff
@@ -1,29 +1,49 @@
...
-  plug :load_resource, model: GithubAppInstallation, only: [:show], preload: @preloads
-  plug :load_and_authorize_changeset, model: GithubAppInstallation, only: [:create], preload: @preloads
-
-  plug JaResource
+  action_fallback CodeCorpsWeb.FallbackController
+  plug CodeCorpsWeb.Plug.DataToAttributes
+  plug CodeCorpsWeb.Plug.IdsToIntegers
 
   @spec model :: module
   def model, do: CodeCorps.GithubAppInstallation
```
begedin: You can completely remove the `model` function. It's used internally by `ja_resource` only.

lib/code_corps/policy/github_app_installation.ex
```diff
@@ -6,8 +6,8 @@ defmodule CodeCorps.Policy.GithubAppInstallation do
 
   alias CodeCorps.{GithubAppInstallation, User}
 
-  def create?(%User{} = user, %Ecto.Changeset{} = changeset),
-    do: changeset |> get_project |> owned_by?(user)
+  def create?(%User{} = user, params),
+    do: params |> get_project |> owned_by?(user)
 
   def update?(%User{} = user, %GithubAppInstallation{} = github_app_installation),
```
begedin: Seems like you pointed out a bug we weren't aware of. There github app installation endpoint should only support `:index`, `:show` and `:create` in the controller. That means we do not really need an `:update` policy here, so feel free to remove it as well as the associated tests.

lib/code_corps_web/controllers/github_app_installation_controller.ex
```diff
@@ -1,29 +1,49 @@
...
+      conn |> put_status(:created) |> render("show.json-api", data: installation)
+    end
   end
 
-  @spec handle_create(Plug.Conn.t, map) :: Ecto.Changeset.t
-  def handle_create(_conn, attributes) do
-    %GithubAppInstallation{}
-    |> GithubAppInstallation.create_changeset(attributes)
+  @spec delete(Plug.Conn.t, map) :: Conn.t
+  def delete(%Conn{} = conn, %{"id" => id} = params) do
```
begedin: Again, a "bug" here. We should not be supporting a `:delete`. The controller should only have `:index`, `:show` and `:create`.

begedin (dismissed):
Some changes that are actually unrelated to your work, but were pointed out thanks to it. Otherwise, good job!

lib/code_corps_web/controllers/github_app_installation_controller.ex
treble37: thx

lib/code_corps/policy/github_app_installation.ex
treble37: Ok, all changes should be live now, thanks @begedin !

lib/code_corps/policy/github_app_installation.ex
```diff
@@ -6,9 +6,7 @@ defmodule CodeCorps.Policy.GithubAppInstallation do
 
   alias CodeCorps.{GithubAppInstallation, User}
 
-  def create?(%User{} = user, %Ecto.Changeset{} = changeset),
-    do: changeset |> get_project |> owned_by?(user)
+  def create?(%User{} = user, params),
+    do: params |> get_project |> owned_by?(user)
```
joshsmith: Minor, but would kill this line.

joshsmith (dismissed):

joshsmith:
Tiny tiny change and then this is good to go.

lib/code_corps/policy/github_app_installation.ex
treble37: removed newline and it's all one one line now, thx!

begedin (approved):

begedin:
Great work @treble37

PR #1059, opened 2017-10-11
Refactor event handlers to take only the payload

### LINKED ISSUE #983: Handle cases where webhook handler does not fail with an {:error, atom} tuple.
As far as I'm aware, all of our webhook handlers succeed with `{:ok, data}` or `{:error, reason}`, but we should verify this to. The return should always be a two-element tuple, with the first being an `:ok` or `:error`.

The `reason` should also always be an `atom` and that's what our event processing module expects. It takes the reason, converts it using `Atom.to_string/1` and stores it as `failure_reason` into the database.

However, now that I'm thinking about it, there is a possibility that some handlers fail with something else as a reason, for example, a changeset or a github API error payload.

We should take a look at `Event.Issues`, `Event.IssueComment` and `Event.Installation`. If any of the failed transaction cases fail with something other than an `{:error, atom}`, we should add a `case` clause for that specific case and translate it to an `{:error, atom}`. 

For example

```Elixir
case Repo.transaction(multi) do
  {:ok, %{tasks: tasks}} -> {:ok, tasks}
  {:error, :repo, :unmatched_project, _steps} -> {:ok, []}
  {:error, _errored_step, error_response, _steps} -> {:error, error_response}
end
```

becomes

```Elixir
case Repo.transaction(multi) do
  {:ok, %{tasks: tasks}} -> {:ok, tasks}
  {:error, {tasks, failed_changesets}} -> {:error, :task_validation_errors}
  {:error, :repo, :unmatched_project, _steps} -> {:ok, []}
  {:error, _errored_step, error_response, _steps} -> {:error, error_response}
end
```

Note, that the example is at least one identified case where this bug is happening, but there may be more. 

As of right now, if this does happen, the event handling will fail and the event will stay in the "processing" state forever.

joshsmith:
# What's in this PR?

Refactors the event handlers to remove reliance on `GithubEvent` and instead just takes the payload.

## References
Fixes #983

### DIFF
--- a/lib/code_corps/github/event/handler.ex
+++ b/lib/code_corps/github/event/handler.ex
@@ -3,14 +3,12 @@ defmodule CodeCorps.GitHub.Event.Handler do
   Default behavior for all GitHub webhook event handlers.
   """
 
-  alias CodeCorps.GithubEvent
-
   @doc ~S"""
   The only entry point a GitHub webhook event handler function should contain.
 
-  Receives a `CodeCorps.GithubEvent` record and a payload, returns an `:ok`
-  tuple if the process was successful, or an `:error` tuple, where the second
-  element is an atom, if it failed.
+  Receives the GitHub payload, returns an `:ok` tuple if the process was
+  successful, or an `:error` tuple, where the second element is an atom, if it
+  failed.
   """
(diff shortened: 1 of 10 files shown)


joshsmith:
@begedin I decided to just go ahead and merge this since the changes are so small. Let me know if you see anything wrong ex post facto, though.

PR #1067, opened 2017-10-12
Create GithubIssue and modify Task

### LINKED ISSUE #1063: Create GithubIssue model and update relationships
# Problem

- add a migration which creates the model, with data migration included. It could be complicated in regards to seeds, with models and fields being moved around, etc ~1-2 hours
  - remove the github repo relationship
  - remove the github fields
  - move it all to the github issue relationship
- fix seeds (they are likely to get broken) ~1 hour

joshsmith:
# What's in this PR?

WIP

## References
Fixes #1063


joshsmith:
A few things we definitely still need to do:

- Implement uniqueness constraint on the `GithubIssue` `github_id`
- Add `GithubIssueComment` – too much of the syncing is tied up with the relationship between issues and issue comments
- Fix the `CodeCorps.Task.Service` module to actually create a `GithubIssue`

Questions that came up for me:

- What happens when a comment webhook comes in before the issue? Do we create the issue regardless? Wait until we've received a webhook? I'm not sure how this works currently so I can't answer with any meaningful context right now.

Other notes:

- I don't think it's necessary to ask the client to pass in a `GithubIssue`. This is an implementation detail for the backend. However, if the client wants to sync an issue with a given _repository_, then it makes sense to pass just that repository in. Some of the implementation here reflects that difference from the initial writeup.

joshsmith:
I've now rewritten `IssueLinker` to take the issue params instead of the event payload since it can be used with arbitrary GitHub issue JSON params.

`CodeCorps.Task.Service` is rewritten to create the `GithubIssue` upon a successful return from the GitHub API.

We're down to just failing tests on the comments side now.

Still need to:

- Likely move `CodeCorps.GitHub.Event.Issues.IssueLinker` out of the event hierarchy and somewhere that's more representative of its newfound abstraction.
- Generally will need to reconsider the boundaries of some of these modules; their interplay is becoming a bit clearer but they're still tangled unnecessarily.

joshsmith:
@begedin this isn't really ready for a review qua review yet, but I could use some feedback still prior to going deeper on this. Bear in mind I likely won't be able to make progress until the weekend is over due to Army commitments.

lib/code_corps/github/event/issues/changeset_builder.ex
```diff
@@ -51,18 +54,20 @@ defmodule CodeCorps.GitHub.Event.Issues.ChangesetBuilder do
...
     |> Changeset.put_change(:github_repo_id, github_repo_id)
     |> Changeset.put_change(:project_id, project_id)
     |> Changeset.put_change(:task_list_id, task_list_id)
     |> Changeset.put_change(:user_id, user_id)
     |> Changeset.validate_required([:project_id, :task_list_id, :title, :user_id])
+    |> Changeset.assoc_constraint(:github_issue)
     |> Changeset.assoc_constraint(:github_repo)
     |> Changeset.assoc_constraint(:project)
     |> Changeset.assoc_constraint(:task_list)
     |> Changeset.assoc_constraint(:user)
```
begedin, line 67: Considering the amount of associations here, maybe we should go with 

```Elixir
|> Changeset.put_change(:github_issue_id, github_issue_id)
|> Changeset.assoc_constraint(:github_issue)
```

etc.

Might be easier to keep track of these that way and ensure they all have the proper assoc constraint. I'm not strongly for one or the other approach, but wanted to mention the idea.

lib/code_corps/task/service.ex
```diff
@@ -38,27 +39,31 @@ defmodule CodeCorps.Task.Service do
...
-  defp connect_to_github(%Task{github_repo_id: nil} = task), do: {:ok, task}
-  defp connect_to_github(%Task{github_repo_id: _} = task) do
-    with {:ok, issue} <- task |> Repo.preload(@preloads) |> GitHub.Issue.create do
-      task |> link_with_github_changeset(issue) |> Repo.update
+  @spec create_on_github(Task.t) :: {:ok, Task.t} :: {:error, GitHub.api_error_struct}
+  defp create_on_github(%Task{github_repo_id: nil} = task), do: {:ok, task}
+  defp create_on_github(%Task{github_repo: _} = task) do
+    with %Task{github_repo: github_repo} = task <- task |> Repo.preload(@preloads),
+         {:ok, payload} <- GitHub.Issue.create(task),
+         {:ok, %GithubIssue{} = github_issue } <- IssueLinker.create_or_update_issue(github_repo, payload) do
```
begedin, line 49: These could be two distinct transaction steps in the function higher up. Just because it doesn't perform a database action, doesn't mean it can't be a step and there might be less branching paths that way.

Steps could be something like `:create_on_github` and `:create_on_codecorps` and then `marshall_result` could deal with a  github api error and a changeset error instead of the `with` clause.

I think it would make the code easier to read kind of like what we did with all the webhook handlers.

lib/code_corps/task/service.ex
```diff
@@ -38,27 +39,31 @@ defmodule CodeCorps.Task.Service do
...
-  @spec sync_to_github(Task.t) :: {:ok, Task.t} :: {:error, GitHub.api_error_struct}
-  defp sync_to_github(%Task{github_repo_id: nil} = task), do: {:ok, task}
-  defp sync_to_github(%Task{github_repo_id: _} = task) do
-    with {:ok, _issue} <- task |> Repo.preload(@preloads) |> GitHub.Issue.update do
+  @spec update_on_github(Task.t) :: {:ok, Task.t} :: {:error, GitHub.api_error_struct}
+  defp update_on_github(%Task{github_repo_id: nil} = task), do: {:ok, task}
+  defp update_on_github(%Task{github_repo_id: _} = task) do
+    with %Task{github_repo: github_repo} = task <- task |> Repo.preload(@preloads),
+         {:ok, payload} <- GitHub.Issue.update(task),
+         {:ok, %GithubIssue{} } <- IssueLinker.create_or_update_issue(github_repo, payload) do
```
begedin, line 66: Same as above

begedin (commented):
Generally looking good. I just have a couple of suggestions to consider.

begedin:
@JoshSmith I did what I think needs to be done here.

Note that there is a separate commit that deals specifically with eliminating the github repo argument from the task syncer module. The reasoning is, it's contained within the github issue. This decouples the process slightly and helps paving the way into further simplifying and decoupling it in the future.

A downside is that it requires performing a preload within the module, which is something we tend not to do, but in this specific case, the preload was already happening atypically in the `RepoFinder` module so I feel it still makes sense to do it this way.

Some ideas on what we could do later.

- if we have a `GithubUser`, we would not need a `UserLinker` anymore. A `GithubUser` would be associated to a `User` and a `GithubIssue`, simplifying the whole process here. 

- This would allow us to also simplify `RepoFinder` and eliminate checking the project github repo count there, since we no longer have to match user for each task - the check that there are no duplicate users occurs beforehand.

PR #1068, opened 2017-10-14
Remove ja resource and canary

### LINKED ISSUE #864: Remove ja_resource

### LINKED ISSUE #889: Switch DonationGoalController away from ja_resource and canary
Progress on #864 

# Problem

# Procedure

- pick a controller to switch
- for each action

```Elixir
  end
end

  end
end

  end
end

  end
end
```

### `index` action

- render data

### `show` action

- load resource
- render resource

### `create` action

- perform insert action
- render resulting resource

- load resource
- perform update/delete

### Examples in code

## How to rewrite policies

```Elixir
```

becomes

```Elixir
```

### Examples in code

### LINKED ISSUE #1056: Alphabetize function clauses in policy.ex
# Problem

It's difficult to keep track of the function clauses of policy.ex as individual controller pull requests are being submitted.

It would be a far more effective use of time to simply alphabetize everything in one fell swoop once all individual pull requests are done.

This can be considered blocked by #889, #890, #891, #895, #897 and can be tackled once those are done, NOT sooner.

landongrindheim:
This PR addresses a few, distinct issues.  I'm open to and would like feedback on any of the sections :)

First, JaResource/Canary are removed from the DonationGoalController. Note that there is an open pull request for this already. I don't want to make someone's work redundant, so I understand if this is rejected. However, competing this step was necessary for the other two parts of this pull request.

Second, JaResource and Canary are removed as dependencies to CodeCorpsApi.

Third, the authorization function clauses are alphabetized (by resource being authorized). They can be sorted by action being authorized if preferred. 

Since this PR addresses multiple issues, it seems best to me to not squash into a single commit. Doing so would obscure git history. However, if it's preferred that these commits be submitted individually as their own pull requests, I can certainly do that.

## References
Closes #864 
Closes #889  
Closes #1056


joshsmith:
Just a heads up that you do need to write `Closes #[number] Closes #[number] Closes #[number]` (instead of the CSV) for the GitHub keywords to work correctly. They're doing just regex parsing on it.

begedin (approved):
Amazing work here. I completely agree on merging this as 3 separate commits as part of the  same pull request, so this is going in as is.

PR #1075, opened 2017-10-17
Add info to GitHub issue when codecorps user created via bot

### LINKED ISSUE #928: Add content into issue/comment body during github sync, explaining which unconnected codecorps user created it on codecorps
# Problem

From #874, seeing that a task created by a user not connected to github is set to originate from our installation bot, any user viewing it on github would have no information on who actually created the task/comment on codecorps.

I believe we should consider inserting a bit of content into the issue body.

In the case of the source being a bot, we should get the username of the user on codecorps and insert it at the top (or bottom) of an issue. 

Something like `created by CodeCorps User %{username}`. It could even be a link to the user's profile, or to the task itself.

Additionally, when handling a webhook for such an issue coming from github, we should make sure to remove this content as we propagate back into CC.

I'm not sure how strongly I feel about this, but I believe it's something to consider.

begedin:
# What's in this PR?

Closes #928 

This probably needs some architectural work. My goal was to spike out the basic behavior in the limited time I had.

## Notes

Since `Adapters.Task` is used from the `GithubIssue` endpoint to create an issue from a task, I decided to put the behavior there.

A new requirement for the adapter now is that the task needs the project and the organization preloaded in order to build the url. However, there's already preloading involved, so this does not change much. 

Another thing of note is that I'm building the client URL here, using the `:code_corps, :site_url` config value, so it forces the backend to know quite a bit about the frontend. Should we consider moving this logic all the way to the client? That would make this PR useless, though.


begedin:
This now includes related changes for comments.

I've also extracted the url computation into a separate module. Ideally, it should not exist, but at least it's all in one place this way.

The adapters could use some drying up to, but I'm not sure how to structure it.

lib/code_corps/github/adapters/comment.ex
```diff
@@ -40,11 +39,53 @@ defmodule CodeCorps.GitHub.Adapters.Comment do
...
+  update an GitHub Issue Comment through the GitHub API.
+  """
+  @spec to_api(Comment.t) :: map
+  def to_api(%Comment{} = comment) do
+    comment
+    |> Map.from_struct
+    |> MapTransformer.transform_inverse(@mapping)
+    |> Map.drop(@autogenerated_github_keys)
+    |> add_user_info_header(comment)
+  end
```
begedin: I'm sort of rethinking our approach with adapters here. We have a separate `issue` and `task` adapter, for example.

With the approach here, we would instead have a single `issue` adapter, which has a `to_api` accepting a `Task` or a `GithubIssue` and mapping based on type, as well as a `to_task` and `to_github_issue` both accepting a map.

What do you think, @JoshSmith ?

lib/code_corps/github/adapters/comment.ex
```diff
@@ -40,11 +39,53 @@ defmodule CodeCorps.GitHub.Adapters.Comment do
...
+
+  @spec remove_user_info_header(map) :: map
+  defp remove_user_info_header(%{"body" => body} = attrs) when is_binary(body) do
+    clean_body = case body |> String.split(@separator) do
+      ["Posted by" <> _rest | tail] -> tail |> Enum.join(@separator)
+      _ -> body
+    end
+    attrs |> Map.put("body", clean_body)
+  end
+  defp remove_user_info_header(%{} = attrs), do: attrs
```
begedin: All of this here is almost identical to the task stuff, but I'm not sure where to move it or how to name it.

lib/code_corps/github/adapters/task.ex
```diff
@@ -39,5 +43,33 @@ defmodule CodeCorps.GitHub.Adapters.Task do
     |> Map.from_struct
     |> MapTransformer.transform_inverse(@mapping)
     |> Map.drop(@autogenerated_github_keys)
+    |> add_user_info_header(task)
   end
+
+  @separator "\r\n\r\n---\r\n\r\n"
```
begedin: This could probably be configurable.

There is also the problem that it isn't necessarily unique. Issues and comments could have this separator anyway, so we should have additional logic to ensure the first section (before the first separator) is actually our header section.

We can't compute the entire section and check for it, since we just have  payload, not the actual comment/task record.

lib/code_corps/task/service.ex
```diff
@@ -39,7 +39,15 @@ defmodule CodeCorps.Task.Service do
...
-  @preloads [:github_issue, [github_repo: :github_app_installation], :user]
+  # :user, :github_issue and :github_repo are required for connecting to github
+  # :project and :organization are required in order to add a header to the
+  # github issue body when the user themselves are not connected to github, but
+  # the task is
+  #
+  # Right now, all of these preloads are loaded at once. If there are
+  # performance issues, we can split them up according the the information
+  # provided here.
+  @preloads [:github_issue, [github_repo: :github_app_installation], :user, [project: :organization]]
```
begedin, line 50: Decided to add info here, because we're bound to forget.

lib/code_corps/github/adapters/comment.ex
joshsmith: This makes perfect sense to me, and I would agree.

One thing to also consider is that the payload for an issue from an event is different from the payload for an issue from the endpoint itself. This might get us into trouble in the future.

lib/code_corps/ember_client.ex
```diff
@@ -0,0 +1,32 @@
+defmodule CodeCorps.EmberClient do
+  @moduledoc ~S"""
+  Confirms URLs for the Ember client app routes
+  """
```
joshsmith: I wouldn't reference Ember specifically here, but instead just call this something like `WebClient`.

lib/code_corps/github/adapters/comment.ex
joshsmith: Maybe `Adapters.Utils.BodyDecorator` (or something?) and `add_code_corps_header` / `remove_code_corps_header` which accepts both `Comment` and `Task`.

lib/code_corps/github/adapters/task.ex
joshsmith: We should definitely keep that visible separator but we could also hard end with a Markdown comment:

```
\r\n\r\n[//]: # (Please type your edits below this line)\r\n---\r\n\r\n
```

joshsmith (changes requested):

joshsmith (approved):

PR #1079, opened 2017-10-17
Set order on tasks from issues/PRs coming from GitHub

### LINKED ISSUE #1078: Tasks created from GitHub do not have their order set initially
# Problem

Tasks created from GitHub do not have their order set initially. This is likely due to the fact that the changesets in `Task` itself apply position, but we have no such changes applied inside the changeset builder for the GitHub event processing.

joshsmith:
# What's in this PR?

This fixes the task ordering when an issue/PR comes from GitHub.

## References
Fixes #1078


begedin (approved):
I'm not seeing any problems with this. 

For the purpose of documenting what this does....

Basically, when we create or update a task, in all cases, we now ensure that the task is positioned at the top of it's respective task list, provided it hasn't been positioned already.

PR #1080, opened 2017-10-17
Add missing indexes for github_id to GithubComment, GithubIssue, GithubRepo

joshsmith:
# What's in this PR?

This adds missing unique indexes for `github_id` to `GithubComment`, `GithubIssue`, and `GithubRepo`.

## References
Fixes #1076

PR #1081, opened 2017-10-18
Ensure GitHubRepo is always find or create

### LINKED ISSUE #1077: Github-x model creation should be find or create by
# Problem

All our `Github-x` model creation should be done as a find or create by using the `github_id`. Right now, the assumption is that this is the first insert which is incorrect.

joshsmith:
# What's in this PR?

WIP.

## References
Progress on #1077


joshsmith:
Uses #1080 as a base.

joshsmith:
This still needs a test to ensure it's working as intended.

PR #1086, opened 2017-10-18
Code and comment cleanups for IssueComment

begedin:
# What's in this PR?

This was done as part of looking into #1077. 

While searching for spots where the desired behavior (comment is always created_or_updated), I found some comment/code mistakes (not really bugs) which I corrected


begedin:
Merging this, since it does not change the code in any way

Issue #1038, opened 2017-10-10
Add analytics tracking for core actions


joshsmith:
# Problem

Here are the remaining core actions we need to track:

- [ ] Reordered Task in List
- [ ] Connected GitHub App Installation to Project

Some we can't do yet but should be able to soon:

- [ ] Sent Suggested Tasks Email
- [ ] Opened Suggested Tasks Email (using email provider's API?)
- [ ] Clicked Suggested Task in Email (using email provider's API?)
- [ ] Signed Up with GitHub

joshsmith:
From @begedin in #1105:

> I've spent time trying to implement tracking for the case of a position change. However, due to our use of `EctoOrder`, I was unable to find a way to reliably determine if the change took place.
> 
> - `:position` field is virtual and is nil upon initially loading the `Task`
> - `:order` field is set internally by `EctoOrdered`. A "rebalancing" step takes place during insert or update, which effectively can change its value even if no actual change took place
> - tracking params received from the client also isn't an option, since they could contain a value for position which doesn't actually make a change to it
> 
> For now, best I could do is add the order field to the task traits in tracking and work with "Edited Task". If you have a better idea, I'm happy to implement it.

> On the note of GithubAppInstallation tracking
> 
> - this PR tracks `CodeCorpsWeb` controller actions only, meaning
>   - when user creates an installation from the client app (not actually on github yet, necessarily)
>     - this installation is immediately connected to a project, so I don't think separate tracking of that makes sense in the context
>   - when user clicks connect on the repo, creating a `ProjectGithubRepo`
> 
> - what else do we need to track
>   - when an installation is created from github - it will either just be saved, unconnected to a project, or it will be matched to an existing installation created above and then connected to a project - 2 events
>   - when an installation is removed from github
> 
> The problem with this new tracking, however, is that there's no `current_user` within that context, so we either need to track without it, or figure out a way to provide it. In most cases, for example, we could use the `sender` object within the github payload to try and match to a `CodeCorps` user. However, there are still cases where this does not exist, so in that case we either provide a bot user as context, or we provide no context at all.

zacck-zz:
@joshsmith  @begedin  I would like to try out implementing the remaining tracking?

Issue #927, opened 2017-09-13
Make segment tracking more explicit


begedin:
# Problem

We're moving away from using `ja_resource` and `canary` through #864 

This makes controllers and authorization policies much more explicit and easier to follow.

However, segment tracking still relies on a plug which does it's tracking through inference of actions. We should make this explicit as well, by moving the code into controllers.

Additionally, all we should ensure all API actions which are supposed to call segment tracking actually do so.

begedin:
Blocked, partially, by #864

begedin:
In addition to the above, I'm thinking we should write assertion helpers which are more explicit, such as `assert_segment_tracked(current_user, action, resource)`

begedin:
I'm moving this out of the milestone. The changes in the milestone should serve as an example of what we want here, but this specific task is not directly related.

I'm also tagging it as "needs clarification".

I'll tackle a controller when I get the chance, so it becomes clearer what we have to do. Then I'll do a write up and create individual issues.

joshsmith:
@begedin what is the status of this?

begedin:
@joshsmith Still in progress, I'm afraid. We've veered off of the implicit tracking in newer controller actions, but several older ones still use implicit tracking via the plug:

```Elixir
# lib/code_corps/analytics/segment_tracking_support.ex

def includes?(:create, %CodeCorps.Comment{}), do: true
def includes?(:update, %CodeCorps.Comment{}), do: true
def includes?(:create, %CodeCorps.DonationGoal{}), do: true
def includes?(:update, %CodeCorps.DonationGoal{}), do: true
def includes?(:create, %CodeCorps.ProjectUser{}), do: true
def includes?(:update, %CodeCorps.ProjectUser{}), do: true
def includes?(:create, %CodeCorps.StripeConnectAccount{}), do: true
def includes?(:create, %CodeCorps.StripeConnectCharge{}), do: true
def includes?(:create, %CodeCorps.StripeConnectPlan{}), do: true
def includes?(:create, %CodeCorps.StripeConnectSubscription{}), do: true
def includes?(:create, %CodeCorps.StripePlatformCard{}), do: true
def includes?(:create, %CodeCorps.StripePlatformCustomer{}), do: true
def includes?(:create, %CodeCorps.User{}), do: true
def includes?(:update, %CodeCorps.User{}), do: true
def includes?(:create, %CodeCorps.UserCategory{}), do: true
[...]

PR #1089, opened 2017-10-18
Added tracking of "Connected to GitHub" event to segment

begedin:
# What's in this PR?

Adds tracking of "Connected to GitHub" event with Segment.

Considering the existence of #927, I did not rely on our tracker "plug" to implement tracking here and instead made it explicit by calling the `track/3` function from the controller.

Progress on: #1038


joshsmith (approved):
Looks good. Also added some extra traits we weren't tracking.

PR #1093, opened 2017-10-19
Move syncers and changesets out of event

joshsmith:
# What's in this PR?

Moves syncers, linkers, and changesets out of event.


begedin (approved):
Moves seem fine to me. Not much else to say.

PR #1094, opened 2017-10-19
Rewrite our GitHub API event syncing infrastructure

### LINKED ISSUE #1082: Fix IssueComment handler to handle pull request comments
# Problem

Currently the `Event.IssueComment` handler has an `operational_multi` that matches the following payload:

```
%{"action" => action, "issue" => _, "comment" => _}
```

Unfortunately, this makes the assumption that every `IssueComment` is a comment on a `GithubIssue` and _not_ potentially a comment on a `GithubPullRequest`. This can lead to us creating a `GithubIssue` even though a `GithubPullRequest` already exists.

In the latter case, the payload has an additional `"pull_request":` key in the JSON payload, nested inside the `"issue":` key. We could potentially use this to match against the pull request.

An additional idea: we could also try, as an additional safety measure, to `find_or_create` by looking for _both_ `GithubIssue` and `GithubPullRequest` records by the `github_id` of the issue. This could help us avoid a bad situation because it would effectively enforce uniqueness of `github_id` across both `GithubPullRequest` and `GithubIssue` records.

### LINKED ISSUE #1097: Change the way issues and pull requests sync from GitHub
# Problem

Right now, both issues and pull requests that we receive from the GitHub API sync directly to `CodeCorps.Task` records. We need to modify this so that:

- A pull request only syncs to `GithubPullRequest`
- An issue syncs to `GithubIssue`, `Task`, and kicks off a sync to `GithubPullRequest`

When the issue is syncing to `GithubPullRequest`, we'll need to use the `pull_request` `url` in the issue payload body to fetch the pull request data from the GitHub API. This data can then be used in the `GithubPullRequest`.

This means that `Sync.PullRequest.Task` will go away entirely.

joshsmith:
# What's in this PR?

Rewrites our GitHub API event syncing infrastructure:

- Change GithubIssue and GithubPullRequest relationships
- Add GitHub.Sync module and submodules for handling syncing
- Remove pull request task syncing since tasks since to issues instead
- Remove linking user and pull requests; rely on issue instead
- Add some docs about the GitHub testing helpers
- Add docs and proper error types
- Fix syncing of Github issue comment events for a pull request

Fixes #1082 
Fixes #1097


joshsmith:
Getting closer to passing tests here. Need to stop for the night. Massive changes to tests since everything has changed so significantly.

The biggest changes are essentially that the `GitHub.Sync` expects all the other syncers to simply return `Multi.t` so that we can `Multi.merge` them together.

joshsmith:
This could still use a test for `Github.Sync` directly. There are no guarantees the `_api` functions work as of right now, but I'd be happy to loop back and test that _after_ this is merged.

.github/CONTRIBUTING.md
```diff
@@ -193,3 +193,5 @@ We've written some convenience helper modules and functions to help with API tes
 - `CodeCorpsWeb.ApiCase` - used to simplify testing JSON API endpoints
 - `CodeCorps.AuthenticationTestHelpers` - provides authentication helpers for authenticating a `conn`
 - `CodeCorps.Factories` - provides factories using [`ex_machina`](https://github.com/thoughtbot/ex_machina), which makes it easy to create test data and associations with Ecto
+- `CodeCorps.GitHub.TestHelpers` - provides test helpers for loading GitHub API endpoint or event JSON fixtures, as well as mocking the API
+- `CodeCorps.GitHub.SuccessAPI` - provides some automatic mocking of the GitHub API, or raises an exception if you have an unhandled request to the API
```
joshsmith, line 197: Although this is documented here, I am concerned that this is not more explicit right now, as it tripped me up how I'm supposed to do the API mocking and took some digging.

lib/code_corps/github/sync/sync.ex
```diff
@@ -0,0 +1,154 @@
...
+
+  @type outcome :: {:ok, list(Comment.t)}
+                 | {:ok, list(Task.t)}
+                 | {:error, :repo_not_found}
+                 | {:error, :validating_issue}
+                 | {:error, :validating_user}
+                 | {:error, :multiple_github_users_match}
+                 | {:error, :validating_tasks}
+                 | {:error, :validating_comments}
+                 | {:error, :unexpected_transaction_outcome}
```
joshsmith: It might make sense to write these outcomes as individual outcomes for each event instead, but have some typespecs that are reusable across them.

lib/code_corps/github/sync/sync.ex
```diff
@@ -0,0 +1,154 @@
...
+  defp marshall_result({:error, :repo, :unmatched_repository, _steps}), do: {:error, :repo_not_found}
+  defp marshall_result({:error, :github_issue, %Ecto.Changeset{}, _steps}), do: {:error, :validating_github_issue}
+  defp marshall_result({:error, :github_comment, %Ecto.Changeset{}, _steps}), do: {:error, :validating_github_comment}
+  defp marshall_result({:error, :comment_user, %Ecto.Changeset{}, _steps}), do: {:error, :validating_user}
+  defp marshall_result({:error, :comment_user, :multiple_users, _steps}), do: {:error, :multiple_comment_users_match}
+  defp marshall_result({:error, :issue_user, %Ecto.Changeset{}, _steps}), do: {:error, :validating_user}
+  defp marshall_result({:error, :issue_user, :multiple_users, _steps}), do: {:error, :multiple_issue_users_match}
+  defp marshall_result({:error, :comments, {_comments, _errors}, _steps}), do: {:error, :validating_comments}
+  defp marshall_result({:error, :tasks, {_tasks, _errors}, _steps}), do: {:error, :validating_tasks}
+  defp marshall_result({:error, _errored_step, _error_response, _steps}), do: {:error, :unexpected_transaction_outcome}
```
joshsmith, line 152: I'm not quite confident this is an exhaustive list.

lib/code_corps/github/sync/user/record_linker.ex
```diff
@@ -39,18 +38,17 @@ defmodule CodeCorps.GitHub.Sync.User.RecordLinker do
     - If there are multiple matching users, this is an unexpected scenario and
       should error out.
   """
-  @spec link_to(GithubComment.t | GithubIssue.t | GithubPullRequest.t, map) :: linking_result
-  def link_to(%GithubComment{} = comment, %{"comment" => %{"user" => user}}), do: do_link_to(comment, user)
-  def link_to(%GithubIssue{} = issue, %{"issue" => %{"user" => user}}), do: do_link_to(issue, user)
-  def link_to(%GithubPullRequest{} = pull_request, %{"pull_request" => %{"user" => user}}), do: do_link_to(pull_request, user)
+  @spec link_to(GithubComment.t | GithubIssue.t, map) :: linking_result
+  def link_to(%GithubComment{} = comment, %{"user" => user}), do: do_link_to(comment, user)
+  def link_to(%GithubIssue{} = issue, %{"user" => user}), do: do_link_to(issue, user)
```
joshsmith, line 43: We no longer match the `GithubPullRequest` because the linking to users happens with issues (and tasks). Pull requests can be separate from this affair since the issue users are the same as the task users.

lib/code_corps/model/github_issue.ex
```diff
@@ -34,12 +35,15 @@ defmodule CodeCorps.GithubIssue do
...
+    |> cast(params, [:github_pull_request_id, :github_repo_id])
+    |> assoc_constraint(:github_pull_request)
     |> assoc_constraint(:github_repo)
   end
 
   def update_changeset(struct, params) do
     struct
     |> changeset(params)
+    |> cast(params, [:github_pull_request_id])
+    |> assoc_constraint(:github_pull_request)
```
joshsmith, line 47: I update the pull request ID here, too, since it's conceivable that we could receive an issue that does not have the pull request synced, but I am not positive this is true. It may be worth writing tests for this and removing its reliance, or keeping it there as a safety.

joshsmith (commented):
Some comments below.

A general comment that there are a lot of tests that have been removed here that were testing for edge cases that, in many cases, I felt were no longer necessary to handle. It's possible that I was overeager, though, so I could use some fairly thorough doublechecking on that.

joshsmith:
This is still missing a test for an issue comment event for a pull request. That I think is the primary reason the coverage has decreased.

.github/CONTRIBUTING.md
begedin: Other than setting up the testing framework in each test module, I can't think of a way to make it more explicit.

We could add more info to the error message, maybe?

lib/code_corps/github/sync/comment/comment.ex
```diff
@@ -48,52 +28,31 @@ defmodule CodeCorps.GitHub.Sync.Comment do
...
   """
-  @spec sync(map) :: outcome
-  def sync(payload) do
-    payload
-    |> operational_multi()
-    |> Repo.transaction
-    |> marshall_result()
+  @spec sync(map, map) :: Multi.t
+  def sync(changes, payload) do
+    sync_multi(changes, payload)
```
begedin: Seems like there's just an extra nested function here we don't really need right now. Did you plan to add another case/clause here, or are you still planning?

lib/code_corps/github/sync/comment/comment/changeset.ex
```diff
@@ -22,22 +22,22 @@ defmodule CodeCorps.GitHub.Sync.Comment.Comment.Changeset do
   @spec build_changeset(Comment.t, map, GithubComment.t, Task.t, User.t) :: Changeset.t
   def build_changeset(
     %Comment{id: nil} = comment,
-    %{"comment" => attrs},
+    attrs,
```
begedin: `%{} = attrs` would provide some type checking here. Not strictly needed, but you mentioned before you prefer to have it.

lib/code_corps/github/sync/sync.ex
begedin: Probably easier to have it all in one place.

You could also add a `@typedoc` to explain which come from where if you want

lib/code_corps/github/sync/sync.ex
```diff
@@ -0,0 +1,100 @@
...
+                 | {:ok, list(Task.t)}
+                 | {:error, :repo_not_found}
+                 | {:error, :validating_issue}
+                 | {:error, :validating_user}
+                 | {:error, :multiple_github_users_match}
+                 | {:error, :validating_tasks}
+                 | {:error, :validating_comments}
+                 | {:error, :unexpected_transaction_outcome}
+
+  def issue_event(%{"issue" => issue} = payload) do
```
begedin: You should probably add `@doc` for all of these.

begedin (changes requested):
I have a couple of comments here and there, but generally, this is looking good.

begedin (dismissed):

begedin (dismissed):

begedin (approved):

PR #1098, opened 2017-10-21
Add tests for MapTransformer

### LINKED ISSUE #601: Write tests for CodeCorps.Adapter.MapTransformer
# Problem

`transform/2` and `transform_inverse/2` need at least some basic testing done.

Just basic tests for the clear case described by docs should do for now. If we encounter bugs to the code along the way, we can then add regression tests and fix as it happens.

tteerawat:
# What's in this PR?

Add basic tests for `CodeCorps.Adapter.MapTransformer` module.

## References
Fixes #601


joshsmith:
Excellent! This looks great. Thanks so much for doing this.

joshsmith (approved):

PR #1103, opened 2017-10-22
Update exdoc to add groups for modules and improve some docs

joshsmith:
# What's in this PR?

Update exdoc to add groups for modules and improve some docs


joshsmith:
Since these are just changes to docs, I'm merging to save some time.

PR #1104, opened 2017-10-22
Refactor comment deletion and ensure all comment records are deleted

### LINKED ISSUE #1100: Comment deletion should delete both Comment and GithubComment records
# Problem

Right now the comment deleter for the GitHub API events only deletes `Comment` records and not the accompanying `GithubComment` records. This should be rewritten to delete both, since a comment is no longer accessible from Github's API after deletion, which makes it not useful on our side.

joshsmith:
# What's in this PR?

- Moves the comment deleter logic into the separate appropriate syncing responsibilities
- Ensures both `Comment` and `GithubComment` records are deleted when a comment is deleted on GitHub

## References
Fixes #1100


begedin (approved):

begedin:
@JoshSmith I'm expecting to get conflicts in #1105 if I don't (at the very least in tests), so I'll merge this one myself.

PR #1113, opened 2017-10-24
GitHub pagination

### LINKED ISSUE #1107: Add support for GitHub requests which involve pagination
# Problem

Our api needs support for pagination in order to deal with initial syncing of github requests

http://justincalleja.com/2016/01/26/elixir-streams-reload/#How_to_fetch_Github_organization_repos_and_their_pagination has a pretty nice writeup on how we could potentially implement that.

However, while streams can be done in paralel, the write up in that link does it sequentially, since it relies on `links.next_page` to fetch each page of results.

GitHub does not recommended this, but we definitely could
- get the first page to figure out how many pages there are
- create a parallel request for each page

- entrypoint should be `CodeCorps.GitHub`
- since pagination is on GET requests, there should not be a need to support specifying the method as an argument - `get_all` or something like that would work
- it's used for syncing, so it should be an integration request

Once we have that, we could continue by adding higher level functions to our individual endpoints, **in separate issues**.

- `GitHub.API.Installation.repositories` may have to be enhanced to support this
- `GitHub.API.Repository.issues` needs to be implemented
- `GitHub.API.Issue.comments` needs to be implemented

### LINKED ISSUE #1109: Implement GitHub.API.Repository module
# Problem

#1107 will allow us to implement a  `GitHub.API.Repositories` module with, initially, an `issues(GithubRepo.t)` function which fetches all issues for a github repository.

Considering how we want to deal with our syncing, we may want the function return a stream instead of a loaded list.

This is blocked until we actually implement #1107

begedin:
- Adds GitHub request pagination
- Adds repository issue syncing
- Adds acceptance testing of the GitHub API

Closes #1109


begedin:
Note that if we want to figure out the eagier approach, we could still do it in a semi-streamable fashion. 

For example, we can get a stream of requests, then evaluate it using `Stream.chunk_every`

test/support/github/test_helpers.ex
```diff
@@ -33,4 +33,16 @@ defmodule CodeCorps.GitHub.TestHelpers do
...
+  defmacro with_real_api(do: block) do
+    quote do
+      old_mock = Application.get_env(:code_corps, :github)
+      Application.put_env(:code_corps, :github, CodeCorps.GitHub.API)
+
+      unquote(block)
+
+      Application.put_env(:code_corps, :github, old_mock)
+    end
+  end
```
joshsmith, line 104: This uses the real API to test. I feel fairly confident that we should introduce _some_ acceptance tests rather than letting our production server serve as that fallback. They should be kept to a bare minimum (like in this case testing a sync of a repository should only be done maybe when merging from staging to master).

I'm not comfortable right now, though, with the best way to do this with `tags`. Could use thoughts.

test/lib/code_corps/github/api/repository_test.exs
```diff
@@ -0,0 +1,47 @@
...
+
+    test "calls github API with the real API" do
+      owner = "baxterthehacker"
+      repo = "public-repo"
+      github_app_installation = insert(:github_app_installation, github_account_login: owner)
+      github_repo = insert(:github_repo, github_app_installation: github_app_installation, name: repo)
+
+      {:ok, issues} = Repository.issues(github_repo)
+
+      assert Enum.count(issues) == 8
```
joshsmith: This doesn't actually use the `with_real_api` right now. It should.

lib/code_corps/validators/time_validator.ex
```diff
@@ -11,7 +11,10 @@ defmodule CodeCorps.Validators.TimeValidator do
   def validate_time_after(%{data: data} = changeset, field) do
     previous_time = Map.get(data, field)
     current_time = Changeset.get_change(changeset, field)
-    case current_time |> Timex.after?(previous_time) do
+    is_after = current_time |> Timex.after?(previous_time)
+    is_equal = current_time |> Timex.equal?(previous_time)
+    after_or_equal = is_after || is_equal
+    case after_or_equal do
```
joshsmith, line 17: Nothing is testing for this change yet.

lib/code_corps/github/api/eager_api.ex
```diff
@@ -0,0 +1,83 @@
+defmodule CodeCorps.GitHub.EagerAPI do
+  @moduledoc """
+  Eager loads a resource from the GitHub API by fetching all of its pages in
+  parallel.
+  """
+
+  def get_all(url, headers, options) do
+    HTTPoison.start
+    {:ok, response} = HTTPoison.get(url, headers, options)
```
joshsmith, line 9: This should likely be modified to call into a more central place so we can mock the `%HTTPoison.Response{}` (headers included). Otherwise it will be hard to test pagination.

joshsmith:
Coverage of `GitHub.API` has gone down due to the `defdelegate get_all`

lib/code_corps/github/api/eager_api.ex
begedin: I'm thinking we have an `API.Gateway` which simply has a request method that calls `HTTPosion`. Response parsing can be done in this module. We then mock the gateway in tests.

lib/code_corps/github/api/eager_api.ex
```diff
@@ -0,0 +1,83 @@
...
+
+    first_page = Poison.decode!(response.body)
+    case response.headers |> retrieve_total_pages do
+      1 -> first_page
+      total ->
+        first_page ++ get_remaining_pages(total, url, headers, options) |> List.flatten
+    end
```
begedin, line 16: With having a separate gateway, our `marshall_response` could then

- check if the headers contains a `"Links"` key 
- if it does (and optionally, doable in a separate PR, if an option is set to get all), it calls this code which makes additional requests for further pages

This isn't absolutely needed in this PR, but could be an approach

lib/code_corps/validators/time_validator.ex
begedin: I'm OK with creating an issue and doing it separately.

test/support/github/test_helpers.ex
begedin: Not sure what the perfect approach is, but  we could make a command alias, for example `mix test.acceptance`, which basically calls `mix test --include :acceptance`

Then we setup  `mix test` to auto-exclude tests with the tag `:acceptance`. 

If we want to use the same test in acceptance and mock API mode, that's probably also doable, but we'll have to be creative. I'm not sure what the approach would be in that case.

begedin (commented):
I have some comments here, hope that helps.

joshsmith (approved):

PR #1116, opened 2017-10-24
Fixes create policies and refactors tests to be more explicit

joshsmith:
# What's in this PR?

Some policies were broken to not take params, and our tests were not catching this.

This fixes the tests and the policies.


joshsmith:
I'm going to merge this as it's a bug in production and I have confidence in the changes.

PR #1118, opened 2017-10-25
Change timestamps fields in model to utc timezone #988

### LINKED ISSUE #988: Replace all timestamps with utc_datetime
- [ ] `@timestamps_opts [type: :utc_datetime] in the model file.`

This is a great article explaining the differences.

https://www.amberbit.com/blog/2017/8/3/time-zones-in-postgresql-elixir-and-phoenix/

paulorsouza:
# What's in this PR?

List the changes you made and your reasons for them.

## References
Fixes #988

### DIFF
--- a/lib/code_corps/model/auth_token.ex
+++ b/lib/code_corps/model/auth_token.ex
@@ -17,7 +17,7 @@ defmodule CodeCorps.AuthToken do
 
     belongs_to :user, CodeCorps.User
 
-    timestamps()
+    timestamps(type: :utc_datetime)
   end
 
   @doc """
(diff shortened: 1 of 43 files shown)


snewcomer (commented):
Nice work!

Not sure if this comment applies, but the only other thing I saw from this article was updating the migrations to include `timestampz` for when you connect to the application via psql or something.  Not something critical, but just thought I would bring it up.

https://www.amberbit.com/blog/2017/8/3/time-zones-in-postgresql-elixir-and-phoenix/

begedin (commented):
I can't find any information on how this would deal with existing timestamps.

The way I understand it, the default type for migrations is `:naive_datetime`.

The changes here make it so the existing timestamps are interpreted as having the `utc` timezone even if the database field itself has no timezone information? 

If that is the case, this looks good to me.

If it is not, we may need a migration to change the existing fields.

snewcomer:
This PR could, sadly, be closed.  Sry since this was prob quite a bit of work on your end :(.  See bottom of #988 for how it is already configured.  @paulorsouza wish we could have got this PR in for ya.  Lmk if there is another PR you want to jump on.

PR #1119, opened 2017-10-25
Fix identifiers in views to not preload for belongs_to IDs

joshsmith:
# What's in this PR?

Fixes identifiers in views to not preload when all we need is the relationship ID displayed.


begedin (dismissed):
Looks good. Love the performance boost. Should be good to go once tests are fixed.

joshsmith:
I love it, too, especially later in the summer.

PR #1122, opened 2017-10-26
Fix performance issues

joshsmith:
# What's in this PR?

Some fixes for API performance issues.

- Remove preloads from views and write explicitly in controllers
- Reduce id filter repetition
- Ensure id query filters are run on every controller that has an index action
- Add missing indexes


test/lib/code_corps_web/views/project_view_test.exs
```diff
@@ -94,14 +94,6 @@ defmodule CodeCorpsWeb.ProjectViewTest do
...
             ]
-          },
-          "tasks" => %{
-            "data" => [
-              %{
-                "id" => task.id |> Integer.to_string,
-                "type" => "task"
-              }
-            ]
           }
```
begedin: Did you check how this works with ember. Do we access `project.tasks` on the client end anywhere? I seem to remember this missing causing problems for us, but the problem may have gone away due to introducing task lists.

begedin (commented):
This is a great change overall. 

I especially like the move of preloads to controllers. It's more explicit and something that's generally recommended in phoenix.

Just a potential concern with removing tasks from a project, but should be fine otherwise.

joshsmith:
Do you agree with breaking out the preloads into the model @begedin?

test/lib/code_corps_web/views/project_view_test.exs
joshsmith: This is breaking things in Ember. Even worse, I wanted to be able to specify the tasks `index` to load only `open` tasks by default, but this appears to break functionality if the tasks are not included on the project.

There's some work to be done here figuring out the best way forward.

begedin:
@JoshSmith If you mean, into the controller, then yeah, I absolutely do. Can't find a link for it, but pretty sure that's the recommended approach for phoenix

begedin:
~~To update myself, as I said on slack, nut sure the model module is the best place to put them, but we need them and I don't have a better idea, so I'm fine with placing them there.~~

Seconds later we agreed the view is a good spot.

joshsmith:
AuthToken

- [x] value

Category

- [x] slug

Comment

- [x] user
- [x] task
- [x] github_id

DonationGoal:

- [x] current (unique to project)
- [x] project_id

GithubAppInstallation

- [x] github_id (unique)
- [x] user_id
- [x] project_id

GithubComment

- [x] github_id (unique)
- [x] github_issue_id

GithubEvent

- [x] github_delivery_id
- [x] status

GithubIssue

- [x] github_id (unique)
- [x] github_repo_id
- [x] github_pull_request_id

GithubPullRequest

- [x] github_id
- [x] github_repo_id

GithubRepo

- [x] github_id (unique)
- [x] github_app_installation_id (used by InstallationRepositories event but not sure it should be)

OrganizationGithubAppInstallation

- [x] organization_id
- [x] github_app_installation_id

> These should probably be made a unique pair?

OrgnanizationInvite

- [x] code (unique)
- [x] email

Organization

- [x] approved
- [x] owner_id
- [x] slug (unique)

Preview

- [x] user_id

Project has:

- [x] slug (unique)
- [x] approved
- [x] organization_id

> This has a github_repo and github_owner – what are these?

StripeX:

- [x] id_from_stripe

StripeConnectSubscription, StripePlatformCustomer

- [x] user_id

StripeExternalAccount

- [x] id_from_stripe (unique)

StripeFileUpload

- [x] id_from_stripe (unique)

Task

- [x] github_issue_id
- [x] github_repo_id
- [x] project_id
- [x] task_list_id
- [x] number
- [x] order
- [x] archived
- [x] status

TaskList has:

- [x] inbox
- [x] project_id
- [x] order

User:

- [x] github_id (unique)
- [x] email (unique)

joshsmith:
Additionally, it doesn't look like we sort tasks by order, which is maybe a lost opportunity since that's expected on the front-end.

joshsmith:
Should `OrganizationGithubAppInstallation` have a unique pair of `organization_id` and `github_app_installation_id`?

`Project` has a `github_repo` and `github_owner` – what are these?

begedin:
>Should OrganizationGithubAppInstallation have a unique pair of organization_id and github_app_installation_id?

I think no. Consider this case.

On my github account, I have 2 repos, intended to be two separate projects in two organizations. I install the CC app once and then from each project's installation tab, can see that installation and connect it to that organization.

It's a rare case, but I think it is possible.

>Project has a github_repo and github_owner – what are these?

Probably remainder from before extracting `project_github_repo` or possibly even `github_repo`. We may have added them way early in this milestone.

joshsmith:
> Probably remainder from before extracting project_github_repo or possibly even github_repo. We may have added them way early in this milestone.

Can you add an issue for this?

begedin (approved):
Looks good. I see you decided to keep preloads in the controller. Works for me and makes sense, considering that's where they are used from.

PR #1124, opened 2017-10-27
Add Scout APM

joshsmith:
# What's in this PR?

Adds Scout performance monitoring.


config/scout_apm.exs
```diff
@@ -0,0 +1,12 @@
+# This configuration file is used for Scout APM.
+# See our help docs at http://help.apm.scoutapp.com for more information.
+# config/scout_apm.exs
+use Mix.Config
+
+config :scout_apm,
+  name: System.get_env("SCOUT_APP_NAME") || "", # The app name that will appear within the Scout UI
+  key: System.get_env("SCOUT_APP_KEY") || ""
```
begedin, line 8: You should maybe add these to `.env.example`

begedin (dismissed):
I have a minor suggestion, but otherwise, this looks good

config/scout_apm.exs
joshsmith: Added.

Issue #1132, opened 2017-10-28
Add unit tests for GitHub pagination


joshsmith:
# Problem

We need to add unit tests for the `GitHub.API` pagination that actually test that it paginates.

- Probably need to be able to specify `per_page` and `status` in the `options`
- Probably mock the returned headers
- Probably needs to be multiple fixtures for each page of the response
- Must test at least 3 pages to be able to get the full testing benefit here
- Should use the `baxterthehacker/public-repo` stub repository for response fixtures

Blocked by #1131

PR #1138, opened 2017-10-30
Switched the GitHub API to use a simple Gateway module

### LINKED ISSUE #1131: Add GitHub.API.Gateway which calls HTTPoison
# Problem

We should add an `API.Gateway` which simply has a request method that calls `HTTPosion`. Response parsing can be done in this module. We can then mock the gateway in tests, and mock `%HTTPoison.Response{}` so that we can include the headers (e.g. for pagination).

We'll need to refactor `GitHub.API.EagerAPI` to make use of this in `marshall_response`

- check if the headers contains a `"Links"` key
- if it does, it calls the code which makes additional requests for further pages

begedin:
Closes #1131, 
Progress on #1132 

This PR

- extracts a simple `API.Gateway` module from out of `GitHub.API`
- rewrites tests and the rest of the code to use/mock this module
- rewrites `API.EagerAPI` into `API.Pagination`, which is a module containing a set of utility functions to be used from `API`, so `Gateway` is only ever called from `API`
- rewrites `API.get_all` so it uses a `:head` request first, to further consolidate and decouple
- fixes a bunch of dialyzer issues found along the way


lib/code_corps/github/api/api.ex
```diff
@@ -2,19 +2,37 @@ defmodule CodeCorps.GitHub.API do
...
+
+    response_headers
+    |> Pagination.retrieve_total_pages()
+    |> Pagination.to_page_numbers()
+    |> Enum.map(&Pagination.add_page_param(options, &1))
+    |> Enum.map(&gateway().request(:get, url, "", headers, &1))
+    |> Enum.map(&marshall_response/1)
+    |> Enum.map(&Tuple.to_list/1)
+    |> Enum.map(&List.last/1)
+    |> List.flatten
```
joshsmith, line 34: I just read (oddly in `benchee`'s README) that `Enum.map |> List.flatten` is 1.83x slower than `Enum.flat_map`. Wonder if we should use that instead here?

lib/code_corps/github/api/api.ex
```diff
@@ -47,29 +65,32 @@ defmodule CodeCorps.GitHub.API do
...
+  defp marshall_response({:ok, %Response{body: body, status_code: status}}) when status in 200..299 do
     case body |> Poison.decode do
       {:ok, json} ->
         {:ok, json}
       {:error, _value} ->
         {:error, HTTPClientError.new(reason: :body_decoding_error)}
     end
   end
-  defp marshall_response({:ok, 404, _headers, body}) do
+  defp marshall_response({:ok, %Response{body: body, status_code: 404}}) do
```
joshsmith, line 80: Good fix here.

lib/code_corps/github/api/api.ex
```diff
@@ -47,29 +65,32 @@ defmodule CodeCorps.GitHub.API do
...
     case body |> Poison.decode do
       {:ok, json} ->
         {:error, APIError.new({status, json})}
       {:error, _value} ->
         {:error, HTTPClientError.new(reason: :body_decoding_error)}
     end
   end
+  defp marshall_response({:error, %Error{reason: reason}}) do
+    {:error, HTTPClientError.new(reason: reason)}
+  end
```
joshsmith, line 93: Much needed.

test/lib/code_corps/github/api/api_test.exs
```diff
@@ -4,87 +4,122 @@ defmodule CodeCorps.GitHub.APITest do
...
+    defmodule MockPaginationAPI do
+      def request(:head, "/one-page", _body, _headers, _opts) do
+        {:ok, %HTTPoison.Response{headers: [], status_code: 200}}
+      end
+      def request(:head, "/two-pages", _body, _headers, _opts) do
+        next = '<two-pages/?page=2>; rel="next"'
+        last = '<two-pages/?page=2>; rel="last"'
+
+        headers = [{"Link", [next, last] |> Enum.join(", ")}]
+        {:ok, %HTTPoison.Response{headers: headers, status_code: 200}}
```
joshsmith, line 90: This is all we needed really for the pagination test. IMO this satisfies most of what I wanted in #1132. What do you think @begedin?

joshsmith (dismissed):

joshsmith (approved):

Issue #1139, opened 2017-10-30
Investigate perf benefits of using jsonapi


snewcomer:
- [ ] investigate perf benefits with https://github.com/jeregrine/jsonapi

snewcomer:
```
Name                    ips        average  deviation         median         99th %
ja_serializer        4.02 K        0.25 ms    ±24.68%        0.23 ms        0.48 ms
jsonapi              0.21 K        4.87 ms   ±105.40%        4.18 ms        9.60 ms

Comparison:
ja_serializer        4.02 K
jsonapi              0.21 K - 19.60x slower
```

TaskListView with a lot of tasks shows that jsonapi is significantly slower.  However, when not dealing with relationships, jsonapi is faster.

```
Name                ips        average  deviation         median         99th %
jsonapi        213.45 K        4.69 μs  ±3440.93%           4 μs          10 μs
existings       38.04 K       26.29 μs   ±283.89%          23 μs          50 μs
Comparison: 
jsonapi             213.45 K
ja_serializer       38.04 K - 5.61x slower
```

mitchellhenke:
For the tests, is each rendering producing the same result?

I ask because the defaults between the two libraries are different, and so one may be rendering more or less data in different situations 🙂

snewcomer:
Hey @mitchellhenke.  Yeah jsonapi is rendering more data.

Here is the benchee [example](https://github.com/code-corps/code-corps-api/blob/jsonapi-test/priv/repo/benchit.exs#L12).

Moreover, this is based on the seeds for [Tasks](https://github.com/code-corps/code-corps-api/blob/jsonapi-test/priv/repo/seeds.exs#L222).

So a TaskList will have many tasks (~50).  Here is an example payload with one task:

jsonapi
[...]

PR #1140, opened 2017-10-30
Replace jaserializer with jsonapi

snewcomer:
Ref https://github.com/code-corps/code-corps-api/issues/1139

This is a very initial commit using jsonapi.  Can try it out by running `category_view_test.exs`.

Essentially, this lib gives us a `show` and `index` function for which we can call.

Couple of things.

1.  The [jsonapi](https://github.com/jeregrine/jsonapi) lib might need some TLC to make it easier to work with.
~~2. We use dasherized plural types.  The dasherized config is not working for relationships right now.~~ Solved with master branch

TODO:
- [ ] cleanup and w/ potentially abstractions for working with lib
- [ ] Identify PR's to lib
- [ ] move to separate view and perf test against category_view w/ ja_serializer

### DIFF
--- a/lib/code_corps_web/controllers/category_controller.ex
+++ b/lib/code_corps_web/controllers/category_controller.ex
@@ -16,7 +16,7 @@ defmodule CodeCorpsWeb.CategoryController do
   @spec show(Conn.t, map) :: Conn.t
   def show(%Conn{} = conn, %{"id" => id}) do
     with %Category{} = category <- Category |> Repo.get(id) |> preload() do
-      conn |> render("show.json-api", data: category)
+      conn |> render(CodeCorpsWeb.CategoryView, "show.json-api", %{ data: category, conn: conn, params: id })
     end
   end
 
--- a/lib/code_corps_web/views/category_view.ex
+++ b/lib/code_corps_web/views/category_view.ex
@@ -1,9 +1,13 @@
 defmodule CodeCorpsWeb.CategoryView do
   @moduledoc false
   use CodeCorpsWeb, :view
-  use JaSerializer.PhoenixView
+  use JSONAPI.View
 
-  attributes [:name, :slug, :description]
+  def render("show.json-api", %{ data: category, conn: conn, params: params }) do
+    __MODULE__.show(category, conn, params)
+  end
 
-  has_many :project_categories, serializer: CodeCorpsWeb.ProjectCategoryView, identifiers: :always
+  def fields, do: [:name, :slug, :description]
+  def type, do: "category"
+  def relationships, do: [project_categories: CodeCorpsWeb.ProjectCategoryView]
 end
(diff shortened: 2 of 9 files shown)


lib/code_corps_web/views/category_view.ex
```diff
@@ -1,9 +1,13 @@
...
 
-  attributes [:name, :slug, :description]
+  def render("show.json-api", %{ data: category, conn: conn, params: params }) do
+    __MODULE__.show(category, conn, params)
+  end
 
-  has_many :project_categories, serializer: CodeCorpsWeb.ProjectCategoryView, identifiers: :always
+  def fields, do: [:name, :slug, :description]
+  def type, do: "category"
+  def relationships, do: [project_categories: CodeCorpsWeb.ProjectCategoryView]
```
snewcomer: Will also try `includes` for relationships once we figure out how to use this lib.

lib/code_corps_web/views/category_view.ex
```diff
@@ -1,9 +1,13 @@
 defmodule CodeCorpsWeb.CategoryView do
   @moduledoc false
   use CodeCorpsWeb, :view
-  use JaSerializer.PhoenixView
+  use JSONAPI.View
 
-  attributes [:name, :slug, :description]
+  def render("show.json-api", %{ data: category, conn: conn, params: params }) do
+    __MODULE__.show(category, conn, params)
```
snewcomer: `params` is an unused arg to the `show` function.  Candidate for a PR to the lib.

lib/code_corps_web/controllers/category_controller.ex
```diff
@@ -16,7 +16,7 @@ defmodule CodeCorpsWeb.CategoryController do
   @spec show(Conn.t, map) :: Conn.t
   def show(%Conn{} = conn, %{"id" => id}) do
     with %Category{} = category <- Category |> Repo.get(id) |> preload() do
-      conn |> render("show.json-api", data: category)
+      conn |> render(CodeCorpsWeb.CategoryView, "show.json-api", %{ data: category, conn: conn, params: id })
```
snewcomer, line 19: I wonder if we can just call the `show` action provided by the [jsonapi](https://github.com/jeregrine/jsonapi/blob/master/lib/jsonapi/view.ex#L106) lib instead of calling render....?

joshsmith: Seems plausible. Have you tried this?

snewcomer:
https://github.com/jeregrine/jsonapi/issues/78

First step.  Unless my understanding is off, I am not sure if the lib could be improved for handling relationships.  It seems like it requires preloading all relationships down the tree when.  So for example, serializing `project-category` data, we only want to ask for the `project_category_id` from the `category` data instead of walking down the relationship tree.

Lmk if anybody sees it otherwise.

snewcomer:
Based on some of the perf findings with lots of tasks for a task_list view shown in the issue, was given the green light to work on `includes` with ja_serializer.  I'll open in another PR.

PR #1143, opened 2017-10-31
Sync repo

### LINKED ISSUE #971: When a project github repo is created, launch the sync process for that project and github repo
Blocked by #1131 

The process imports all issues and comments from the github repo into the codecorps project.

Later on, we should refines the process further so it potentially does not happen automatically, or, at least, requires confirmation from the user and shows some progress information on the UI.

This issue should serve as a good place for discussing it.

# Potential concerns

**The github repo that just got linked to the current project may have also been previously linked to other projects**

This is not really a problem, since we only sync from github to cc, not the other way around.

# Time assessments

- **4 hours** to create function (in a suitable new or existing module) which, for a provided `ProjectGithubRepo`
   - pulls list of issues for associated `GithubRepo`
   - for each issue, creates `Task` on associated `Project`
   - test function behavior
- **3 hours** to, on the ember side of things, add a "sync now" button on the project integration page, which hits a new `project/:id/sync` endpoint. The endpoint will respond with a list of `ProjectGithubRepo` records all marked as processed, so we push those into the store.
  - the `ProjectGithubRepo` component on the integration page now needs a `processing` state
- **4 hours to** implement endpoint behavior
  - load `Project`, preload all `ProjectGithubRepo` records, respond with those (we can use `update_all` with `returns: true`, so it returns updated records.
  - launch an async process which for each `ProjectGithubRepo`
    - launches function from step 1 
    - tags as processed when done
    - test

## Notes

We could simplify the processing status and shave an hour or two off by having one async process on the whole `Project` instead on individual `ProjectGithubRepo` records and have the `processing` flag on the project itself. However, that would end up being more hours in total, if we want to switch to step-based later.

# Future enhancements

_Before we have real time processing via channels_

- **2 hours** to have the integration page check for progress from the server, by refreshing the `ProjectGithubRepo` records periodically. Otherwise, MPV requires the user to refresh manually.

_Once we have some real time processing implemented via channels_

- **1 hour** to notify client via channel when a `ProjectGithubRepo` is processed
- **0 additional hours** for the client to handle this, since presumably, this will be implemented as a generalized solution at a service/adapter level

### LINKED ISSUE #1111: Implement GitHub.API.Repository.comments
# Problem

We can fetch all comments for a repository, instead of dealing with individual issues. Due to  pagination and the fact that most issues have less than a page of comments, fetching from repositories would save up on request count and reduce the chance of hitting the rate limits.

https://developer.github.com/v3/issues/comments/#list-comments-in-a-repository

Blocked by #1131

### LINKED ISSUE #1133: Implement GitHub.API.Repository.pull_requests
# Problem

We need to fetch all pull requests for a repository.

https://developer.github.com/v3/pulls/#list-pull-requests

Blocked by #1131

### LINKED ISSUE #1136: Set GithubRepo syncing statuses during sync
# Problem

We need to set the `GithubRepo` syncing statuses during the sync of the repo.

Blocked by #1135 #1111 #1133 #1134

joshsmith:
# What's in this PR?

This is almost good to merge now.

Closes #1111 
Closes #1133 
Closes #1136 
Closes #971


lib/code_corps/model/github_pull_request.ex
```diff
@@ -73,4 +74,14 @@ defmodule CodeCorps.GithubPullRequest do
...
+
+  defp add_merged(%Ecto.Changeset{changes: %{merged: merged}} = changeset) when is_boolean(merged) do
+    changeset
+  end
+  defp add_merged(%Ecto.Changeset{changes: %{merged_at: _}} = changeset) do
+    changeset |> put_change(:merged, true)
+  end
+  defp add_merged(%Ecto.Changeset{} = changeset) do
+    changeset |> put_change(:merged, false)
+  end
```
joshsmith: Apparently GitHub does not pass `merged` along when listing the records.

Can also remove:

```
additions changed_files comments commits deletions mergeable_state review_comments 
```

lib/code_corps/github/sync/sync.ex
```diff
@@ -99,18 +101,53 @@ defmodule CodeCorps.GitHub.Sync do
...
+    {:ok, comments} = GitHub.API.Repository.comments(repo)
+    Enum.map(comments, &sync_comment(&1, repo, github_issues, tasks))
+  end
+
+  def sync_comment(comment, repo, github_issues, tasks) do
+    Multi.new
+    |> Multi.merge(__MODULE__, :return_repo, [repo])
+    |> Multi.merge(__MODULE__, :find_issue, [comment, github_issues])
+    |> Multi.merge(__MODULE__, :find_tasks, [comment, tasks])
+    |> Multi.merge(GitHub.Sync.Comment, :sync, [comment])
```
JoshSmith: This is the part that appears not to be working. I got everything up until here working.

lib/code_corps/github/sync/sync.ex
```diff
@@ -99,18 +101,53 @@ defmodule CodeCorps.GitHub.Sync do
...
+    {:ok, pulls} = GitHub.API.Repository.pulls(repo)
+    github_pull_requests = Enum.map(pulls, &sync_pull(&1, repo))
+    results = sync_issues(repo, github_pull_requests)
+    github_issues = for {:ok, %{github_issue: github_issues}} <- results, do: github_issues
+    github_issues = Enum.map(github_issues, fn issue -> {:ok, issue} end)
+    tasks = for {:ok, %{tasks: tasks}} <- results, do: tasks
+    tasks = tasks |> List.flatten
+    tasks = Enum.map(tasks, fn task -> {:ok, task} end)
+    sync_comments(repo, github_issues, tasks)
+    repo
```
JoshSmith: Inside here we could sync each step, modify the `GithubRepo` sync state, and then return the repo when done.

joshsmith:
I will mention that it is really, _really_ hard to debug problems happening inside of these `Multi` fns.

joshsmith:
I fixed an issue where `marshall_result` was not providing us with the `github_issue` result for `github_pull_requests`, but there appears to be a result missing. For our `test` repo we should be seeing 31 issues created, but it looks like there are only 30, with perhaps a missed validation error.

I wanted to at least get this spike working before worrying too much about error handling. Hard to get it working without it, though.

joshsmith:
I managed to debug it. I'm getting a single task back that's got an `{:error, :validating_tasks}` that's failing due to the timestamp from GitHub being before the `modified_at`.

It turns out this is due to a bug: we're not setting the `modified_from` field to `"code_corps"` for anything that's being update from the Code Corps app.

Of course, this means my strategy here is probably flawed. We won't get back updated issues/tasks when a sync isn't happening for the first time. We _could_ modify this by returning back the _matched_ issues regardless of whether the sync happens. That would allow my strategy here to continue to work. But it's probably worth stepping back and re-evaluating this approach.

That said, the general outline of how to approach the problem is there as a start.

joshsmith:
I ran a `mix ecto.reset` and it works all the way through. So this is effectively spiked. But lots of issues obviously remain with how this works.

lib/code_corps/github/sync/sync.ex
```diff
@@ -99,18 +100,54 @@ defmodule CodeCorps.GitHub.Sync do
...
+    results = sync_issues(repo, github_pull_requests)
+    github_issues = for {:ok, %{github_issue: github_issue}} <- results, do: github_issue
+    github_issues = Enum.map(github_issues, fn issue -> {:ok, issue} end)
+    tasks = for {:ok, %{tasks: tasks}} <- results, do: tasks
+    tasks = tasks |> List.flatten
+    tasks = for %CodeCorps.Task{} = task <- tasks, do: task
+    tasks = Enum.map(tasks, fn task -> {:ok, task} end)
+    sync_comments(repo, github_issues, tasks)
+    repo
+  end
```
joshsmith: Inside here we could sync each step, modify the GithubRepo sync state, and then return the repo when done.

begedin:
@JoshSmith I managed to spike out a less coupled, fully working `GithubRepo` sync approach. I've encountered several difficulties while doing this, one of which I'm still trying to find a solution for.

# Problems


## `/repos/:owner/:name/comments` retrieves **repository comments**, NOT issue comments

Really, in hindsight, it makes sense, but it completely dodged my attention. The comments retrieved by that endpoint are code comments, not comments from within issues.

I had to implement an `API.Issue.comments` endpoint and then also implemented a utility endpoint `API.Repository.issue_comments` which preloads all `GithubIssue` records for a provided `GithubRepo`, then calls `API.Issue.comments/1` for each of them and aggregates the results.

This makes the request somewhat slow, since it fires off parallel `:head` requests equal to the issue count on the repo, but it's still faster than doing each of those individually.

To allow for this, I also had to add a function clause to the `GithubComment.create_or_update_comment` which only requires a payload and associates with an issue matching by `Repo.get_by(&1, url: issue_url)`. This decouples the function and we can likely switch the webhook syncing code to using it as well. More importantly, it allows for a simple pipe during the repo sync process.

## i believe **we need** a `sync_state` on the `ProjectGithubRepo` as well

Once a `GithubRepo` is synced, and it receiving webhooks, there is no need to sync it again. When a project is connected to it, the only thing that's left to do is to create  necessary `Task` and `Comment` records on that `Project`.

I added a virtual `sync_state` field to the `ProjectGithubRepo` for now, just so I could spike things out, but we should a proper field. I'm thinking the states are 

```
~w(unsynced syncing_tasks errored_syncinc_tasks syncinc_comments errored_syncinc_comments)
```

## our current sync flow **does not allow** creating `Task` or `Comment` records separate from `GithubIssue` and `GithubComment` respectively.

This is the issue that I'm stuck with.

The major part of the reason we introduced `GithubIssue` and `GithubComment` as separate entities was to decouple this process and allow easier updating when a single issue needs to sync as a task across multiple projects, etc.

However, due to the way we match the user, there is no way for us to actually separate the two. The user is matched from the initial payload, but no user information is stored on the `GithubIssue`/`GithubComment` for later use.

I considered several potential solutions for this

### Make fetching payloads a process separate from the main syncing flow

Basically, we would sync a `project_github_repo`, along the lines of

```Elixir
with {:ok, pr_payloads} <- repo |> fetch_pull_requests,
        {:ok, issue_payloads} <- repo |> fetch_issues,
        {:ok, comment_payloads} <- repo |> fetch_comments,
        {:ok, %GithubRepo{sync_state: "receiving_webhooks"} = repo} <- repo |> sync_repo(pr_payloads, issue_payloads, comment_payloads),
        {:ok, %ProjectGithubRepo{sync_state: "synced"} = project_github_repo} <- project_github_repo |> sync_project_github_repo(issue_payloads, comment_payloads) 
do
  {:ok, project_githubrepo}
end
```

Then, we would sync a project by syncing each of its `project_github_repo` children.

Really, though, this is a bandaid fix, since it does not resolve the actual problem of too tight coupling and just allows us to implement the sync process sooner.

### Try to make it work like webhooks work, keep it coupled and perform the sync of `Task/GithubIssue` as well as `Comment/GithubComment` in unison

I don't often say this, but I'm strongly against this option. 

The whole process is complicated and hard to understand as is, and this will only make it worse.

### Add `github_user_id` field to `GithubIssue` and `GithubComment`

This would be a temporary fix, but it would possibly allow us to defer the full solution into a separate PR.

Basically, we store a github user id field with the records, allowing us to match later. However, this does not provide us with enough information for the case when the author of the comment/issue does not have a codecorps accoutn linked to github, so that case would be unhandled.

I don't favor this solution, but wanted to mention it for the sake of argument.

### Introduce  `GithubUser`

#### As part of this PR we

- introduce a `GithubUser` record
- as we sync comments and issues, a `GithubUser` record is found or created alongside each of them

The record basically holds the fields we've added to `User` 

```Elixir
field :avatar_url, :string
field :email, :string
field :github_id, :integer
field :username, :string
field :type, :string, default: "User" # or "Bot"
```

Once we start syncing tasks and comments, we match a `GithubUser` with a `User` (find or create) and associate appropriately.

#### As part of subsequent PRs we can then

- Switch the webhook syncing code to use `GithubUser` to. We could do it here as well, but we don't have to.
- Potentially decouple our webhook syncing code, consolidate further, move out of multis, etc.
- Potentially clean up our `User` model a bit
- ???

begedin:
We could speed up comment fetching with `Repository.issue_comments` by storing `comment_count` on the `GithubIssue` (there is a `"comments": comment_count` key in the github payload) and then only sync github issues with comment count greater than 0.

begedin:
From our chat offline, what we need to do here

- [x] Add sync state field to project github repo as well, with changesets, 
- [x] make `Sync.mark/1` work correctly
- [x] Add `GithubUser`
- [x] Modify `GithubIssue`/`GithubComment` sync to match or create a `GithubUser`, to decouple from `Comment`/`Task` sync
- [x] Modify `Comment`/`Task` sync to match against `GithubUser` and find or create `User`
- [x] Finalize project github repo sync
- [x] Remove `Repository.comments` and `Issue.comments` if we end up not using them
- [ ] Add fns for counting number of records being worked and add to syncing details
- [ ] Write tests
- [ ] Create issue to decouple webhook syncing using new sync process as guideline, making use of `GithubUser`
- [ ] Create issue to extract `User` `:github_` fields into `GithubUser` (maybe needs a migration)

joshsmith:
For adding `GithubUser` we'll need to:

- [x] add references to `User`, `GithubIssue`, `GithubPullRequest`, `GithubComment`
- [x] add all these to models
- [ ] add model tests
- [ ] add all these to views and view tests

joshsmith:
This is going to need some significant cleanup of changesets and the way changes are propagated to the various GithubX models during sync.

joshsmith:
This is mostly working, but on a first pass through a project, it seems to not create tasks in the `syncing_tasks` step, moves onto `syncing_comments`, and then hits a `nil` Task that it can't match on:

```
** (WithClauseError) no with clause matching: nil
    (code_corps) lib/code_corps/github/sync/comment/comment/comment.ex:63: CodeCorps.GitHub.Sync.Comment.Comment.find_or_create_comment/2
    (elixir) lib/enum.ex:1255: Enum."-map/2-lists^map/1-0-"/2
    (code_corps) lib/code_corps/github/sync/comment/comment/comment.ex:58: CodeCorps.GitHub.Sync.Comment.Comment.sync_project_github_repo/1
    (code_corps) lib/code_corps/github/sync/sync.ex:163: CodeCorps.GitHub.Sync.sync_project_github_repo/1
```

joshsmith:
I wonder if it is worth adding in a `synced_at` timestamp to the `GithubRepo` and `ProjectGithubRepo`.

joshsmith:
My first goals tomorrow:

- [x] get the existing tests passing
- [x] write an acceptance test for the full sync
- [ ] write unit tests for
  - [x] views
  - [ ] models
- [ ] refactor
- [x] add record counts to sync state

lib/code_corps/github/adapters/issue.ex
```diff
@@ -54,6 +54,12 @@ defmodule CodeCorps.GitHub.Adapters.Issue do
     payload |> MapTransformer.transform(@task_mapping)
   end
 
+  def to_issue_attrs(%GithubIssue{} = github_issue) do
+    github_issue
+    |> Map.from_struct
+    |> MapTransformer.transform_inverse(@issue_mapping)
+  end
```
begedin, line 66: The naming makes sense, but considering we were switching already, I'm thinking maybe we want to write a behavior module and document what an adapter needs to look like and what the naming convention is.

I'm just imagining where, in a couple of weeks, I write a new one and end up using a new scheme because I wasn't 100% what the old one is.

begedin (commented):
This is pretty much implemented the way I would continue off.

I got a suggested issue for adapter behavior, but that's about it.

joshsmith:
This is missing tests for the `GitHub.Sync` that hit the error states for `repo` and `project_repo` sync.

lib/code_corps/accounts/changesets.ex
```diff
@@ -15,11 +15,21 @@ defmodule CodeCorps.Accounts.Changesets do
...
+    |> Changeset.change(params |> Adapters.User.to_user())
     |> Changeset.put_change(:sign_up_context, "github")
     |> Changeset.validate_inclusion(:type, ["bot", "user"])
     |> RandomIconColor.generate_icon_color(:default_color)
     |> Changeset.unique_constraint(:email)
+    |> Changeset.assoc_constraint(:github_user)
+    |> unique_github_constraint()
+  end
+
+  def update_with_github_user_changeset(struct, %{} = params) do
```
joshsmith: Missing spec and docs.

lib/code_corps/github/adapters/comment.ex
```diff
@@ -36,6 +37,12 @@ defmodule CodeCorps.GitHub.Adapters.Comment do
     {:url, ["url"]}
   ]
 
+  def to_comment_attrs(%GithubComment{} = github_comment) do
```
joshsmith: Missing spec and docs.

lib/code_corps/github/adapters/issue.ex
```diff
@@ -54,6 +54,12 @@ defmodule CodeCorps.GitHub.Adapters.Issue do
     payload |> MapTransformer.transform(@task_mapping)
   end
 
+  def to_issue_attrs(%GithubIssue{} = github_issue) do
+    github_issue
```
joshsmith: Missing spec and docs.

lib/code_corps/github/sync/comment/comment/comment.ex
```diff
@@ -41,6 +47,47 @@ defmodule CodeCorps.GitHub.Sync.Comment.Comment do
     |> ResultAggregator.aggregate
   end
 
+  def sync_project_github_repo(%ProjectGithubRepo{github_repo: %GithubRepo{} = _} = project_github_repo) do
```
joshsmith: Missing spec and doc.

lib/code_corps/github/sync/issue/task/task.ex
```diff
@@ -32,6 +34,36 @@ defmodule CodeCorps.GitHub.Sync.Issue.Task do
     |> ResultAggregator.aggregate
   end
 
+  def sync_project_github_repo(%ProjectGithubRepo{github_repo: %GithubRepo{} = _} = project_github_repo) do
```
joshsmith: Missing spec and doc.

lib/code_corps/github/sync/sync.ex
```diff
@@ -99,22 +103,94 @@ defmodule CodeCorps.GitHub.Sync do
...
   end
 
-  @doc false
-  def return_repo(_, repo) do
-    Multi.new
-    |> Multi.run(:repo, fn _ -> {:ok, repo} end)
+
+  @count_fields [:syncing_comments_count, :syncing_issues_count, :syncing_pull_requests_count]
+
+  defp build_sync_params(sync_state, opts) do
```
joshsmith: Missing spec and doc.

lib/code_corps/github/sync/sync.ex
```diff
@@ -99,22 +103,94 @@ defmodule CodeCorps.GitHub.Sync do
...
+    Enum.reduce @count_fields, %{sync_state: sync_state}, fn field, acc ->
+      put_sync_opt(opts, field, acc)
+    end
+  end
+
+  defp put_sync_opt(opts, key, map) do
```
joshsmith: Missing spec and doc.

lib/code_corps/github/sync/sync.ex
```diff
@@ -99,22 +103,94 @@ defmodule CodeCorps.GitHub.Sync do
...
+    case Keyword.get(opts, key) do
+      nil -> map
+      count -> map |> Map.put(key, count)
+    end
+  end
+
+  @spec sync_repo(GithubRepo.t) :: {:ok, GithubRepo.t}
+  def sync_repo(%GithubRepo{} = repo) do
```
joshsmith: Missing doc.

lib/code_corps/github/sync/sync.ex
```diff
@@ -99,22 +103,94 @@ defmodule CodeCorps.GitHub.Sync do
...
+      {:error, :fetch_pull_requests} -> repo |> mark_repo("errored_fetching_pull_requests")
+      {:error, :sync_pull_requests} -> repo |> mark_repo("errored_syncing_pull_requests")
+      {:error, :fetch_issues} -> repo |> mark_repo("errored_fetching_issues")
+      {:error, :sync_issues} -> repo |> mark_repo("errored_syncing_issues")
+      {:error, :fetch_comments} -> repo |> mark_repo("errored_fetching_comments")
+      {:error, :sync_comments} -> repo |> mark_repo("errored_syncing_comments")
+    end
+  end
+
+  def sync_project_github_repo(%ProjectGithubRepo{} = project_github_repo) do
```
joshsmith: Missing spec and doc.

lib/code_corps/validators/time_validator.ex
```diff
@@ -11,6 +11,13 @@ defmodule CodeCorps.Validators.TimeValidator do
   def validate_time_after(%{data: data} = changeset, field) do
     previous_time = Map.get(data, field)
     current_time = Changeset.get_change(changeset, field)
+    case current_time do
+      nil -> changeset
+      _ -> do_validate_time_after(changeset, field, previous_time, current_time)
+    end
+  end
```
joshsmith, line 18: Needs updated unit tests.

lib/code_corps_web/controllers/project_github_repo_controller.ex
```diff
@@ -28,6 +29,8 @@ defmodule CodeCorpsWeb.ProjectGithubRepoController do
          {:ok, :authorized} <- current_user |> Policy.authorize(:create, %ProjectGithubRepo{}, params),
          {:ok, %ProjectGithubRepo{} = project_github_repo} <- create_project_repo_changeset(params) |> Repo.insert do
 
+      TaskSupervisor.start_child(:background_processor, fn -> CodeCorps.GitHub.Sync.sync_project_github_repo(project_github_repo) end)
```
joshsmith: This is not tested anywhere right now.

joshsmith (commented):
Some notes on things we can still improve here.

lib/code_corps/github/sync/sync.ex
```diff
@@ -99,22 +103,94 @@ defmodule CodeCorps.GitHub.Sync do
...
+    %ProjectGithubRepo{github_repo: %GithubRepo{} = repo} = project_github_repo =
+      project_github_repo
+      |> preload_project_github_repo
+
+    with {:ok, project_github_repo} <- project_github_repo |> mark_project_repo("syncing_github_repo"),
+         {:ok, %GithubRepo{sync_state: "receiving_webhooks"}} <- repo |> sync_repo(),
+         project_github_repo <- Repo.get(ProjectGithubRepo, project_github_repo.id) |> preload_project_github_repo(),
```
begedin, line 212: I'd actually prefer something like 

```Elixir
preloads <- [:project, github_repo: [:github_app_installation, [github_issues: [:github_comments, :github_user]]]],
project_github_repo <- project_github_repo |> Repo.preload(preloads)
```

I think it's actually easier to follow. Either that, or renaming `preload_project_gihub_repo` to `load_associations`. 

Seems somewhat more explicit to me.

lib/code_corps/github/sync/sync.ex
```diff
@@ -99,22 +103,94 @@ defmodule CodeCorps.GitHub.Sync do
...
+         {:ok, project_github_repo} <- project_github_repo |> mark_project_repo("syncing_tasks"),
+         {:ok, _tasks} <- project_github_repo |> Sync.Issue.Task.sync_project_github_repo() |> sync_step(:sync_tasks),
```
begedin: Trying to read `Sync.Issue.Task.sync_project_github_repo` makes me think we may have nested too deep.

```Elixir
project_github_repo |> Sync.TaskSyncer.sync
```

Looks a bit clearer, but I'm still not happy with it. However, the we could alias `Sync.TaskSyncer` and just call 

```Elixir
project_github_repo |> TaskSyncer.sync
```

What do you think about flattening out our Sync namespace that way? Nesting `Task` and `GithubIssue` under `Issue` together seems like we did just for the sake of them being semantically related, but logically, we probably want the two processes as decoupled as possible.

```
sync/sync.ex
sync/task_syncer.ex
sync/github_issue_syncer.ex
```

etc. looks nicer to me.

For now, as part of this PR, I'm thinking we 

```Elixir
alias CodeCorps.GitHub.Sync.Issue.Task, as: TaskSyncer
```

and do the same for other modules. What do you think?

lib/code_corps/github/sync/sync.ex
```diff
@@ -145,7 +221,10 @@ defmodule CodeCorps.GitHub.Sync do
   @spec marshall_result(tuple) :: tuple
   defp marshall_result({:ok, %{comments: comments}}), do: {:ok, comments}
   defp marshall_result({:ok, %{deleted_comments: _, deleted_github_comment: _}}), do: {:ok, nil}
+  defp marshall_result({:ok, %{github_pull_request: _, github_issue: _}} = result), do: result
   defp marshall_result({:ok, %{github_pull_request: pull_request}}), do: {:ok, pull_request}
+  defp marshall_result({:ok, %{github_issue: _, tasks: _}} = result), do: result
+  defp marshall_result({:ok, %{github_issue: github_issue}}), do: {:ok, github_issue}
```
begedin: Are these still needed? I think we didn't touch transactions in this PR

begedin (commented):
A couple of ideas I have, plus a potential change that needs to be reverted, but other than that, this looks great!

PR #1146, opened 2017-11-02
Include tasks in task_list view

snewcomer:
Add `include` key to tasks for the task_list view

## References
Fixes #

Progress on: #

### DIFF
--- a/lib/code_corps_web/controllers/task_list_controller.ex
+++ b/lib/code_corps_web/controllers/task_list_controller.ex
@@ -28,7 +28,7 @@ defmodule CodeCorpsWeb.TaskListController do
     end
   end
 
-  @preloads [:tasks]
+  @preloads [{:tasks, [:comments, :github_pull_request, :task_skills, :user_task]}]
 
   def preload(data) do
     Repo.preload(data, @preloads)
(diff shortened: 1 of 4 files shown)


lib/code_corps_web/controllers/task_list_controller.ex
```diff
@@ -28,7 +28,7 @@ defmodule CodeCorpsWeb.TaskListController do
     end
   end
 
-  @preloads [:tasks]
+  @preloads [{:tasks, [:comments, :github_pull_request, :task_skills, :user_task]}]
```
snewcomer, line 31: This is using nested preloads.  I am not sure if we should just add the whole list of task preloads or not.  Was just adding them based on the feedback from the tests.  @begedin thoughts?

snewcomer: ah these are the ones marked with `identifier: always`

snewcomer: @JoshSmith would you also `include` the task's relationships?

joshsmith: The real answer here is that it doesn't much matter what I think, but what the performance implications are. Testing this out in the browser with Scout running and doing more precise benchmarking will combined yield a better answer.

snewcomer:
Closing this in favor of #1187

PR #1148, opened 2017-11-05
Syncing does not match ResultAggregator errors

joshsmith:
# What's in this PR?

Right now this PR just has a failing test case demonstrating that syncing does not match `ResultAggregator` errors in the form of `{:error, {list, list}}`.

This also indicates a much larger problem lurking beneath the surface: once a project disconnects from GitHub, the sync logic when reconnecting will not work if anything has changed. Even worse, we have no way of queue up a sync of the newly re-connected tasks back up to GitHub in order to bring everything back into place.


joshsmith:
We should use changes in #1198 as a basis for thinking about how to implement an audit log here and raise any errors in the syncing.

PR #1149, opened 2017-11-05
Add tests for time validator

### LINKED ISSUE #1130: Add tests for TimeValidator
# Problem

Our TimeValidator has changed to check whether the time is after or equal to the current time. We likely need to refactor it some, and we need to add some testing for this explicitly.

joshsmith:
# What's in this PR?

Refactors and adds tests for the `TimeValidator`.

## References
Fixes #1130


snewcomer (approved):
LGTM!

PR #1155, opened 2017-11-06
Archived tasks don't need task lists or order

### LINKED ISSUE #1150: Make task lists optional for archived tasks
# Problem

When a `Task` has `archived: true` in its changeset, the `TaskList` association should be made optional.

- change `order_task` in `Task.changeset` to `set_order` and from that function:
  - add `validate_required([:task_list_id])` here instead of the main `changeset`
  - call `order_task` if `archived: false`
  - set the `order` to `NULL` if `archived: false`
- add tests for the changesets

### LINKED ISSUE #1154: Unset the order when a task is archived
# Problem

When a task is archived, the `order` value (its position in a task list) should be unset, since it no longer belongs to a task list.

joshsmith:
# What's in this PR?

- Moves `:archived` casting up into `changeset` since it can happen on creation (e.g. importing from GitHub)
- Removes any `task_list_id` or `order` when archived
- Moves required validation on `task_list_id` to only the case when task is not archived

## References
Fixes #1150 
Fixes #1154

### DIFF
--- a/test/lib/code_corps/model/task_test.exs
+++ b/test/lib/code_corps/model/task_test.exs
@@ -16,18 +16,36 @@ defmodule CodeCorps.TaskTest do
     end
 
     test "renders body html from markdown" do
-      user = insert(:user)
-      project = insert(:project)
-      task_list = insert(:task_list)
       changes = Map.merge(@valid_attrs, %{
         markdown: "A **strong** body",
-        project_id: project.id,
-        task_list_id: task_list.id,
-        user_id: user.id
+        project_id: 1,
+        task_list_id: 1,
+        user_id: 1
       })
       changeset = Task.changeset(%Task{}, changes)
       assert changeset.valid?
       assert changeset |> get_change(:body) == "<p>A <strong>strong</strong> body</p>\n"
+    end
+
+    test "removes the order and task list when the task is archived" do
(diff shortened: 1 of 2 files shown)


test/lib/code_corps/model/task_test.exs
```diff
@@ -46,6 +64,26 @@ defmodule CodeCorps.TaskTest do
...
+        task_list_id: task_list.id,
+        user_id: user.id
+      })
+      changeset = Task.create_changeset(%Task{}, changes)
+      assert changeset.valid?
+      {:ok, %Task{order: order}} = Repo.insert(changeset)
+
+      # We really want to test the order is set, but we have no good way to
+      # test this since the column default is `0`
+      assert order !== 0
```
snewcomer, line 85: Ah so assert order fails.  Would have thought since refute only handles nil and false that assert would have worked!  :walking_man:

snewcomer (commented):
nice!

test/lib/code_corps/model/task_test.exs
joshsmith: `assert order` succeeds but that's insufficient to tell us anything. Since the default is `0` it could just be set to `0`, but instead it will receive a random ranked value from `EctoOrdered`.

snewcomer (approved):

PR #1157, opened 2017-11-06
Add pull_requests && done to task_list

### LINKED ISSUE #1151: Add pull_requests boolean column to TaskList
# Problem

We need to add a `pull_requests` `boolean` column to `TaskList`, similar to the `inbox` column, to indicate whether the column should be the recipient of new and open pull requests.

- Add the column where the default is `false`
- Add the column to the `default_task_lists` in `TaskList` model
  - our current "In Progress" column should be the `pull_requests: true` one here
- Add model tests
- Add to the views
- Add view tests
- We'll need to add migrations for existing projects to add this (can look to the `priv/repo/add_default_tasks_to_projects.exs` for inspiration).
  - we want the columns titled "In Progress" to be `pull_request: true`
  - for each project, we should find any tasks that have an open `GithubIssue` with a `GithubPullRequest` and move them to this list

### LINKED ISSUE #1152: Add done boolean column to TaskList
# Problem

We need to add a `done` `boolean` column to `TaskList`, similar to the `inbox` column, to indicate whether the column should be the recipient of completed tasks and issues.

- Add the column where the default is `false`
- Add the column to the `default_task_lists` in `TaskList` model
  - our current "Done" column should be the `done: true` one here
- Add model tests
- Add to the views
- Add view tests
- We'll need to add migrations for existing projects to add this (can look to the `priv/repo/add_default_tasks_to_projects.exs` for inspiration).
  - we want the columns titled "Done" to be have `done: true`
  - for each project, we should find any tasks that are `closed` and move them to this list

snewcomer:
- [x] done boolean
- [x] pull_requests boolean
- [x] script for modifying task_lists and project tasks
- [x] unique constraints for pairs of (project_id, inbox), (project_id, pull_requests), (project_id, done)

Fixes #1151, fixes #1152


lib/code_corps/model/task_list.ex
```diff
@@ -52,7 +56,7 @@ defmodule CodeCorps.TaskList do
 
   def create_changeset(struct, params) do
     struct
-    |> cast(params, [:inbox])
+    |> cast(params, [:inbox, :done, :pull_requests])
```
joshsmith: Would alpha-order here.

priv/repo/modify_task_list_done.exs
```diff
@@ -0,0 +1,40 @@
...
+    |> preload(:task_lists)
+    |> Repo.all()
+    |> Enum.each(&update_tasks_to_done(&1))
+  end
+
+  defp update_tasks_to_done(project) do
+    Task
+    |> CodeCorps.Helpers.Query.project_filter(%{project_id: project.id})
+    |> Repo.all()
+    |> Enum.each(&assign_task_to_done(&1, Enum.at(project.task_lists, 3)))
```
joshsmith: What's this doing?

lib/code_corps/model/task_list.ex
```diff
@@ -30,10 +32,12 @@ defmodule CodeCorps.TaskList do
...
         inbox: false,
+        pull_requests: true,
         name: "In Progress",
         position: 3
       }, %{
         inbox: false,
+        done: true,
         name: "Done",
         position: 4
       }
```
joshsmith: I would probably remove `inbox: false` from the others here for consistency.

priv/repo/modify_task_list_done.exs
```diff
@@ -0,0 +1,40 @@
...
+  defp assign_task_to_done(task, task_list_done) do
+    Task.changeset(task, %{task_list_id: task_list_done.id})
+    |> Repo.update()
+  end
+
+  defp update_done_column(%TaskList{name: "Done"} = task_list) do
+    TaskList.create_changeset(task_list, %{done: true})
+    |> Repo.update()
+  end
+  defp update_done_column(task_list), do: task_list
```
joshsmith: This is good, but will also need a similar fn for `%TaskList{name: "In Progress"}` to make it the `pull_requests` list.

lib/code_corps/model/task_list.ex
snewcomer: @JoshSmith remove all `inbox: false`?

priv/repo/modify_task_list_done.exs
```diff
@@ -11,6 +11,10 @@ defmodule CodeCorps.Repo.Seeds.ModifyTaskListDone do
     |> Repo.all()
     |> Enum.each(&update_done_column/1)
 
+    TaskList
+    |> Repo.all()
+    |> Enum.each(&update_pr_column/1)
+
```
joshsmith: This could be condensed down to:

```
TaskList
|> Repo.all()
|> Enum.each(&update_done_column/1)
|> Enum.each(&update_pr_column/1)
```

priv/repo/modify_task_list_done.exs
begedin: Calls `assign_task_to_done(&1, 3rd_task_list_in_project)`, which is, I assume, the task list with `done: true`. It would, however, be preferable to call `update_tasks_to_done/1` with that task list as an argument.

begedin:
@snewcomer @JoshSmith 

I've used the migration script you wrote, Scott, and moved it's behavior into the migrations themselves. I did not get a chance to comment on the issues this PR resolves, so the instructions there do not match what I did, but I really feel this should be our preferable approach to handling migrations.

The code that migrates the data uses schema names instead of our schema modules, so it's agnostic to our actual modules and works regardless of what the rest of our code looks like.

Due to separate `up` and `down` steps, we are also able to rollback and I was able to test it using _semi-real_ data (by syncing a mock repo and trying it out on that).

test/lib/code_corps/model/task_list_test.exs
```diff
@@ -16,12 +17,122 @@ defmodule CodeCorps.TaskListTest do
...
+  test "defaults :inbox to 'false'" do
+    {:ok, record} =
+      %TaskList{} |> TaskList.changeset(@valid_attrs) |> Repo.insert
+    assert record.inbox == false
+  end
+
+  test "defaults :pull_requests to 'false'" do
+    {:ok, record} =
+      %TaskList{} |> TaskList.changeset(@valid_attrs) |> Repo.insert
+    assert record.pull_requests == false
```
begedin, line 35: @snewcomer Your assertion here was `refute record.pull_requests`, but that one would have passed even if `record.pull_requests == nil`, so we want a more explicit one.

begedin (approved):
@snewcomer Great work. I made some changes to how migrations are done, worded a couple of your tests differently and cleaned up.

This is about done, so I'm approving your part from my end. @JoshSmith can review and approve my part.

begedin (dismissed):
Just adding a blank "request changes" review here, so we don't accidentally merge before @JoshSmith reviews.

priv/repo/migrations/20171106050209_add_pull_requests_to_task_list.exs
```diff
@@ -0,0 +1,53 @@
...
+        t in "tasks",
+        where: [project_id: ^project_id],
+        where: not is_nil(t.github_issue_id),
+        inner_join:
+          gi in "github_issues",
+          on: t.github_issue_id == gi.id,
+          where: not is_nil(gi.github_pull_request_id),
+        update: [set: [task_list_id: ^pr_list_id]]
+      ) |> Repo.update_all([])
+    end)
```
begedin, line 46: @JoshSmith I really feel this is the way to go with migrations.

If our schema modules do not match the schemas in any migration step, the code will fail, but if we go like this, "schemaless",  the only case where the migration step would not match the actual state of the schema were if we somehow run it manually, out of order, which would be incorrect usage of the migration.

I understand using actual schema modules and changesets in a seed script, but they should not be used in migrations.

test/lib/code_corps/model/task_list_test.exs
snewcomer: :man_facepalming:

test/lib/code_corps/model/task_list_test.exs
```diff
@@ -43,24 +44,104 @@ defmodule CodeCorps.TaskListTest do
...
+        |> Repo.insert
+
+      {:error, changeset} =
+        %TaskList{}
+        |> TaskList.create_changeset(attrs)
+        |> Changeset.put_change(:project_id, project_id)
+        |> Repo.insert
+
+      refute changeset.valid?
+        assert changeset |> Map.get(:errors) |> Keyword.get(:done)
```
snewcomer: minor spacing...unless its the GUI.

priv/repo/migrations/20171106103153_add_unique_constraints_to_specific_task_lists.exs
```diff
@@ -0,0 +1,21 @@
...
+  @moduledoc false
+
+  use Ecto.Migration
+
+  def change do
+    # There is already a "task_lists_project_id_index", so we name explicitly
+
+    create unique_index(
+      "task_lists", [:project_id],
+      where: "done = true", name: "task_lists_project_id_done_index")
```
snewcomer, line 11: Nice work!!

priv/repo/migrations/20171106050209_add_pull_requests_to_task_list.exs
snewcomer: I really like this.

snewcomer (commented):
:+1:

priv/repo/migrations/20171106050209_add_pull_requests_to_task_list.exs
joshsmith: Unfortunately it looks like this will override some of the work that was done in the earlier migration that moved tasks to closed, because this is not checking for whether the issues are closed.

joshsmith (changes requested):

priv/repo/migrations/20171106050209_add_pull_requests_to_task_list.exs
begedin: @JoshSmith You are correct about checking for gi not closed. You could also check for task not closed, it would likely serve the same purpose, since presumably, the task is otherwise in sync with the issue.

On a similar note, my other migration should probably add a clause for archived. I'm not sure what the exact query would be, but it would likely involve a `where: date_add(t.modified_at, 30, "day") > ^Date.utc_today` somewhere in it.

I can't push the change myself due to not being home, but that should help.

joshsmith (approved):

PR #1161, opened 2017-11-06
Fix installation_repositories event to pull in account data

### LINKED ISSUE #1117: Installation repositories event does not pull in some data
# Problem

When we receive the `installation_repositories` event, we are unfortunately not pulling in some data about the repository, including:

- `github_account_id`
- `github_account_login`
- `github_account_avatar_url`
- `github_account_type`

This data _is_ available in the `installation`'s `account` object.

I'm wondering if we should simply defer to the installation's data on this and instead remove the `github_account_` details from the repos themselves.

joshsmith:
# What's in this PR?

Fixes the installation repositories event to pull in GitHub account data.

## References
Fixes #1117

### DIFF
--- a/lib/code_corps/github/event/installation/changeset_builder.ex
+++ b/lib/code_corps/github/event/installation/changeset_builder.ex
@@ -25,7 +25,7 @@ defmodule CodeCorps.GitHub.Event.Installation.ChangesetBuilder do
     github_app_installation
     |> Changeset.change(attrs)
     |> Changeset.put_change(:installed, true)
-    |> inferr_origin()
+    |> infer_origin()
     |> Changeset.unique_constraint(:github_id, name: :github_app_installations_github_id_index)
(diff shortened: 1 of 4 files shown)


lib/code_corps/github/event/installation/changeset_builder.ex
```diff
@@ -25,7 +25,7 @@ defmodule CodeCorps.GitHub.Event.Installation.ChangesetBuilder do
     github_app_installation
     |> Changeset.change(attrs)
     |> Changeset.put_change(:installed, true)
-    |> inferr_origin()
+    |> infer_origin()
```
begedin, line 28: Oops! That one's on me.

begedin (approved):
Looks good. Not sure what my reasoning behind _not_ getting the account data was, but this makes much more sense.

PR #1162, opened 2017-11-06
Add pagination support to GitHub.API.Installation.repositories

### LINKED ISSUE #1108: Implement pagination support into GitHub.API.Installation.repositories
# Problem

I'm not 100% positive on this, but it may be that our `Installation.repositories` endpoint needs to add in pagination support implemented in #1107

joshsmith:
# What's in this PR?

Adds pagination support to `GitHub.API.Installation.repositories`.

Wasn't sure if a test was necessary given that it was failing until I got it working. Don't think we need to do extra testing for this to be paginating.

Any thoughts here?

## References
Fixes #1108

### DIFF
--- a/lib/code_corps/accounts/accounts.ex
+++ b/lib/code_corps/accounts/accounts.ex
@@ -71,7 +71,7 @@ defmodule CodeCorps.Accounts do
     end
   end
 
-  @spec do_update_with_github_user(Usert.t, GithubUser.t) :: {:ok, User.t} | {:error, Changeset.t}
+  @spec do_update_with_github_user(User.t, GithubUser.t) :: {:ok, User.t} | {:error, Changeset.t}
   defp do_update_with_github_user(%User{} = user, %GithubUser{} = github_user) do
     user
     |> Changesets.update_with_github_user_changeset(github_user |> Adapters.User.to_user_attrs())
--- a/lib/code_corps/github/adapters/utils/body_decorator.ex
+++ b/lib/code_corps/github/adapters/utils/body_decorator.ex
@@ -5,6 +5,7 @@ defmodule CodeCorps.GitHub.Adapters.Utils.BodyDecorator do
 
   alias CodeCorps.{
     Comment,
+    GithubIssue,
     Task,
     User,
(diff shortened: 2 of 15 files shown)


joshsmith:
It looks like this is erroring due to the fact that `get_all` may not handle individual errors from the GitHub API well.

For example, the list returned seems to be:

```
[%CodeCorps.GitHub.APIError{documentation_url: nil, errors: nil,
  message: "{\"message\":\"Maximum number of login attempts exceeded. Please try again later.\",\"documentation_url\":\"https://developer.github.com/v3\"}",
  status_code: 404}]
```

And `Installation.repositories` does not indicate that this is an error, but instead tries to pass this off as a list of repos.

begedin:
@JoshSmith Took me a bit to get here, but I managed to write some behavior which should allow us to more easily recover from various potential issues during a paginated request.

There were a bunch of inconsistencies that needed fixing, as well as miscellaneous minor corrections I had to do to get to that point, but overall, this should increase stability.

Just as a minor tip, I found that looking at dialyzer errors, which we have quite a lot of, using `mix dialyzer` was helpful in identifying potential bugs. We should strive to resolve as much of those as we can and keep them to a minimum, but I don't think it should be our focus. Just something we do while we work on a feature or bug.

The commit message should clarify what I did but I'll also copy it here for convenience

- added `API.Errors` namespace, to move other errors into
- added `API.Errors.PaginationError`
- rewrote some specs that were breaking contracts
- standardized `get_all/3` response to match `request/5` response (`{:ok, data}` or `{:error, error}`)
- corrected several type definitions
- added support for recovering from failed individual page requests during `get_all/3`
- added support for recovering from failed initial `head` request during `get_all/3`
- added support for marshalling a failed and successful `head` response (`body` is `""`so poison was failing to decode)
- rewrote `ResultAggregator` docs and argument names so they reflect the fact that it can be used for aggregating any standard response list, not just db commit
- wrote tests to cover new behaviors

joshsmith:
This looks great to me. I can't approve since it's my PR, but will rebase and merge now. Great work!

PR #1163, opened 2017-11-07
Set modified_from to "code_corps" on tasks and comments when updating from codecorps

### LINKED ISSUE #1145: modified_from is not being set on comments and tasks
# Problem

This field should be set on `Comment` and `Task` to either "codecorps" or "github" depending on where the modification originates from. It should be handled in the respective changesets, but it's not currently being done from either origin.

begedin:
# What's in this PR?

This PR makes it so that every time a task or comment is modified from the client application, thus, from CodeCorps, not some other location, the `:modified_from` field on that `Task` or `Comment` respectively is set to `"code_corps"`.

`"code_corps"` was already our field default, so I wen't with that, instead of `"codecorps"` as instructed in the issue.

Since it was our field default, the field was set correctly upon initial creation from the client app. However, a subsequent modification from github, followed by modification from codecorps, would result in `:modified_from` being incorrectly stuck on `"github"`, even though the last source of the modification was in fact the client app.

Note that the `create_changeset` is also used in our seeds. It makes sense as it is, but if we want to seed our db with tasks or comments linked to github, we need to take that into account.

## References
Fixes #1145

### DIFF
--- a/test/lib/code_corps/comment/service_test.exs
+++ b/test/lib/code_corps/comment/service_test.exs
@@ -29,6 +29,11 @@ defmodule CodeCorps.Comment.ServiceTest do
       refute_received({:post, "https://api.github.com/" <> _rest, _body, _headers, _options})
     end
 
+    test "sets modified_from to 'code_corps'" do
+      {:ok, comment} = valid_attrs() |> Comment.Service.create
+      assert comment.modified_from == "code_corps"
+    end
+
     test "returns errored changeset if attributes are invalid" do
       {:error, changeset} = Comment.Service.create(@base_attrs)
       refute changeset.valid?
@@ -90,6 +95,13 @@ defmodule CodeCorps.Comment.ServiceTest do
       refute updated_comment.github_comment_id
 
       refute_received({:patch, "https://api.github.com/" <> _rest, _body, _headers, _options})
+    end
+
+    test "sets modified_from to 'code_corps'" do
+      comment = insert(:comment, modified_from: "github")
+      {:ok, updated_comment} = comment |> Comment.Service.update(@update_attrs)
+
+      assert updated_comment.modified_from == "code_corps"
     end
 
     @preloads [task: [github_repo: :github_app_installation]]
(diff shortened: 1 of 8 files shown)


test/lib/code_corps/comment/service_test.exs
```diff
@@ -92,6 +97,13 @@ defmodule CodeCorps.Comment.ServiceTest do
       refute_received({:patch, "https://api.github.com/" <> _rest, _body, _headers, _options})
     end
 
+    test "sets modified_from to 'code_corps'" do
+      comment = insert(:comment, modified_from: "github")
+      {:ok, updated_comment} = comment |> Comment.Service.update(@update_attrs)
+
+      assert updated_comment.modified_from == "code_corps"
+    end
```
begedin, line 105: A changeset test might seem enough, but I wanted to add a sort of "integration" test to. The service is where the modification gets commited to the db.

test/lib/code_corps/github/sync/issue/task/task_test.exs
```diff
@@ -3,44 +3,59 @@ defmodule CodeCorps.GitHub.Sync.Issue.TaskTest do
...
 
-      task_ids = tasks |> Enum.map(&Map.get(&1, :id))
-      assert task_1.id in task_ids
+      assert existing_task.id in (tasks |> Enum.map(&Map.get(&1, :id)))
+    end
+
+    test "sets task :modified_from to 'github'" do
+      %{github_issue: github_issue, user: user} = setup_test_data()
+      {:ok, tasks} = github_issue |> IssueTaskSyncer.sync_all(user)
+      assert tasks |> Enum.all?(fn task -> task.modified_from == "github" end)
```
begedin, line 58: For the same reason as the service test above, this is a sort of an integration test, even though our sync changeset already has the needed assertions in its own tests.

I did not do the same for comments because of the complexity of the test data there. I'm hoping I'll get a chance to decouple the comment syncer from payloads first.

joshsmith (approved):

PR #1167, opened 2017-11-08
Add github-events endpoint

### LINKED ISSUE #1166: Add /github-events API endpoint
# Problem

We need to add a `/github-events` API endpoint in order to allow the admin UI to work. This should allow only an `index` and `show` and the policy should only allow this for admins.

- Add policy test
- Add policy
- Add view test
- Add view
- Add controller tests for index and show
- Add controller endpoints to router
- Add controller logic

joshsmith:
# What's in this PR?

- Adds show to github-events
- Adds index to github-events
- Adds pagination to index
- Renames controller to correct name
- Adds `admin` to `UserView`
- Tests for everything

## References
Fixes #1166


begedin (approved):
This is, I believe, mergeable, but we should open an issue to split up webhook processing and admin actions into separate controllers. we can probably put webhook processing into a separate namespace, for example.

PR #1179, opened 2017-11-10
Sort GithubEvent list by inserted_at close #1172

### LINKED ISSUE #1172: GithubEventController should order events by most recent insertedAt
The `GithubEventController` right now orders the events payload by oldest events first. We need to reverse the order so we're always seeing the newest events first.

snewcomer:
Fixes #1172

### DIFF
--- a/lib/code_corps_web/controllers/github_event_controller.ex
+++ b/lib/code_corps_web/controllers/github_event_controller.ex
@@ -19,6 +19,7 @@ defmodule CodeCorpsWeb.GithubEventController do
       github_events =
         GithubEvent
         |> Query.id_filter(params)
+        |> Ecto.Query.order_by([asc: :inserted_at])
         |> paginate(params)
 
       conn |> render("index.json-api", data: github_events)
--- a/test/lib/code_corps_web/controllers/github_event_controller_test.exs
+++ b/test/lib/code_corps_web/controllers/github_event_controller_test.exs
@@ -28,13 +28,19 @@ defmodule CodeCorpsWeb.GithubEventControllerTest do
     end
 
     @tag authenticated: :admin
-    test "lists all entries on index", %{conn: conn} do
-      [github_event_1, github_event_2] = insert_pair(:github_event)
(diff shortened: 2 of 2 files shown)


snewcomer:
@begedin could you kick off the ci again?

lib/code_corps_web/controllers/github_event_controller.ex
```diff
@@ -19,6 +19,7 @@ defmodule CodeCorpsWeb.GithubEventController do
       github_events =
         GithubEvent
         |> Query.id_filter(params)
+        |> Query.sort_by_inserted_at()
```
joshsmith: I wonder if it makes more sense to just call `|> order_by([asc: :inserted_at])` here directly?

joshsmith: Technically we would eventually want to wrap this up in a context outside the controller anyway, so it would be more like `GithubEvents.list()` or something in Phoenix 1.3+ style.

test/lib/code_corps_web/controllers/github_event_controller_test.exs
```diff
@@ -28,13 +28,18 @@ defmodule CodeCorpsWeb.GithubEventControllerTest do
...
 
     @tag authenticated: :admin
-    test "lists all entries on index", %{conn: conn} do
-      [github_event_1, github_event_2] = insert_pair(:github_event)
+    test "lists all entries on index by inserted_at", %{conn: conn} do
+      github_event_1 = insert(:github_event, inserted_at: Timex.now())
+      github_event_2 = insert(:github_event, inserted_at: Timex.now() |> Timex.shift(days: 3))
 
-      conn
+      data = conn
```
joshsmith: Since it looks like we do need to do this extra work here, I think the convention is:

```
data =
  conn
  |> my_func()
```

lib/code_corps_web/controllers/github_event_controller.ex
snewcomer: @JoshSmith Should we do it right now and put a method on GithubEvent?

PR #1180, opened 2017-11-11
Handle error in StripeConnectSubscriptionService.find_or_create/1

### LINKED ISSUE #1164: Elixir.MatchError: no match of right hand side value: {:error, %Stripe.APIErrorResponse{code: "expired_card", messag...
https://sentry.io/code-corps/code-corps-api-staging/issues/400925827/

```
Elixir.MatchError: no match of right hand side value: {:error, %Stripe.APIErrorResponse{code: "expired_card", message: "Your card has expired.", status_code: 402, type: "card_error"}}
  File "lib/code_corps/stripe_service/stripe_connect_subscription_service.ex", line 40, in CodeCorps.StripeService.StripeConnectSubscriptionService.find_or_create/1
  File "lib/code_corps_web/controllers/stripe_connect_subscription_controller.ex", line 27, in CodeCorpsWeb.StripeConnectSubscriptionController.create/2
  File "lib/code_corps_web/controllers/stripe_connect_subscription_controller.ex", line 1, in CodeCorpsWeb.StripeConnectSubscriptionController.action/2
  File "lib/code_corps_web/controllers/stripe_connect_subscription_controller.ex", line 1, in CodeCorpsWeb.StripeConnectSubscriptionController.phoenix_controller_pipeline/2
  File "lib/code_corps_web/endpoint.ex", line 1, in CodeCorpsWeb.Endpoint.instrument/4
...
(3 additional frame(s) were not displayed)

(MatchError) no match of right hand side value: {:error, %Stripe.APIErrorResponse{code: "expired_card", message: "Your card has expired.", status_code: 402, type: "card_error"}}
```

landongrindheim:
Handle error case in `StripeConnectSubscriptionService.find_or_create/1`

## References
Fixes #1164

### DIFF
--- a/lib/code_corps/stripe_service/stripe_connect_subscription_service.ex
+++ b/lib/code_corps/stripe_service/stripe_connect_subscription_service.ex
@@ -35,10 +35,9 @@ defmodule CodeCorps.StripeService.StripeConnectSubscriptionService do
     with {:ok, %Project{} = project} <- get_project_with_preloads(project_id),
          {:ok, %Project{}} <- ProjectSubscribable.validate(project),
          {:ok, %User{} = user} <- get_user_with_preloads(user_id),
-         {:ok, %User{}} <- UserCanSubscribe.validate(user)
+         {:ok, %User{}} <- UserCanSubscribe.validate(user),
+         {:ok, %StripeConnectSubscription{} = subscription} <- do_find_or_create(project, user, attributes)
     do
-      {:ok, %StripeConnectSubscription{} = subscription} = do_find_or_create(project, user, attributes)
-
       ProjectService.update_project_totals(project)
       DonationGoalsService.update_project_goals(project)
 


joshsmith (approved):

joshsmith:
Awesome, thanks for this!

I think we do need to provide some sort of feedback to the user in this case, but that's a far more complex thing than simply handling the error here at some level.

joshsmith:
I wonder if it's possible for us to have a failing test here, too, to see what the behavior was like before and how it looks now in practice?

landongrindheim:
That's a good idea. Not exactly sure how to replicate the state found in #1164. Any pointers on getting there?

joshsmith:
I think in the `StripeConnectSubscriptionServiceTest` you would want to insert a `stripe_platform_card` that has an expired date. That should be enough to trigger it.

As for what's presented to the user, we just need to be sure that the API is sending errors down that are human-readable. This would be a good bet to cover in a controller test, but it might be out of scope for this PR. Would maybe consider opening a separate issue for that?

landongrindheim:
I've attempted triggering the error with the following setup:

```elixir
      insert(:stripe_connect_account, organization: organization)

      stripe_connect_plan = insert(:stripe_connect_plan)
      project = insert(:project, stripe_connect_plan: stripe_connect_plan, organization: organization)

      user_with_expired_card = insert(:user)
      insert(:stripe_platform_customer, user: user_with_expired_card)
      insert(:stripe_platform_card, user: user_with_expired_card, exp_month: 1, exp_year: 2000)

      {:error, error} =
        StripeConnectSubscriptionService.find_or_create(
          %{
            "project_id" => project.id,
            "user_id" => user_with_expired_card.id,
            "quantity" => 200
          }
        )
```

However, I get a `MatchError` because `StripeConnectSubscription.find_or_create/1` is returning an `{:ok, Model}` tuple.

I see other places where the expiration date is in the past, most notably in `StripeTesting.card`, but haven't seen any explicit handling of errors returned from Stripe about the card being expired.

joshsmith:
@landongrindheim my best guess is that this isn't triggering anything because we're mocking out the Stripe API response as being successful. And I'm not sure that we have a particularly great ability right now to inject a different mock.

In `config/test.exs` we set:

```
config :code_corps, :stripe, CodeCorps.StripeTesting
```

@begedin do you remember precisely how we're constructing the mocked data?

landongrindheim:
@begedin Do you have any suggestions for triggering/mocking a failure from the test Stripe API?

joshsmith:
So I think I have no good answers here.

Right now the way Stripe testing works is that we do an `@api Application.get_env(:code_corps, :stripe)` which uses our Stripe test mode: `config :code_corps, :stripe, CodeCorps.StripeTesting`

This mocks out every call to `stripity_stripe`, so in this case an `@api.Card.create` is actually calling `CodeCorps.StripeTesting.Card.create`.

Without manually overriding these in `StripeTesting`, there wouldn't be any way for a `create` or `retrieve` or `update` (or any other operation) to get the specialized mocked data we want to use, in this case returning a `%Stripe.APIErrorResponse{}` rather than ` %Stripe.Card{}`.

I think you should proceed with your PR by just fixing what we know to be fixable here and ignoring the fact that we don't even have a failure case to test against. I can't think of any solution here that isn't fundamentally more effort than just getting `stripity_stripe` 2.0 up to speed and relying in the future on a solution that lets us do more explicit testing by calling `stripe-mock` or using some other mocking pattern that takes it into account.

lib/code_corps/stripe_service/stripe_connect_subscription_service.ex
```diff
@@ -37,12 +37,14 @@ defmodule CodeCorps.StripeService.StripeConnectSubscriptionService do
...
-
-      {:ok, subscription}
+      case do_find_or_create(project, user, attributes) do
+        {:ok, %StripeConnectSubscription{} = subscription} ->
+          ProjectService.update_project_totals(project)
+          DonationGoalsService.update_project_goals(project)
+          {:ok, subscription}
+        {:error, error} ->
+          {:error, error}
+      end
```
joshsmith: I think the simpler solution here is to not do a `case` statement but instead simply move the `do_find_or_create` logic as it was up into the `with` and only call the other services and return `{:ok, subscription}` in the `do`, letting the failure of `Stripe.APIErrorResponse` be dealt with by the `else`.

joshsmith (changes requested):
And now for the actual code at hand, as we discussed earlier today in Slack:

landongrindheim:
@JoshSmith Updated with the changes you suggested. Thanks for the feedback!

joshsmith (approved):

PR #1184, opened 2017-11-12
Add processor that switches between sync/async

joshsmith:
# What's in this PR?

Attempting to fix some of the following errors:

```
2017-11-12T08:00:11.048315Z [error] Task #PID<0.510.0> started from #PID<0.508.0> terminating
** (DBConnection.OwnershipError) cannot find ownership process for #PID<0.510.0>.
```

There will now be some failing tests due to areas where the logic won't work synchronously.


joshsmith:
@begedin as a first go, how does this look to you compared to the alternative?

I _will_ fix those tests.

Also wow, test coverage? What?

joshsmith:
@begedin I merged this because it's such a painful state for me right now. If you have some changes you'd like made, could you open a followup issue for them? Thanks!

PR #1185, opened 2017-11-13
Cleaned up unsupported and ignored event handling

begedin:
This PR deals with us storing partially unsupported (we support the type, but not the action) events as errored. 

Instead, they are now stored as "unsupported".

Additionally, this PR adds `{"pull_request", "syncronized"}` to the list of unsupported events.

While all this works and removes some noise from our _event log_, I should not that functionally, these partially supported events are no different than events we outright ignore. Once and if we add full support, we will not be digging through our event log to restore them, se we should either

- store all events and mark those unsupported as unsupported
- store only fully supported events

The first gives us the benefit of tracking down events we may have missed potentially. The second gives us les work and less things to store, so I'm not sure of our approach (mainly due to not being sure how much we would end up storing).


lib/code_corps/github/event/installation/installation.ex
```diff
@@ -16,15 +16,13 @@ defmodule CodeCorps.GitHub.Event.Installation do
   alias Ecto.{Changeset, Multi}
 
   @type outcome :: {:ok, GithubAppInstallation.t} |
-                    {:error, :not_yet_implemented} |
-                    {:error, :unexpected_action} |
```
begedin, line 20: `:not_yet_implemented` and `:unexpected_action` are no longer reachable with the other changes, so I've removed them for the benefit of simplifying. They would technically still be reachable through a unit test, but not through any sort of integration test.

lib/code_corps/github/webhook/handler.ex
```diff
@@ -11,46 +11,59 @@ defmodule CodeCorps.GitHub.Webhook.Handler do
...
-         {:ok, %GithubEvent{} = event} <- params |> create_event(),
-         {:ok, %GithubEvent{status: "processing"} = event} <- event |> Event.start_processing
-    do
-      payload |> do_handle(type) |> Event.stop_processing(event)
+  @spec handle_supported(String.t, String.t, map) :: {:ok, GithubEvent.t}
+  def handle_supported(type, id, %{} = payload) do
+    with {:ok, %GithubEvent{} = event} <- type |> build_params(id, "unprocessed", payload) |> create_event() do
+      payload |> apply_handler(type) |> Event.stop_processing(event)
     end
   end
```
begedin, line 31: Removed what's basically an `if` from this module and simply _flattened it out_ from the POV of the controller. Keeps things simpler, allows us to test more directly, and peels off an abstraction layer.

lib/code_corps/github/webhook/processor.ex
```diff
@@ -1,22 +0,0 @@
...
-  Receives the event JSON as the only parameter.
-
-  Returns `{:ok, pid}`
-  """
-  def process_async(type, id, payload) do
-    Processor.process(fn -> process(type, id, payload) end)
-  end
-
-  defdelegate process(type, id, payload), to: Handler, as: :handle
-end
```
begedin, line 22: It's basically just a delegate, so no real need for it.

lib/code_corps_web/controllers/github_event_controller.ex
```diff
@@ -36,23 +41,31 @@ defmodule CodeCorpsWeb.GithubEventController do
...
+  def create(%Conn{} = conn, %{} = payload) do
+    type = conn |> get_event_type
+    delivery_id = conn |> get_delivery_id()
+    action = payload |> Map.get("action", "")
 
-    case event_type |> EventSupport.status do
+    case type |> EventSupport.status(action) do
       :supported ->
-        Processor.process_async(event_type, conn |> get_delivery_id, event_payload)
+        Processor.process(fn -> Handler.handle_supported(type, delivery_id, payload) end)
```
begedin, line 52: This is the `CodeCorps.Processor` now, calling `CodeCorps.GitHub.Webhook.Handler` directly

joshsmith (approved):

joshsmith (changes requested):
Actually this is not good to go without a migration that takes all `GithubEvent` records where the `failure_reason` is `not_fully_implemented` or `unexpected_action` and nullifies the `failure_reason` and sets `status` to `unsupported`.

lib/code_corps/github/webhook/event_support.ex
```diff
@@ -3,23 +3,66 @@ defmodule CodeCorps.GitHub.Webhook.EventSupport do
...
+    {"issues", "unlabeled"},
+    {"issues", "milestoned"},
+    {"issues", "demilestoned"},
+    {"pull_request", "assigned"},
+    {"pull_request", "unassigned"},
+    {"pull_request", "review_requested"},
+    {"pull_request", "review_request_removed"},
+    {"pull_request", "labeled"},
+    {"pull_request", "unlabeled"},
+    {"pull_request", "synchronized"},
```
joshsmith: This is actually called `synchronize` and not `synchronized`.

joshsmith (changes requested):

joshsmith (dismissed):

joshsmith:
There were actually a number of things amiss here, including most importantly that the `GithubEventController` was not responding to unsupported events correctly and our tests were giving false confidence that this was not an issue.

PR #1186, opened 2017-11-13
Fix webhook validation events

begedin:
Beginnings of changes to deal with validation errors during comment/task sync

I went deep into this and think I have a solution, but there's still work to be done. Here are my findings

# sync up

## current approach

x `Tasks` -> 1 `GithubIssue` -> API

if a `Task` changes we
- update the `Task`
- update the associated `GithubIssue` if any
- propagate `GithubIssue` change to API
- right now, do nothing else

This works. We get a webhook back, and if there are other projects linked to same `GithubRepo`, `Task` records get created on those.

However, if Github now goes down somehow, we don't get a webhook and things go out of sync. Not with just one project we currently have, but techincally, with multiple projects.

## proposed change

What I think we should do is

- (_this PR_) propagate changes from `GithubIssue` to other tasks immediately, regardless of API push outcome
- (_separate PR_) add a mechanism to retry failed API push

This allows us to keep sync working locally, across linked `Task` records, and then sync to github when it goes back up.

# sync down

## current approach

if an issue changes on github, we get webhook and right now
- find associated github issue, update it always
- find associated tasks, update those, maybe that fails due to `:modified_at`
- since it's all a transaction right now, everything fails, but we talked about ditching transactions
- once we ditch transactions, suddenly we're in a situation where
  - `GithubIssue` matches the github API
  - some linked `Task` records match `GithubIssue`, some don't
  - this is a nightmare to deal with and I don't think there's an objectively correct method of resolving

## proposed change

what, to me, makes more sense is

- the github issue is our gateway to sync in the **down** direction as well
 
- if github sends webhook about issue update we check `:modified_at` **on** `GithubIssue`, **not** `Task`
  - if `:modified_at` check passed, update `GithubIssue` and associated `Task` records
  - if `:modified_at` check failed, reject `GithubIssue` update, never even get to `Task` records
  
The same approach applies to `Comment` and `GithubComment` respectively
  
# overall solution

- we merge the other PR
- we add a `:modified_at` to `GithubIssue`
- i think we can remove `:modified_at` from `Task`, or, at least, repurpose it if it's still needed for timestamp/UI purposes
- we add time validator behavior to `GithubIssue`, except I'm not sure we want it as a changeset validation. It certainly cann work that way - we would just have to catch that outcome and intepret it in a way that would either ignore the event, mark it as processed, or possibly mark it as a new state - `:ignored_out_of_sync`
- do the same to `Comment`/`GithubComment`

# sync up cases

Just so we can discuss them and potentially find any oversights...

## Github down while creating `Task`

Simplest thing we could do is 

- create `Task`
- create `GithubIssue`
- create other `Tasks`
- API request fails
- `GithubIssue` exists without ID, we can create it when github goes back up

## Github down while updating `Task`

- update `Task`
- update `GithubIssue`
- update other `Tasks`
- API request fails
- `GithubIssue` marked as failed to sync, we can sync later

## Github down while creating `Comment`

- create `Comment`
- create `GithubComment`
- create other `Comments`
- API request fails
- `GithubComment` exists without ID, we can create it when github goes back up

## Github down while updating `Comment`

- update `Comment`
- update `GithubComment`
- update other `Comments`
- API request fails
- `GithubComment` marked as failed to sync, we can sync later

Really, update and create are the same. In both cases, the _gateway_ record is marked as failed to sync, except in the case of create, it also has no `github_id`. When our _sync later_ module retries the sync, it simply does a post or patch bashed on id value.

# What this PR does

...up to this point is to decouple some things with comment syncing similarly to how it was done with task/issue syncing in previous PRs.

The changeset and several other steps no longer rely on payloads and instead comments rely on github comment records for data, etc.


joshsmith:
I've since deleted the original commit and changed the base and rebased.

joshsmith:
I'm sitting on this some until we can talk through the path forward.

@begedin it would also be helpful if you pointed out which things would likely remain regardless of how we handle sync going forward, i.e. things that are just cleanup.

begedin:
@joshsmith This is obsolete now, so it can be safely closed.

PR #1187, opened 2017-11-13
Replace TaskList with jsonapi

snewcomer:
This is a WIP.  A few PR's to jsonapi may be in order.  Need to investigate more.

Progress on: #1139


config/config.exs
```diff
@@ -84,6 +84,9 @@ config :sentry,
 
 config :code_corps, :processor, CodeCorps.Processor.Async
 
+config :jsonapi, 
+  remove_links: true
```
snewcomer: `underscore_to_dash` is giving me this error.  Perhaps a PR.

```
** (Protocol.UndefinedError) protocol Enumerable not implemented for #DateTime<2017-11-13 15:38:42Z>. This protocol is implemented for: DBConnection.PrepareStream, DBConnection.Stream, 
Date.Range, Ecto.Adapters.SQL.Stream, File.Stream, Function, GenEvent.Stream, HashDict, HashSet, IO.Stream, List, Map, MapSet, Postgrex.Stream, Range, Scrivener.Page, Stream, Timex.Interval
```

lib/code_corps_web/views/task_list_view.ex
```diff
@@ -1,11 +1,21 @@
...
 
-  has_one :project, type: "project", field: :project_id
+  def render("index.json-api", %{data: task_list, conn: conn}) do
+    __MODULE__.index(task_list, conn, nil)
+  end
 
-  has_many :tasks, serializer: CodeCorpsWeb.TaskView, identifiers: :always
+  def render("show.json-api", %{data: task_list, conn: conn, params: params}) do
+    __MODULE__.show(task_list, conn, params)
+  end
```
snewcomer: Would rather just call the show/index action directly from the controller, but was getting failing controller tests around `expected connection to have a response but no response was set/sent.  Please verify that you assign to "conn" after a request`.  Will have to investigate more.

test/lib/code_corps_web/views/task_list_view_test.exs
```diff
@@ -7,7 +7,7 @@ defmodule CodeCorpsWeb.TaskListViewTest do
     task = insert(:task, order: 1000, task_list: task_list)
 
     task_list = CodeCorpsWeb.TaskListController.preload(task_list)
-    rendered_json =  render(CodeCorpsWeb.TaskListView, "show.json-api", data: task_list)
+    rendered_json =  render(CodeCorpsWeb.TaskListView, "show.json-api", %{data: task_list, conn: %Plug.Conn{}, params: task_list.id})
 
     expected_json = %{
```
snewcomer: @begedin this is failing.  I returns a map with atom keys.  Should the json be a map with string keys or doesn't it matter?

begedin: We have keys such as `"pull-requests"`, so I'm not sure they can be atoms. Is this configurable in jsonapi?

begedin: I see you said `underscore_to_dash: true,` is giving you errors above. We need that to. Our Ember client is expecting dashes.

config/config.exs
begedin: Looks like problems with serializing a date. We'll have to look into it. Any open or closed issues related to date serialization?

begedin: Narrowing down to which field is creating the problem would likely help. Maybe disable one attribute at a time to see what happens?

begedin (commented):
These comments are pending for a couple of days now. Not sure how I missed it. Sorry about that. I hope they help.

snewcomer:
@begedin @joshsmith So I got a good working version for task_list.  However, it kind of creates a snowball effect b/c ProjectView / TaskView needs to be changed.  Then those relationship views need to be updated.

I wanted to know your thoughts on tackling this.  It's a big refactor so just wanted to ping you guys to make sure its a ok.  I can try to get it done today...

snewcomer:
task_list

Here is the latest w/o includes...

snewcomer:
task_list_include

and about net w/ the includes flag

snewcomer:
@joshsmith So here is where I am at.  I separated out the project and task views to their own files specific for the task_list view so to avoid a huge refactor.  These relationships are now `included` in the view as well.  

The only piece of housekeeping is ensuring the custom attribute methods that are defined on the view can be included in the attributes...which may be a PR to jsonapi.

snewcomer:
@begedin @joshsmith Just to add to what the goal is...It would be great to use the `include` key, whether it is jsonapi or ja-serializer.  So maybe I close this and start a PR to use the `include` key w/ ja-serializer....or I keep going with this PR.  Lmk your thoughts.

snewcomer:
Going to close this in favor of #1302 for right now.

PR #1188, opened 2017-11-13
Improve pwd reset security

### LINKED ISSUE #961: Password Reset Error handling
# Problem

Phoenix.Token.verify can return `{:error, :expired}` or `{:error, :invalid}`.  Moreover `Guardian.encode_and_sign` can return an `{:error, any}`.  Another error may be `{:error, :missing}` as well.

Thus we need to update the reset controller and associated tests to handle these error cases.

snewcomer:
# What's in this PR?

- [x] Remove AuthToken model after reset occurs so can't reset pwd multiple times.
- [x] delete current session if hacker is currently impersonating a session.
- [x] ensure proper errors are handled.

Fixes #961

### DIFF
--- a/lib/code_corps_web/controllers/password_controller.ex
+++ b/lib/code_corps_web/controllers/password_controller.ex
@@ -9,6 +9,7 @@ defmodule CodeCorpsWeb.PasswordController do
   """
   def forgot_password(conn, %{"email" => email}) do
     ForgotPasswordService.forgot_password(email)
+    conn = Guardian.Plug.sign_out(conn, :default)
     conn |> put_status(:ok) |> render("show.json", email: email)
   end
 
--- a/lib/code_corps_web/controllers/password_reset_controller.ex
+++ b/lib/code_corps_web/controllers/password_reset_controller.ex
@@ -18,9 +18,10 @@ defmodule CodeCorpsWeb.PasswordResetController do
     - If no, a `422` response will return the error.
   """
   def reset_password(conn, %{"token" => reset_token, "password" => _password, "password_confirmation" => _password_confirmation} = params) do
-    with %AuthToken{user: user} <- AuthToken |> Repo.get_by(%{ value: reset_token }) |> Repo.preload(:user),
-         {:ok, _} <- Phoenix.Token.verify(CodeCorpsWeb.Endpoint, "user", reset_token, max_age: 3600),
+    with %AuthToken{user: user} = auth_token <- AuthToken |> Repo.get_by(%{ value: reset_token }) |> Repo.preload(:user),
+         {:ok, _} <- Phoenix.Token.verify(conn, "user", reset_token, max_age: Application.get_env(:code_corps, :password_reset_timeout)),
          {:ok, updated_user} <- user |> User.reset_password_changeset(params) |> Repo.update,
+         {:ok, _auth_token} <- auth_token |> Repo.delete,
          {:ok, auth_token, _claims} = updated_user |> Guardian.encode_and_sign(:token)
     do
       conn
@@ -29,6 +30,7 @@ defmodule CodeCorpsWeb.PasswordResetController do
       |> render("show.json", token: auth_token, user_id: updated_user.id, email: updated_user.email)
     else
       {:error, %Changeset{} = changeset} -> conn |> put_status(422) |> render(CodeCorpsWeb.ErrorView, :errors, data: changeset)
+      {:error, _} -> conn |> put_status(:not_found) |> render(CodeCorpsWeb.ErrorView, "404.json")
       nil -> conn |> put_status(:not_found) |> render(CodeCorpsWeb.ErrorView, "404.json")
     end
   end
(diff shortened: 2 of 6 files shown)


lib/code_corps_web/controllers/password_reset_controller.ex
```diff
@@ -29,6 +30,7 @@ defmodule CodeCorpsWeb.PasswordResetController do
       |> render("show.json", token: auth_token, user_id: updated_user.id, email: updated_user.email)
     else
       {:error, %Changeset{} = changeset} -> conn |> put_status(422) |> render(CodeCorpsWeb.ErrorView, :errors, data: changeset)
+      {:error, _} -> conn |> put_status(:not_found) |> render(CodeCorpsWeb.ErrorView, "404.json")
```
snewcomer, line 33: Having trouble testing this (this is why coverage dropped).  Only works if token was expired or perhaps the `secret_key_base` was different.

lib/code_corps_web/controllers/password_controller.ex
```diff
@@ -9,6 +9,7 @@ defmodule CodeCorpsWeb.PasswordController do
   """
   def forgot_password(conn, %{"email" => email}) do
     ForgotPasswordService.forgot_password(email)
+    conn = Guardian.Plug.sign_out(conn)
```
snewcomer: @begedin having trouble testing this.  Looking in the `conn`, I see differences in the `password_controller_test` when this line is here, but haven't figured out how to verify.

lib/code_corps_web/controllers/password_reset_controller.ex
joshsmith: Could you change the `env` to change the `secret_key_base` and reset it? Or manually set the token to be expired?

lib/code_corps_web/controllers/password_controller.ex
joshsmith: Maybe check to see how `Guardian` tests? https://github.com/ueberauth/guardian/blob/d65227145f72b290106c06cecbe42728fbf05fe2/test/guardian/plug_test.exs#L255

lib/code_corps_web/controllers/password_reset_controller.ex
snewcomer: @JoshSmith Yeah that is a good idea.  I guess I just don't know how to `put` this environmental variable.  I don't think this works.
```
    secret_key = Application.get_env(:guardian, Guardian)[:secret_key]
    Application.put_env(:guardian, :secret_key, "1234")
```

snewcomer: Anyways, I configured the password_reset_timeout in the config file and that work!  Thanks for the help!

lib/code_corps_web/controllers/password_controller.ex
snewcomer: So looking at this a second time, I am not seeing a sol'n yet.

`guardian_default_jwt` is still on the `conn` after posting to the forgot password controller. Hmmm.

begedin: Is `Guardian.Plug.authenticated?(conn)` behaving correctly? As in, returns `true` on the "before" conn, `false` on the "after"? 

If that is the case, instead of worrying about the internals, I'd just use that.

snewcomer: password_controller.ex
```
    IO.inspect Guardian.Plug.authenticated?(conn)
    conn = Guardian.Plug.sign_out(conn)
    IO.inspect Guardian.Plug.authenticated?(conn)
```

Both return `true`.  So I'm thinking `sign_out` isn't doing what I think.

snewcomer: Looks like Guardian is changing under the hood as well with some breaking changes: https://github.com/ueberauth/guardian/blob/f397888943629c487d0b6f9ee1ba6ec5d104fa8b/upgrade_guides/0.14.to.1.0.md

snewcomer: Ah figured it out.  Did some debugging and got down to the issue.  @begedin @JoshSmith pending your review, this PR should be g2g!

snewcomer:
@JoshSmith just pinging this PR again!  A review when you got time would be cool!

test/lib/code_corps_web/controllers/password_reset_controller_test.exs
```diff
@@ -29,7 +36,21 @@ defmodule CodeCorpsWeb.PasswordResetControllerTest do
...
+    timeout = Application.get_env(:code_corps, :password_reset_timeout)
+    Application.put_env(:code_corps, :password_reset_timeout, 0)
+
+    current_user = insert(:user)
+    {:ok, auth_token} = AuthToken.changeset(%AuthToken{}, current_user) |> Repo.insert
+    attrs = %{"token" => auth_token.value, "password" => "123456", "password_confirmation" => "123456"}
+    conn = post conn, password_reset_path(conn, :reset_password), attrs
+    assert json_response(conn, 404)
+
+    Application.put_env(:code_corps, :password_reset_timeout, timeout)
```
joshsmith: There's a new test helper for changing the env you should be able to use now.

snewcomer: @JoshSmith Fixed.  Nice I like that helper!

joshsmith (dismissed):
Looks good! Can you rebase and squash to a single commit then merge?

snewcomer:
@JoshSmith squashed

joshsmith (approved):

PR #1190, opened 2017-11-14
Restrict GithubRepo to connect to single project

joshsmith:
# What's in this PR?

Restricts a `GithubRepo` to have a single `Project` via uniqueness on `github_repo_id` on `ProjectGithubRepo`.

Probably needs some work and definitely needs a thorough eye.


lib/code_corps/github/sync/sync.ex
```diff
@@ -26,7 +26,7 @@ defmodule CodeCorps.GitHub.Sync do
                  | {:error, :validating_github_issue}
                  | {:error, :validating_github_comment}
                  | {:error, :validating_user}
-                 | {:error, :validating_tasks}
+                 | {:error, :validating_task}
                  | {:error, :validating_comments}
```
begedin: We're gonna be `:validating_comment` with this to, or am I mistaken?

joshsmith: I'm not quite positive. Part of why I need a thorough poring over this.

begedin:
@JoshSmith 

I got this to a mergeable state and tried to clean up and unravel as much as I could, but did not get further than that and there's still work we could do to make it easier to manage.

## Changes

- merged in changes from #1186, namely the decoupling of modules which can rely on `GithubX` records from payloads
- rewrote parts as much as I could to indicate that now `GithubComment` is 1 to 1 with `Comment`, but there are still cases which could be improved
- removed some abstractions that were confusing the hell out of me. Again, still more could be done. I think we're nesting our sync modules unnecessarily deeply. 
- rewrote parts of code to also indicate `GithubIssue` has one `Task`. You forgot to make that change
- overall cleanup and simplification as much as possible
- rewriting of several tests to account for the new changes
- removed some out of date todos
- removal of unused functions, etc.

## Further unraveling

Even our module namespacing indicates that we're probably nesting them far too much. I think we should flatten this out as much as possible.

For example, we have `Sync.Comment`, `Sync.Comment.Comment`, and `Sync.Comment.Issue``

`Sync.Comment` is probably an unnecessary layer of abstraction. Our main sync functions in `GitHub.Sync` would be made clearer if they were calling the deeply nested stuff directly. In most cases, the middle layer simply assembles a couple of steps from the deeper layers into a multi.

While on the topic of multis, it would be much easier for us to decouple from transactions later if `GitHub.Sync` were the one that deals with multis, while the deeper modules all just use simple functions. It would also point us into the direction of further decoupling.

## Further issues

This still does not effectively deal with validation errors stopping the sync process. We should consider our options here. It may still make sense to move the timestamp check to the `GithubX` record. 

It would further make sense to not perform it as an `Ecto.Changeset` validation and instead a separate `pre-sync` validation step. That would make it easier to switch it to our queue later.

lib/code_corps/github/sync/sync.ex
```diff
@@ -38,10 +38,11 @@ defmodule CodeCorps.GitHub.Sync do
 
   [https://developer.github.com/v3/activity/events/types/#issuesevent](https://developer.github.com/v3/activity/events/types/#issuesevent)
   """
-  def issue_event(%{"issue" => issue} = payload) do
+  @spec issue_event(map) :: outcome
+  def issue_event(%{"issue" => issue_payload} = payload) do
     Multi.new
-    |> Multi.merge(__MODULE__, :find_repo, [payload])
```
begedin, line 43: We had a `find_repo/2` which was made public explicitly so we could use `Multi.merge/4`. That indicates to me we should not use it and instead use `Multi.merge/2`. It reduces code fragmentation, and makes this function easier to understand.

Same for other similar cases below.

lib/code_corps/github/sync/sync.ex
```diff
@@ -258,11 +249,11 @@ defmodule CodeCorps.GitHub.Sync do
   end
```
begedin: I'm not sure this abstraction is a net gain. My vote would be to remove it. The DRYness we get is not worth the loss in information and clarity.

test/support/factories.ex
```diff
@@ -41,7 +41,7 @@ defmodule CodeCorps.Factories do
 
   def github_comment_factory do
     %CodeCorps.GithubComment{
-      body: "I love elixir!",
+      body: sequence(:body, &"I love elixir with GithubComment #{&1}"),
```
begedin, line 44: The static value was identical to `CodeCorps.Comment`, so related changesets would detect changes.

lib/code_corps/github/adapters/comment.ex
```diff
@@ -38,14 +33,15 @@ defmodule CodeCorps.GitHub.Adapters.Comment do
...
   @doc ~S"""
-  Converts a `GithubComment` record into attributes with the same keys as the
-  GitHub API's Issue Comment
+  Converts a GitHub issue comment payload, or a `CodeCorps.GithubComment` into a
+  set of attributes suitable for creating or updating a `CodeCorps.Comment`
   """
-  @spec to_comment_attrs(GithubComment.t) :: map
-  def to_comment_attrs(%GithubComment{} = github_comment) do
+  @spec to_comment(map | GithubComment.t) :: map
+  def to_comment(%GithubComment{} = github_comment) do
```
joshsmith: The `@doc` and `@spec` do not look accurate.

lib/code_corps/github/sync/comment/comment/changeset.ex
```diff
@@ -7,59 +7,30 @@ defmodule CodeCorps.GitHub.Sync.Comment.Comment.Changeset do
...
-    comment |> update_changeset(attrs)
-  end
-
-  @doc ~S"""
-  Constructs a changeset for syncing a task from a `GithubComment` record.
+  The function detects if the `CodeCorps.Comment` is to be inserted or updated
+  and acts accordingly.
   """
   @spec build_changeset(Comment.t, GithubComment.t, Task.t, User.t) :: Changeset.t
   def build_changeset(
```
joshsmith, line 28: I thought our original intent here was to break this out into a `create_changeset` and `update_changeset` to be more explicit. It looks like we've gone the opposite direction some (although we do have the ability to split here on the `build_`).

lib/code_corps/github/sync/comment/comment/comment.ex
```diff
@@ -23,62 +24,58 @@ defmodule CodeCorps.GitHub.Sync.Comment.Comment do
...
 
   @doc ~S"""
-  Creates or updates `Comment` records for the specified `ProjectGithubRepo`.
+  Creates or updates `CodeCorps.Comment` records for the specified
+  `CodeCorps.ProjectGithubRepo`.
 
-  For each `GithubComment` record that relates to the `GithubRepo` for a given
-  `ProjectGithubRepo`:
+  For each `CodeCorps.GithubComment` record that relates to the
+  `CodeCorps.GithubRepo` for a given`CodeCorps.ProjectGithubRepo`:
```
joshsmith, line 54: Missing space.

lib/code_corps/github/sync/comment/comment/comment.ex
```diff
@@ -23,62 +24,58 @@ defmodule CodeCorps.GitHub.Sync.Comment.Comment do
...
 
-  - Find the related `Task` record
-  - Create or update the `Comment` record
-  - Associate the `Comment` record with the `User` that relates to the
-    `GithubUser` for the `GithubComment`
+  - Find the related `CodeCorps.Task` record
+  - Create or update the related `CodeCorps.Comment` record
+  - Associate the `CodeCorps.Comment` record with the `CodeCorps.User` that
+    relates to the `CodeCorps.GithubUser` for the `CodeCorps.GithubComment`
```
joshsmith, line 59: Can we get a handle on where we want to use the `alias`-ed terms and where we don't? I'm fairly sure we can used the `alias`-ed ones. It reads much more easily when that's the case, and there's no reason why if, in code, we can understand the `alias`-ed version, we shouldn't use it in documentation, too.

lib/code_corps/github/sync/comment/comment/comment.ex
```diff
@@ -23,62 +24,58 @@ defmodule CodeCorps.GitHub.Sync.Comment.Comment do
...
+  payload, associated to the specified `CodeCorps.GithubComment`,
+  `CodeCorps.Task` and `CodeCorps.User`
   """
-  @spec sync_all(list(Task.t), GithubComment.t, User.t, map) :: outcome
-  def sync_all(tasks, %GithubComment{} = github_comment, %User{} = user, %{} = payload) do
-    tasks
-    |> Enum.map(&sync(&1, github_comment, user, payload))
-    |> ResultAggregator.aggregate
+  @spec sync_task(Task.t, GithubComment.t, User.t) :: outcome
+  def sync_task(%Task{} = task, %GithubComment{} = github_comment, %User{} = user) do
```
joshsmith: This is probably misnamed now. We're syncing a comment for the task.

lib/code_corps/github/sync/comment/comment/comment.ex
```diff
@@ -23,62 +24,58 @@ defmodule CodeCorps.GitHub.Sync.Comment.Comment do
...
+    %ProjectGithubRepo{github_repo: %{github_comments: github_comments}} =
+      project_github_repo |> Repo.preload(preloads)
 
     github_comments
-    |> Enum.map(&find_or_create_comment(&1, project_github_repo))
+    |> Enum.map(fn %GithubComment{github_user: %GithubUser{user: %User{} = user}} = github_comment ->
+      github_comment
+      |> find_task(project_github_repo)
+      |> sync_task(github_comment, user)
+    end )
```
joshsmith: Extra space.

lib/code_corps/github/sync/sync.ex
joshsmith: I'm guessing you mean `Multi.run/2`?

lib/code_corps/github/sync/sync.ex
```diff
@@ -125,10 +135,9 @@ defmodule CodeCorps.GitHub.Sync do
 
   @count_fields [:syncing_comments_count, :syncing_issues_count, :syncing_pull_requests_count]
 
-  @doc ~S"""
-  Fetches the optional fields (like counter cache fields) for tracking the sync
-  state.
-  """
+
+  # Fetches the optional fields (like counter cache fields) for tracking the
+  # sync state.
```
joshsmith: Good change here. Was bothering me.

lib/code_corps/github/sync/sync.ex
```diff
@@ -274,7 +265,7 @@ defmodule CodeCorps.GitHub.Sync do
   defp marshall_result({:error, :comment_user, :multiple_users, _steps}), do: {:error, :multiple_comment_users_match}
   defp marshall_result({:error, :issue_user, %Ecto.Changeset{}, _steps}), do: {:error, :validating_user}
   defp marshall_result({:error, :issue_user, :multiple_users, _steps}), do: {:error, :multiple_issue_users_match}
-  defp marshall_result({:error, :comments, {_comments, _errors}, _steps}), do: {:error, :validating_comments}
-  defp marshall_result({:error, :tasks, {_tasks, _errors}, _steps}), do: {:error, :validating_tasks}
+  defp marshall_result({:error, :comment, {_comments, _errors}, _steps}), do: {:error, :validating_comment}
+  defp marshall_result({:error, :task, %Ecto.Changeset{}, _steps}), do: {:error, :validating_task}
```
joshsmith, line 274: Now that we have the specific `Task` here, I wonder if we should actually do something with the error changeset since we can actually use this to construct a better audit trail.

lib/code_corps/model/github_issue.ex
```diff
@@ -25,7 +25,7 @@ defmodule CodeCorps.GithubIssue do
     belongs_to :github_user, CodeCorps.GithubUser
 
     has_many :github_comments, CodeCorps.GithubComment
-    has_many :tasks, CodeCorps.Task
+    has_one :task, CodeCorps.Task
```
joshsmith, line 28: Do we have an indication of everything that will need changed in our client? We're going to need to update them side-by-side.

joshsmith (commented):
Some notes/questions I have.

lib/code_corps/github/sync/sync.ex
begedin: Yes. Sorry.

lib/code_corps/github/adapters/comment.ex
begedin: Remains of a previous commit. Sorry, did not notice.

lib/code_corps/github/sync/comment/comment/changeset.ex
begedin: It was my intention to facilitate splitting on the build, yes. Really, the goal is to remove an abstraction that gives us almost no benefit. 

complete split > elimination of a private function because the public function is a near-delegate > what we had

from my pov, at least.

lib/code_corps/github/sync/comment/comment/comment.ex
begedin: I'm ok with renaming.

lib/code_corps/github/sync/comment/comment/comment.ex
begedin: Did you check how the docs generate? As I said in slack, they do not generate correctly for me when aliased. If the issue is just for me, then I'm ok with using aliased. Otherwise, looks like non-aliased is necessary.

lib/code_corps/github/sync/sync.ex
begedin: Maybe. Certainly something worth considering in a separate issue.

lib/code_corps/model/github_issue.ex
begedin: I was hoping you would have a better grasp, since you did some client work already on it.

My guess is, not much would change other than projectGithubRepo related stuff, since most of our UI is from the perspective of `Task` and `Comment`, not `GithubIssue` and `GithubComment`, and in that direction (task 1 - 1 issue), nothing has changed compared to the previous (task n - 1 issue)

lib/code_corps/github/sync/comment/comment/changeset.ex
joshsmith: Will need a follow-up issue if it doesn't exist then.

lib/code_corps/github/sync/sync.ex
joshsmith: Needs a follow-up issue.

PR #1194, opened 2017-11-15
Remove ProjectGithubRepo

joshsmith:
# What's in this PR?

- Removes ProjectGithubRepo
- Adds `project_id` to `GithubRepo`
- Migrates the data from one to the other in batches

Still need to:

- Add `PATCH` to `/github-repos/:id` endpoint


snewcomer (dismissed):
Nice!  Only looked at the model files and policy, but overall looks good.

joshsmith:
This needs tests for the controller, needs tests for the policy, and the segment tracking is not ideal.

I might suggest that we break some of the `update` action logic out into a context, too.

joshsmith:
I've added:

- policy tests
- controller test
- changeset tests

Still could use cleaning up the update action.

PR #1203, opened 2017-11-16
Fix modified_at comparisons to use seconds granularity

joshsmith:
# What's in this PR?

Fixes the way we're doing timestamp comparison to not error when GitHub sends us events with only second-level accuracy.


joshsmith:
I'm going to merge this without review since it's affecting events in production.

@begedin could you look at this ex post facto and tell me if you have any issues with it, and open an issue if so?

PR #1206, opened 2017-11-17
Add update action to /github-events endpoint

### LINKED ISSUE #1168: Add update action to /github-events endpoint
# Problem

We need to add an `update` action to the `GithubEventController` that:

- restarts the processing of the `GithubEvent` if it's not `processing` or `processed`
- does nothing otherwise

joshsmith:
# What's in this PR?

- adds validation of `retry` param when updating
- adds `update_changeset` to the event
- adds inclusion validation of the various statuses
- adds update policy
- adds update action to the controller
- adds tests for all of the above

## References
Fixes #1168


joshsmith:
This needs some tests for the webhook handler for the cases where the event already exists and needs to be reprocessed.

begedin (approved):
Looks fine, but my comment from the ember PR stands. This is an update action, but the user is not really patching a record. They're requesting performance of a non-RESTful action from the system.

PR #1216, opened 2017-11-19
Send Stripe API error message to error view

joshsmith:
# What's in this PR?

We weren't sending the specific error message to the client before.


joshsmith:
My main concern here is whether we can end up leaking Stripe error details we don't want to be leaking into the UI. Of course, it's a _worse_ situation to give vague hand-wavy 500 errors.

joshsmith:
My vote is to merge this, to be clear.

begedin (approved):

PR #1220, opened 2017-11-19
Report all errors to Sentry

joshsmith:
# What's in this PR?

Adds source code context and ensures that Sentry.Logger is running and reporting non-Plug errors.


joshsmith:
Just to comment, this is working now, so we should start seeing more Sentry errors @begedin.

PR #1224, opened 2017-11-19
Remove unnecessary fn

joshsmith:
# What's in this PR?

Removes an unnecessary private function wrapper.


begedin (approved):
Couldn't agree more 👍

PR #1225, opened 2017-11-20
Deduplicate Sync.Issue.GithubIssue, clean up

### LINKED ISSUE #1217: GitHub.Sync.Issue.GithubIssue duplicative logic
# Problem

`GitHub.Sync.Issue.GithubIssue` appears to have some duplicative `create_or_update_issue` logic.

We need to clarify whether this is true and come up with a plan to eliminate it.

begedin:
# What's in this PR?

This PR eliminates duplicative logic in `Sync.Issue.GithubIssue` by loosening the constraint of all these modules adhering to a strict parameter format.

It also fixes an issues we've missed otherwise - one of the two `create_or_update` cases was not ensuring an associated `GithubUser` record.

## References
Fixes #1217

### DIFF
--- a/lib/code_corps/github/sync/issue/github_issue/github_issue.ex
+++ b/lib/code_corps/github/sync/issue/github_issue/github_issue.ex
@@ -1,7 +1,7 @@
 defmodule CodeCorps.GitHub.Sync.Issue.GithubIssue do
   @moduledoc ~S"""
-  In charge of finding a `CodeCorps.GithubIssue` to link with a
-  `CodeCorps.Issue` when processing a GitHub Issue payload.
+  In charge of finding or creating a `CodeCorps.GithubIssue` to link with a
+  `CodeCorps.Task` when processing a GitHub Issue payload.
 
   The only entry point is `create_or_update_issue/2`.
   """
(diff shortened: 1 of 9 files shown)


lib/code_corps/github/sync/issue/github_issue/github_issue.ex
```diff
@@ -17,93 +17,38 @@ defmodule CodeCorps.GitHub.Sync.Issue.GithubIssue do
...
-      {:ok, github_issue}
+  @spec create_or_update_issue(map, GithubRepo.t, GithubPullRequest.t | nil) :: linking_result
+  def create_or_update_issue(%{} = payload, %GithubRepo{} = github_repo, github_pull_request \\ nil) do
+    with {:ok, %GithubUser{} = github_user} <- Sync.User.GithubUser.create_or_update_github_user(payload) do
+      payload
+      |> find_or_init()
+      |> GithubIssue.changeset(payload |> Adapters.Issue.to_issue)
+      |> Changeset.put_assoc(:github_user, github_user)
+      |> Changeset.put_assoc(:github_repo, github_repo)
+      |> Changeset.put_assoc(:github_pull_request, github_pull_request)
```
begedin, line 40: We have associated records, not associated ids, so `put_assoc/3` is more appropriate than `cast/3` + `assoc_constaint/2`. 

It also allows us to
- use a single changeset function
- not care about `github_pull_request` being `nil`

lib/code_corps/github/sync/issue/issue.ex
```diff
@@ -1,45 +1,38 @@
...
-  - create or update the `CodeCorps.Task` for the `CodeCorps.Project` in the
-    matched `CodeCorps.GithubRepo`
-
-  If the sync succeeds, it will return an `:ok` tuple with the created or
-  updated task.
-
-  If the sync fails, it will return an `:error` tuple, where the second element
-  is the atom indicating a reason.
+  Performs work identical to `sync/2`, with the adition of associating the
+  resulting `CodeCorps.GithubIssue` with the specified `CodeCorps.GithubPullRequest`.
```
begedin, line 13: I still find that I'm forced to use the full module path for generated doc html references to work correctly.

lib/code_corps/github/sync/issue/issue.ex
```diff
@@ -1,45 +1,38 @@
...
+  """
+  @spec sync(map, GithubRepo.t) :: Multi.t
+  def sync(%{} = payload, %GithubRepo{} = github_repo) do
     Multi.new
-    |> Multi.run(:github_issue, fn _ -> IssueGithubIssueSyncer.create_or_update_issue({github_repo, github_pull_request}, payload) end)
-    |> Multi.run(:issue_user, fn %{github_issue: github_issue} -> UserRecordLinker.link_to(github_issue, payload) end)
-    |> Multi.run(:task, fn %{github_issue: github_issue, issue_user: user} -> github_issue |> IssueTaskSyncer.sync_github_issue(user) end)
+    |> Multi.run(:github_issue, fn _ -> payload |> Sync.Issue.GithubIssue.create_or_update_issue(github_repo) end)
+    |> Multi.run(:issue_user, fn %{github_issue: github_issue} -> Sync.User.RecordLinker.link_to(github_issue, payload) end)
+    |> Multi.run(:task, fn %{github_issue: github_issue, issue_user: user} -> github_issue |> Sync.Issue.Task.sync_github_issue(user) end)
```
begedin, line 36: Some minor repetition here, but we likely want to drop transactions at some point anyway, so I figured we should be ok with this for now.

lib/code_corps/github/sync/sync.ex
```diff
@@ -37,15 +37,15 @@ defmodule CodeCorps.GitHub.Sync do
...
+  - Syncs the issue portion of the event with `Sync.Issue`
 
   [https://developer.github.com/v3/activity/events/types/#issuesevent](https://developer.github.com/v3/activity/events/types/#issuesevent)
   """
   @spec issue_event(map) :: outcome
   def issue_event(%{"issue" => issue_payload} = payload) do
     Multi.new
     |> Multi.run(:repo, fn _ -> RepoFinder.find_repo(payload) end)
-    |> Multi.merge(GitHub.Sync.Issue, :sync, [issue_payload])
+    |> Multi.merge(fn %{repo: github_repo} -> issue_payload |> Sync.Issue.sync(github_repo) end)
```
begedin, line 48: Reads better, IMO, and allows simplifying `Sync.Issue`

lib/code_corps/github/sync/issue/github_issue/github_issue.ex
```diff
@@ -17,93 +17,38 @@ defmodule CodeCorps.GitHub.Sync.Issue.GithubIssue do
...
-  """
-  @spec create_or_update_issue(GithubRepo.t, map) :: linking_result
-  def create_or_update_issue(%GithubRepo{} = repo, attrs) do
-    with {:ok, %GithubUser{} = github_user} <- GithubUserSyncer.create_or_update_github_user(attrs),
-         {:ok, %GithubIssue{} = github_issue} <- do_create_or_update_issue(repo, attrs, github_user)
-    do
-      {:ok, github_issue}
+  @spec create_or_update_issue(map, GithubRepo.t, GithubPullRequest.t | nil) :: linking_result
+  def create_or_update_issue(%{} = payload, %GithubRepo{} = github_repo, github_pull_request \\ nil) do
+    with {:ok, %GithubUser{} = github_user} <- Sync.User.GithubUser.create_or_update_github_user(payload) do
```
begedin, line 34: Since this whole module is wrapped into a transaction, if this with fails, the transaction will still fail.

However, I do believe we should pull this `GithubUser` creation step out of this module and move it into the parent module, as an extra step of that multi.

The problem with that is that we would then also need an additional step in the github repo sync process -> `syncing github users`.

If we do something similar for comments, we would likely do `syncing github users for issues` and `syncing github users for comments`.

Or we could just not report those steps and consider them parts of `syncing issues` and `syncing comments` steps.

Either way, eliminating this `with` and moving it into `Sync.Issue` would flatten out the behavior and make it slightly more explicit.

lib/code_corps/github/sync/sync.ex
```diff
@@ -125,37 +129,13 @@ defmodule CodeCorps.GitHub.Sync do
...
   defp sync_step({:error, _ = error}, _step), do: {:error, error}
 
-  @spec mark_repo(GithubRepo.t, String.t, Keyword.t) :: {:ok, GithubRepo.t} | {:error, Changeset.t}
-  defp mark_repo(%GithubRepo{} = repo, sync_state, opts \\ []) do
-    params = build_sync_params(sync_state, opts)
+  @spec mark_repo(GithubRepo.t, String.t, map) :: {:ok, GithubRepo.t} | {:error, Changeset.t}
+  defp mark_repo(%GithubRepo{} = repo, sync_state, params \\ %{}) do
     repo
-    |> GithubRepo.update_sync_changeset(params)
+    |> GithubRepo.update_sync_changeset(params |> Map.put(:sync_state, sync_state))
```
begedin, line 135: Instead of accepting a map as the second argument here, then doing this, we were accepting a `Keyword` list and then passed those options into a pipe of functions to convert it to a map.

I was not seeing the benefit, other than an ever so slightly nicer looking function call, assuming `[]` is objectively a nicer wrapper than `%{}` around a set of keys. :)

lib/code_corps/github/sync/sync.ex
```diff
@@ -182,18 +162,19 @@ defmodule CodeCorps.GitHub.Sync do
...
+         {:ok, pr_payloads} <- repo |> API.Repository.pulls |> sync_step(:fetch_pull_requests),
+         {:ok, repo} <- repo |> mark_repo("syncing_github_pull_requests", %{syncing_pull_requests_count: pr_payloads |> Enum.count}),
+         {:ok, pull_requests} <- pr_payloads |> Enum.map(&Sync.PullRequest.GithubPullRequest.create_or_update_pull_request(repo, &1)) |> ResultAggregator.aggregate |> sync_step(:sync_pull_requests),
          {:ok, repo} <- repo |> mark_repo("fetching_issues"),
-         {:ok, issue_payloads} <- repo |> GitHub.API.Repository.issues |> sync_step(:fetch_issues),
-         {:ok, repo} <- repo |> mark_repo("syncing_github_issues", [syncing_issues_count: issue_payloads |> Enum.count]),
-         {:ok, _issues} <- issue_payloads |> Enum.map(&Sync.Issue.GithubIssue.create_or_update_issue(repo, &1)) |> ResultAggregator.aggregate |> sync_step(:sync_issues),
+         {:ok, issue_payloads} <- repo |> API.Repository.issues |> sync_step(:fetch_issues),
+         {:ok, repo} <- repo |> mark_repo("syncing_github_issues", %{syncing_issues_count: issue_payloads |> Enum.count}),
+         paired_issues <- issue_payloads |> pair_issues_payloads_with_prs(pull_requests),
```
begedin, line 171: This does not fail, unless we have an actual bug, so no reason to wrap it in a sync step. It it breaks, it breaks always and needs to be fixed. It does not break occasionally.

test/lib/code_corps/github/sync/sync_test.exs
```diff
@@ -139,6 +153,24 @@ defmodule CodeCorps.GitHub.SyncTest do
...
+    #   requests (matched by "number")
+    test "matches github issue with github pull request correctly" do
+      {:ok, github_repo} = setup_test_repo() |> Sync.sync_repo
+
+      %GithubRepo{github_issues: github_issues} =
+        GithubRepo |> Repo.get(github_repo.id) |> Repo.preload(:github_issues)
+
+      linked_issues =
+        github_issues
+        |> Enum.filter(fn i -> not is_nil(i.github_pull_request_id) end)
```
asummers: Prefer `Enum.reject` here, so you can have a positive predicate. Then you can call it `linked_issue_count` and move the `Enum.count` up outside the `assert`.

joshsmith (dismissed):

joshsmith (approved):

PR #1226, opened 2017-11-21
Update stripity stripe

joshsmith:
# What's in this PR?

Updates stripity_stripe to `2.0-beta`.

This is a WIP.

We're going to need to remove `managed` from accounts and instead migrate them all to be `"custom"` `type` accounts.


joshsmith:
@begedin could you work tomorrow on getting the accounts migrated over and try to test this out locally to see if everything else is working okay?

We'll need to look through the rest of the changelog, too: https://stripe.com/docs/upgrades#api-changelog

We also should probably consider implementing `Stripe.Webhook` from stripity_stripe in order to verify Stripe webhooks that are incoming.

joshsmith:
```
warning: function Stripe.Card.create/3 is undefined or private. Did you mean one of:

      * create/1
      * create/2

  lib/code_corps/stripe_service/stripe_platform_card.ex:24

warning: function Stripe.Token.create_on_connect_account/3 is undefined or private
  lib/code_corps/stripe_service/stripe_connect_card.ex:38

warning: function Stripe.Card.create/4 is undefined or private. Did you mean one of:

      * create/1
      * create/2

  lib/code_corps/stripe_service/stripe_connect_card.ex:40

warning: function Stripe.Card.update/5 is undefined or private. Did you mean one of:

      * update/2
      * update/3

  lib/code_corps/stripe_service/stripe_connect_card.ex:67
```

joshsmith:
One thing we're not doing right now is dropping certain keys on their way out to Stripe that need dropped. The mappings we need on an update are different from those for a create.

joshsmith:
Another huge change: https://stripe.com/docs/transfer-payout-split

`transfers_enabled` is now `payouts_enabled`, with other possible changes forthcoming.

lib/code_corps_web/views/stripe_connect_account_view.ex
```diff
@@ -89,22 +89,47 @@ defmodule CodeCorpsWeb.StripeConnectAccountView do
...
+  defp get_recipient_status(%StripeConnectAccount{
+    legal_entity_verification_status: "pending",
+    verification_fields_needed: needed_fields}) do
+
+    # https://stripe.com/docs/api#account_object-verification-fields_needed
+    # Check if the list of required fields includes any fields from the specified
+    # group.
+    # Required fields are listed as an array, nested in groups using `.`, example:
+    # `group_a.field_a`, `group_a.field_b`, `group_b.field_a`, etc.
+    case needed_fields |> includes_field_from?("legal_entity") do
```
begedin: @joshsmith I rewrote this so I can clarify with a comment, but do keep in mind that **this change eliminates the `verifying status`** and I don't know what the reasoning behind it was.

If you did that by mistake and I reciprocated, we should revert the change.

Otherwise, we should document somewhere why we're doing it. I couldn't find any information about it.

If it was a mistake and you want to revert, you will have to revert the tests as well.

lib/code_corps_web/views/stripe_connect_account_view.ex
```diff
@@ -165,7 +190,6 @@ defmodule CodeCorpsWeb.StripeConnectAccountView do
   end
 
   defp get_bank_account_status(%StripeConnectAccount{
-    legal_entity_verification_status: "verified",
```
begedin: @joshsmith I'm not sure of the reasoning behind this change. It prevents us from differentiating between `"required"` and `"pending_requirement"`

begedin:
@joshsmith I've fixed the tests and warnings that I could. What's left are view tests failing due to changes that aren't clear to me.

begedin:
@joshsmith I performed the migration from `managed: true` to `type: "custom"` as a separate commit. 

It required a pretty major change on the `StripeConnectAccountView` due to how `ja_serializer` internals work.

When specifying a list of attributes for the serializer, a function is generated for each attribute, which simply returns `record.attribute`.

However, when building a payload, a `type` function is also called if it exists, instead of inferring the resource type from the module name. Since we now have a `type` attribute, it causes our account payload to look like:

```
%{
  "id" => 1,
  "type" => "custom" # or "standard" or "express",
  "attributes" => %{
    #...
    "type" => "custom"
  }
}
```

I would this a bug on `ja_serializer`, but apparently, the json API spec recommends to treat `type` and `id` as reserved words in the context of attribute names, i.e., not to use them, if possible. 

Fortunately, `ja_serializer` offers an alternative approach to specifying attributes,  which I tried out and it works fine, albeit ad a significant change linecount (though the actual changes in behavior are minor).

I went ahead and pushed the commit up, but if you disagree with this change and would prefer renaming `type` to something else locally, maybe `account_type`, that works to, but it will take longer and present another inconsistency between us and stripe, that we need to keep track of. 

http://jsonapi.org/format/#document-resource-object-fields

https://github.com/vt-elixir/ja_serializer/issues/236#issuecomment-293060191

joshsmith:
@begedin for future reference, there are other instances where I've run into the `type` problem and have just rendered it out differently in the view (using a different attribute name), but this is obviously not a great fix in the long-term.

PR #1235, opened 2017-11-22
Fix project user membership assignment

### LINKED ISSUE #1233: Bug: Only allow users to create their own project user records
From #1232, 

> A user can create their own `ProjectUser` record - **no role level check happens here**, so this is definitely a bug. Basically, accessing the API directly, any user can make themselves owner, collaborator, admin or pending member of a project

To fix this, we should enforce the following create rule

> A user can create their own `ProjectUser` record, provided the role is `"pending"`. This basically means they are allowed to apply for project memberships

joshsmith:
# What's in this PR?

Fixes #1233

### DIFF
--- a/lib/code_corps/policy/project_user.ex
+++ b/lib/code_corps/policy/project_user.ex
@@ -2,51 +2,83 @@ defmodule CodeCorps.Policy.ProjectUser do
   @moduledoc """
   Handles `User` authorization of actions on `ProjectUser` records
   """
-  import CodeCorps.Policy.Helpers, only: [get_role: 1]
+
+  import CodeCorps.Policy.Helpers, only: [get_membership: 2, get_project: 1, get_role: 1]
 
   alias CodeCorps.{ProjectUser, Repo, User}
 
-  @spec create?(User.t, map) :: boolean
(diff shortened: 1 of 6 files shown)


lib/code_corps/policy/project_user.ex
```diff
@@ -2,16 +2,51 @@ defmodule CodeCorps.Policy.ProjectUser do
...
+      # Non-member can only make pending if they're the user
+      [nil, "pending"] ->
+        user_id == author_id
+
+      # No one else can do anything
+      ["pending", _] ->
+        false
+
+      ["contributor", _] ->
+        false
```
begedin: A good approach if we want to make it extensible for other options down the line.

However, a simpler approach might be

```Elixir
def create?(%User{id: user_id}, %{"user_id" => member_id, "role" => "pending"}) when user_id == member_id, do: true
def create(%User{}, %{}), do: false
```

I think that works for all the cases we currently want to support, unless I'm missing something.

begedin (dismissed):
Looks good to merge for me, once tests pass, but I added a suggestion for a simpler `create?` approach if you want to use it.

begedin (dismissed):

PR #1237, opened 2017-11-23
Dialyzer Fixes

asummers:
Resolved about 10 dialyzer errors, and converted case with true/false to if/else. 

Discussion: the event handlers seem to all be doing the same high level logic. Perhaps it might be better to expose a `validator/0` that returns a validator module, and move the with/else logic to the caller. Haven't gone through all of them yet, but I could see that being advantageous.

### DIFF
--- a/lib/code_corps/github/event/pull_request/pull_request.ex
+++ b/lib/code_corps/github/event/pull_request/pull_request.ex
@@ -13,8 +13,6 @@ defmodule CodeCorps.GitHub.Event.PullRequest do
   }
   alias GitHub.Sync
 
-  @type outcome :: Sync.outcome | {:error, :unexpected_payload}
-
   @doc ~S"""
   Handles the "PullRequest" GitHub webhook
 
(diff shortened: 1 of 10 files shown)


lib/code_corps/github/event/pull_request/pull_request.ex
```diff
@@ -33,12 +32,12 @@ defmodule CodeCorps.GitHub.Event.PullRequest do
     end
   end
 
-  @spec validate_payload(map) :: {:ok, :valid}
-                               | {:error, :unexpected_payload}
+  @spec validate_payload(map) :: {:ok, :valid} | {:error, :unexpected_payload}
```
joshsmith, line 35: I think the old spec is maybe how the formatter is doing it now?

joshsmith (commented):
Overall looks good. One minor comment.

lib/code_corps/github/event/pull_request/pull_request.ex
asummers: Oh possibly. It was inconsistent across the handlers, so I just normalized them by hand. I can change it back if you want, but considering the discussion note in the PR, I feel like this code could be moved to the caller at some point soon. Your call, though.

joshsmith:
Can you rebase and squash into a single commit? Then I'll approve and we can merge.

joshsmith (approved):

PR #1244, opened 2017-11-24
Compile with warnings as errors. This will make warnings fail CI.

asummers:


joshsmith:
@begedin I'd like your thoughts on this before just blanket approving.

asummers:
https://hexdocs.pm/mix/Mix.Tasks.Compile.Elixir.html Docs. 

Basically if the compiler ever emits a warning (likely because of you doing something inadvertent) it will result in a nonzero exit code for that step, forcing you to fix it. The project does not currently have warnings, this just keeps that number to 0 :)

Example warnings (from a dependency, so immaterial to the CC build):

```
==> inch_ex
Compiling 15 files (.ex)
warning: String.strip/1 is deprecated, use String.trim/1
  lib/inch_ex/git.ex:36

warning: String.strip/1 is deprecated, use String.trim/1
  lib/inch_ex/git.ex:37

warning: Kernel.to_char_list/1 is deprecated, use Kernel.to_charlist/1
  lib/inch_ex/reporter/local.ex:33

warning: String.to_char_list/1 is deprecated, use String.to_charlist/1
  lib/inch_ex/reporter/remote.ex:25

warning: Enum.filter_map/3 is deprecated, use Enum.filter/2 + Enum.map/2 or for comprehensions
  lib/inch_ex/docs/retriever.ex:96

warning: Atom.to_char_list/1 is deprecated, use Atom.to_charlist/1
  lib/inch_ex/docs/retriever.ex:133

warning: Enum.filter_map/3 is deprecated, use Enum.filter/2 + Enum.map/2 or for comprehensions
  lib/inch_ex/docs/retriever.ex:202
```

begedin:
I'm completely fine with this, if I'm understanding correctly, and only our own warnings, not dependency warnings, will compile as errors.

begedin (approved):
Looks ok to me, as long as my assumption above is correct.

asummers:
@begedin That is correct :)

begedin:
There seems to be no downside for merging this and if there's a complication, it's easy enough to revert, so finally merged this. Thank you @asummers 👍

PR #1248, opened 2017-11-25
Update guardian

joshsmith:
# What's in this PR?

WIP. Haven't been able to figure it all out yet.


lib/code_corps/auth/bearer_auth_pipeline.ex
```diff
@@ -0,0 +1,8 @@
+defmodule CodeCorps.Auth.BearerAuthPipeline do
+  use Guardian.Plug.Pipeline, otp_app: :code_corps,
+                              module: CodeCorps.Guardian,
+                              error_handler: CodeCorps.Auth.ErrorHandler
+
+  plug Guardian.Plug.VerifyHeader, realm: "Bearer"
+  plug Guardian.Plug.LoadResource, allow_blank: true
```
joshsmith, line 7: This didn't need to be `allow_blank: true` before but appears to need to be now.

lib/code_corps/auth/error_handler.ex
```diff
@@ -0,0 +1,11 @@
+defmodule CodeCorps.Auth.ErrorHandler do
+  use CodeCorpsWeb, :controller
```
joshsmith, line 2: This appeared to be needed even though this isn't a controller.

lib/code_corps/auth/error_handler.ex
```diff
@@ -0,0 +1,11 @@
...
+
+  def auth_error(conn, {type, _reason}, _opts) do
+    # body = Poison.encode!(%{message: })
+    # send_resp(conn, 401, body) |> IO.inspect()
+    conn
+    |> put_status(401)
+    |> render(CodeCorpsWeb.TokenView, "401.json", message: to_string(type))
```
joshsmith: I'm not sure this is the message we want.

lib/code_corps/guardian.ex
```diff
@@ -0,0 +1,40 @@
...
+  end
+  def subject_for_token(user = %User{}, _claims) do
+    {:ok, "User:#{user.id}"}
+  end
+  def subject_for_token(_, _) do
+    {:error, :unknown_resource_type}
+  end
+
+  def resource_from_claims(%{"sub" => sub}), do: resource_from_sub(sub)
+  def resource_from_claims(_), do: {:error, :unknown_resource_type}
```
joshsmith: Not sure if this can happen.

lib/code_corps_web/controllers/password_reset_controller.ex
```diff
@@ -2,37 +2,31 @@ defmodule CodeCorpsWeb.PasswordResetController do
...
   def reset_password(conn, %{"token" => reset_token, "password" => _password, "password_confirmation" => _password_confirmation} = params) do
-    with %AuthToken{user: user} = auth_token <- AuthToken |> Repo.get_by(%{ value: reset_token }) |> Repo.preload(:user),
+    with %AuthToken{user: user} = auth_token <- AuthToken |> Repo.get_by(%{value: reset_token}) |> Repo.preload(:user),
          {:ok, _} <- Phoenix.Token.verify(conn, "user", reset_token, max_age: Application.get_env(:code_corps, :password_reset_timeout)),
-         {:ok, updated_user} <- user |> User.reset_password_changeset(params) |> Repo.update,
-         {:ok, _auth_token} <- auth_token |> Repo.delete,
-         {:ok, auth_token, _claims} = updated_user |> Guardian.encode_and_sign(:token)
+         {:ok, %User{} = updated_user} <- user |> User.reset_password_changeset(params) |> Repo.update(),
+         {:ok, _auth_token} <- auth_token |> Repo.delete(),
+         {:ok, auth_token, _claims} = updated_user |> CodeCorps.Guardian.encode_and_sign()
```
joshsmith, line 24: This seems to have changed in ways I don't understand (i.e. not sure why we don't need `:token` as a param now).

lib/code_corps_web/controllers/password_reset_controller.ex
```diff
@@ -2,37 +2,31 @@ defmodule CodeCorpsWeb.PasswordResetController do
...
     do
       conn
       |> Plug.Conn.assign(:current_user, updated_user)
       |> put_status(:created)
       |> render("show.json", token: auth_token, user_id: updated_user.id, email: updated_user.email)
-    else
-      {:error, %Changeset{} = changeset} -> conn |> put_status(422) |> render(CodeCorpsWeb.ErrorView, :errors, data: changeset)
-      {:error, _} -> conn |> put_status(:not_found) |> render(CodeCorpsWeb.ErrorView, "404.json")
-      nil -> conn |> put_status(:not_found) |> render(CodeCorpsWeb.ErrorView, "404.json")
```
joshsmith, line 34: Letting FallbackController work.

lib/code_corps/auth/bearer_auth_pipeline.ex
joshsmith: Looks like this is accurate and will need to remain this way.

lib/code_corps/auth/error_handler.ex
joshsmith: Not sure of a better way to handle this.

lib/code_corps/guardian.ex
joshsmith: I believe this _can_ happen, but the `:error` reason is probably not accurate. Maybe `:missing_resource`?

lib/code_corps_web/controllers/password_reset_controller.ex
joshsmith: It looks like it just changed quite a bit. This looks like it is the correct implementation now.

joshsmith:
Going to merge since all seems working and seems not so critical. Happy to have follow-up issues based on anything that seems not quite great.

PR #1251, opened 2017-11-26
Add popular skills

joshsmith:
# What's in this PR?

Adds popular skills to the `index` endpoint so you can find the most popular skills across the app.


joshsmith:
The coverage drop is not actually legitimate.

begedin (approved):
Not much to say about this one other than merge it in. 👍

PR #1252, opened 2017-11-26
Add dialyzer.ignore-warnings

joshsmith:
# What's in this PR?

Taking the cheap way out on some dialyzer warnings caused by upstream deps.


asummers (approved):
Ideally we'll be able to nuke all of these soon.

PR #1254, opened 2017-11-26
Ignore repo syncs for projects without project id.

### LINKED ISSUE #1221: Github sync is trying to sync events for repos that are connected to projects
# Problem

```
** (ArgumentError) nil given for :project_id. Comparison with nil is forbidden as it is unsafe. Instead write a query with is_nil/1, for example: is_nil(s.project_id)
```

This error is happening as the webhook handler is trying to sync data for repos that do not have a corresponding `project_id`.

We should ignore the event since there is no project to actually sync to.

asummers:
Fixes #1221

### DIFF
--- a/lib/code_corps_web/controllers/github_event_controller.ex
+++ b/lib/code_corps_web/controllers/github_event_controller.ex
@@ -1,9 +1,11 @@
 defmodule CodeCorpsWeb.GithubEventController do
   @moduledoc false
   use CodeCorpsWeb, :controller
+  import Ecto.Query, only: [from: 2]
 
   alias CodeCorps.{
     GithubEvent,
+    GithubRepo,
     GitHub.Webhook.Handler,
     GitHub.Webhook.EventSupport,
     Helpers.Query,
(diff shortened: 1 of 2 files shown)


joshsmith:
Just note to self that you're working this to:

- Add a regression test
- Fix the existing test failures

asummers:
Things to consider moving forward: 

There are "classes" of webhooks, some of which pertain to the repository and project and other that don't (installation, e.g.). As CC adds more webhook handlers for various reasons, the `should_process?` function will become increasingly further away from the handler logic, so this should eventually move down into the event handler logic, with the controller merely being a dispatcher.

lib/code_corps_web/controllers/github_event_controller.ex
```diff
@@ -50,9 +50,9 @@ defmodule CodeCorpsWeb.GithubEventController do
...
       if should_process?(payload) do
-        type
-        |> EventSupport.status(action)
-        |> process_event(type, delivery_id, payload)
+        process_status = type |> EventSupport.status(action)
+        process_status |> process_event(type, delivery_id, payload)
+        process_status
       else
         :ignored
       end
```
joshsmith: So will this line right now never run, the `:ignored` line?

joshsmith: Nevermind, I get it with the above comment. Should've waited.

asummers: The `:ignored` line will be executed in the case where the webhook event includes `%{"repository" => %{"id" => _}}` but the associated `GithubRepo` does not contain an associated `project_id` in the DB.

lib/code_corps_web/controllers/github_event_controller.ex
```diff
@@ -43,14 +45,29 @@ defmodule CodeCorpsWeb.GithubEventController do
...
   end
 
+  @spec should_process?(map) :: boolean
+  def should_process?(%{"repository" => %{"id" => repository_id}}) do
+    query = from repo in GithubRepo,
+      where: repo.github_id == ^repository_id,
+      where: not(is_nil(repo.project_id))
+    Repo.one(query) != nil
+  end
+  def should_process?(_), do: true
```
joshsmith: We probably want to make this fn private.

joshsmith (commented):
One minor change but otherwise LGTM.

joshsmith:
With that change and a squash into a single commit we're good to merge this.

lib/code_corps_web/controllers/github_event_controller.ex
asummers: We absolutely do, good catch.

joshsmith (approved):

PR #1257, opened 2017-11-26
Fix accidental nulling of pull request in Task.Service

### LINKED ISSUE #1253: Issue syncing to GitHub is trying to change relationship of github_pull_request
https://sentry.io/code-corps/code-corps-api/issues/409216164/

```
Elixir.RuntimeError: you are attempting to change relation :github_pull_request of
CodeCorps.GithubIssue but the `:on_replace` option of
this relation is set to `:raise`.

By default it is not possible to replace or delete embeds and
associations during `cast`. Therefore Ecto requires all existing
data to be given on update. Failing to do so results in this
error message.

If you want to replace data or automatically delete any data
not sent to `cast`, please set the appropriate `:on_replace`
option when defining the relation. The docs for `Ecto.Changeset`
covers the supported options in the "Related data" section.

However, if you don't want to allow data to be replaced or
deleted, only updated, make sure that:

  * If you are attempting to update an existing entry, you
    are including the entry primary key (ID) in the data.

  * If you have a relationship with many children, at least
    the same N children must be given on update.


  File "lib/ecto/changeset/relation.ex", line 201, in Ecto.Changeset.Relation.on_replace/2
  File "lib/ecto/changeset/relation.ex", line 131, in Ecto.Changeset.Relation.change/3
  File "lib/ecto/changeset.ex", line 1049, in Ecto.Changeset.put_change/7
  File "lib/ecto/changeset.ex", line 1136, in Ecto.Changeset.put_relation/5
  File "lib/code_corps/github/sync/issue/github_issue/github_issue.ex", line 40, in CodeCorps.GitHub.Sync.Issue.GithubIssue.create_or_update_issue/3
...
(3 additional frame(s) were not displayed)

(RuntimeError) you are attempting to change relation :github_pull_request of
CodeCorps.GithubIssue but the `:on_replace` option of
this relation is set to `:raise`.

By default it is not possible to replace or delete embeds and
associations during `cast`. Therefore Ecto requires all existing
data to be given on update. Failing to do so results in this
error message.

If you want to replace data or automatically delete any data
not sent to `cast`, please set the appropriate `:on_replace`
option when defining the relation. The docs for `Ecto.Changeset`
covers the supported options in the "Related data" section.

However, if you don't want to allow data to be replaced or
deleted, only updated, make sure that:

  * If you are attempting to update an existing entry, you
    are including the entry primary key (ID) in the data.

  * If you have a relationship with many children, at least
    the same N children must be given on update.
```

joshsmith:
# What's in this PR?

Fixes an issue where we putting a `nil` assoc with `github_pull_request` upon receipt of the payload from GitHub's API.

## References
Fixes #1253

### DIFF
--- a/lib/code_corps/task/service.ex
+++ b/lib/code_corps/task/service.ex
@@ -23,10 +23,12 @@ defmodule CodeCorps.Task.Service do
     :user
   ]
 
+  @type result :: {:ok, Task.t} | {:error, Changeset.t} | {:error, :github} | {:error, :task_not_found}
+
   @doc ~S"""
   Performs all actions involved in creating a task on a project
   """
-  @spec create(map) :: {:ok, Task.t} | {:error, Changeset.t} | {:error, :github}
+  @spec create(map) :: result
   def create(%{} = attributes) do
     Multi.new
     |> Multi.insert(:task, %Task{} |> Task.create_changeset(attributes))
@@ -38,7 +40,7 @@ defmodule CodeCorps.Task.Service do
     |> marshall_result()
   end
 
-  @spec update(Task.t, map) :: {:ok, Task.t} | {:error, Changeset.t} | {:error, :github}
+  @spec update(Task.t, map) :: result
   def update(%Task{} = task, %{} = attributes) do
     Multi.new
     |> Multi.update(:task, task |> Task.update_changeset(attributes))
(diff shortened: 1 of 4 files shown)


lib/code_corps/task/service.ex
```diff
@@ -86,10 +86,9 @@ defmodule CodeCorps.Task.Service do
   defp update_on_github(%Task{github_repo_id: _, github_issue_id: nil} = task), do: task |> create_on_github()
   defp update_on_github(%Task{github_repo: github_repo} = task) do
     with {:ok, payload} <- GitHub.API.Issue.update(task),
-         {:ok, %GithubIssue{}} <-
-           payload
-           |> Sync.Issue.GithubIssue.create_or_update_issue(github_repo) do
-      {:ok, Task |> Repo.get(task.id)}
+         {:ok, %GithubIssue{}} <- payload |> Sync.Issue.GithubIssue.create_or_update_issue(github_repo),
+         task <- Task |> Repo.get(task.id) do
```
asummers: If this is nil, what happens?

joshsmith: Could it be, since the main function expects the `task`?

joshsmith: I could move it back out of the `with`, though.

test/lib/code_corps/task/service_test.exs
```diff
@@ -166,6 +168,33 @@ defmodule CodeCorps.Task.ServiceTest do
...
+      task = insert(:task, github_repo: github_repo, github_issue: github_issue)
+
+      {:ok, updated_task} = task |> Task.Service.update(@update_attrs)
+
+      assert_received({:patch, "https://api.github.com/repos/octocat/Hello-World/issues/1347", _body, _headers, _options})
+
+      assert updated_task.id == task.id
+      assert updated_task.title == @update_attrs["title"]
+      assert updated_task.markdown == @update_attrs["markdown"]
+      assert updated_task.body != task.body
```
asummers: Prefer `refute ==` here

lib/code_corps/task/service.ex
asummers: It's low likelihood but imagine an update and a delete happening in parallel. Could add a `when not(is_nil(task))` guard and handle the nil case in the else.

asummers (dismissed):

asummers (approved):

PR #1260, opened 2017-11-27
Add ability to create Organization for users other than admin

### LINKED ISSUE #1230: Modify `Policy.Organization` to allow creation via invite
Currently, `Policy.Organization.create` only allows creation by admins. We need an additional clause to allow creation via invite.

- `create?/1` needs to become `create?/2` and accept `params` as the second argument
- new `create?/2` clause needs to check if `OrganizationInvite` exists based on `invite_token` parameter
- tests need to be expanded to check for this clause

# Estimates

~1 hour

vishaldeepak:
Fixes #1230

### DIFF
--- a/lib/code_corps/policy/helpers.ex
+++ b/lib/code_corps/policy/helpers.ex
@@ -6,6 +6,7 @@ defmodule CodeCorps.Policy.Helpers do
 
   alias CodeCorps.{
     Organization,
+    OrganizationInvite,
     ProjectUser,
     Project,
     ProjectUser,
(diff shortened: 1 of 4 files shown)


lib/code_corps/policy/helpers.ex
```diff
@@ -66,6 +67,15 @@ defmodule CodeCorps.Policy.Helpers do
 
   def get_organization(_), do: nil
 
+  @doc """
+  Retrieives an organiation invite from a struct, containing a `code` field
+  Returns `CodeCorps.OrganizationInvite` or nil
```
begedin: You got a typo here.

begedin (commented):
Other than the typo, this looks good to go. Quick change there and we can merge it. Great work!

lib/code_corps/policy/helpers.ex
vishaldeepak: Woops, made changes and pushed again!!

joshsmith (approved):

joshsmith:
🙌

PR #1261, opened 2017-11-27
Added filter by code support to org invites

begedin:
Required for code-corps/code-corps-ember#1556 to work correctly.

I took the opportunity to implement a generalized approach for filtering, to which we can potentially switch a few other endpoints as well.


joshsmith (commented):
What’s going on with the mix.lock?

begedin:
Just to answer the question, that's what happens for me when I have to do a `mix deps.get` if someone else updated a package in another PR and it got merged to dev.

joshsmith:
Are you sure you have the latest hex?

begedin:
I thought I did, but apparently not. Updated, so hopefully it does not happen again.

PR #1262, opened 2017-11-28
Improve organization creation

joshsmith:
# What's in this PR?

- Removes fulfilled from invite
- Replaces fulfilled with the organization id

Still needs tests.


lib/code_corps/organizations/organizations.ex
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  @spec fulfill_associated_invite(Organization.t, map) :: {:ok, OrganizationInvite.t | nil} | {:error, Changeset.t}
+  defp fulfill_associated_invite(%Organization{id: organization_id}, %{"invite_code" => code}) do
+    OrganizationInvite
+    |> Repo.get_by(code: code)
+    |> OrganizationInvite.update_changeset(%{organization_id: organization_id})
+    |> Repo.update()
+  end
+  defp fulfill_associated_invite(%Organization{}, %{}), do: {:ok, nil}
```
begedin, line 29: I added this to support the old "creation by admin" behavior. I know it's not actually used anywhere, but i think the options then should be either remove it, or keep it working. Breaking it and leaving it in while broken adds confusion.

lib/code_corps/policy/helpers.ex
```diff
@@ -67,15 +66,6 @@ defmodule CodeCorps.Policy.Helpers do
...
 
-  @doc """
-  Retrieves an organiation invite from a struct, containing a `code` field
-  Returns `CodeCorps.OrganizationInvite` or nil
-  """
-  @spec get_organization_invite(struct) :: OrganizationInvite.t() | nil
-  def get_organization_invite(%{"code" => code}),
-    do: OrganizationInvite |> Repo.get_by(code: code, fulfilled: false)
-  def get_organization_invite(%{}), do: nil
-
```
begedin, line 78: Moved it to `Policy.Organization` since it's the only place it's used in and we don't test ti directly.

joshsmith:
I can't approve since it's my PR, but looks good. Merging when tests pass.

PR #1267, opened 2017-11-29
Add organization approval

joshsmith:
# What's in this PR?

Adds approval of organizations.

### DIFF
--- a/lib/code_corps/model/organization.ex
+++ b/lib/code_corps/model/organization.ex
@@ -61,6 +61,13 @@ defmodule CodeCorps.Organization do
     |> put_change(:approved, false)
   end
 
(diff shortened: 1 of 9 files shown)


lib/code_corps/model/organization.ex
```diff
@@ -61,6 +61,12 @@ defmodule CodeCorps.Organization do
     |> put_change(:approved, false)
   end
 
+  def update_changeset(struct, params \\ %{}) do
```
asummers: Shouldn't `params` default to `[]`? Also `@spec`s

asummers (dismissed):
Seems fine after comment addressed

PR #1268, opened 2017-11-29
Added project filter in users index api

### LINKED ISSUE #1214: Filter users by username and project
# Problem

Right now we can filter users by username, but we cannot further limit that with a JOIN of `projects`.

We want to be able to pass in a `project_id` to the user search endpoint and restrict the query to users in that project only.

vishaldeepak:
Fixes #1214


begedin:
@vishaldeepak This slipped my notification queue. The filter looks good, but we usually place those inside `Module.Query`, so in this case, it would go into `CodeCorps.User.Query` and be placed under `lib/code_corps/user/query.ex`

That being said, we're trying to figure out the approach based on contexts, so I'm not even sure that's the correct place to put it.

@joshsmith Any idea where this should go in? Maybe we should actually have a `Queries` context?

joshsmith:
I think it would probably actually go in a `Users` context.

Perhaps @asummers or @starbelly have some thoughts on this?

ghost:
@joshsmith I agree. I don't think this belongs in the schema... it belongs in a context module. `Users` under the `Accounts` context namespace makes sense to me. My reasoning is that this seems like business logic and thus outside the scope of our repo. The code looks good though 👍 @vishaldeepak

joshsmith:
@vishaldeepak does the conversation above make sense re: a move into a context module?

vishaldeepak:
Sorry was busy. Understood, will raise a pull request soon . You basically would want me to add the code to `lib -> accounts -> users.ex` .
Edit - Made Changes and created a pull request

lib/code_corps_web/controllers/user_controller.ex
```diff
@@ -20,6 +21,7 @@ defmodule CodeCorpsWeb.UserController do
       |> Query.id_filter(params)
       |> Query.limit_filter(params)
       |> Query.user_filter(params)
+      |> Users.project_filter(params)
```
joshsmith: I think it's probably better if we just alias `CodeCorps.Accounts` and use the more explicit `Accounts.Users` here. What do you think?

lib/code_corps/accounts/users.ex
```diff
@@ -0,0 +1,11 @@
+defmodule CodeCorps.Accounts.Users do
+  alias CodeCorps.ProjectUser
+
+  import Ecto.Query
+
+  def project_filter(query, %{"project_id" => project_id}) do
+    from user in query,
+      join: pu in ProjectUser, on: pu.user_id == user.id and pu.project_id == ^project_id
+  end
+  def project_filter(query, _), do: query
```
joshsmith, line 10: Could we add a unit test for this `project_filter`?

joshsmith (commented):
Looking good, just a couple thoughts.

lib/code_corps_web/controllers/user_controller.ex
vishaldeepak: Thats a good idea, its more explicit and will cause less confusion later.

lib/code_corps/accounts/users.ex
vishaldeepak: What do you you think the test should do? There is one test already in the `user_controller_test`. How would this test be different?

ghost: @vishaldeepak  I would say they're testing two different things in isolation. True, we are testing this code via the user controller, but that behavior could change. We want a test for this specific unit of code so that even if the user controller changes, we still know this chunk of code is good... does this make sense?

vishaldeepak: added a test case let me know what you think. @begedin

begedin: Test unit looks fine. This ought to be good to go.

begedin (approved):

PR #1269, opened 2017-11-29
Improve project creation

joshsmith:
# What's in this PR?

Improves the project creation process by allowing nested skills/categories.

### DIFF
--- a/lib/code_corps/model/project.ex
+++ b/lib/code_corps/model/project.ex
@@ -9,9 +9,10 @@ defmodule CodeCorps.Project do
   import CodeCorps.Helpers.Slug
   import CodeCorps.Helpers.URL, only: [prefix_url: 2]
   import CodeCorps.Validators.SlugValidator
+  import Ecto.Query, only: [where: 3]
 
+  alias CodeCorps.{Category, Repo, Skill, TaskList}
   alias CodeCorps.Services.MarkdownRendererService
-  alias CodeCorps.TaskList
 
   alias Ecto.Changeset
 
@@ -42,9 +43,10 @@ defmodule CodeCorps.Project do
     has_many :task_lists, CodeCorps.TaskList
     has_many :tasks, CodeCorps.Task
 
-    has_many :categories, through: [:project_categories, :category]
-    has_many :skills, through: [:project_skills, :skill]
     has_many :stripe_connect_subscriptions, through: [:stripe_connect_plan, :stripe_connect_subscriptions]
+
+    many_to_many :categories, CodeCorps.Category, join_through: CodeCorps.ProjectCategory
+    many_to_many :skills, CodeCorps.Skill, join_through: CodeCorps.ProjectSkill
 
     timestamps()
   end
@@ -54,7 +56,7 @@ defmodule CodeCorps.Project do
     |> cast(params, [:title, :description, :long_description_markdown, :cloudinary_public_id, :default_color, :website])
     |> prefix_url(:website)
     |> validate_format(:website, CodeCorps.Helpers.URL.valid_format())
-    |> validate_required(:title)
+    |> validate_required([:title])
     |> generate_slug(:title, :slug)
     |> validate_slug(:slug)
     |> unique_constraint(:slug, name: :index_projects_on_slug)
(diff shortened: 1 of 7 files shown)


lib/code_corps/model/project.ex
```diff
@@ -42,10 +44,11 @@ defmodule CodeCorps.Project do
     has_many :task_lists, CodeCorps.TaskList
     has_many :tasks, CodeCorps.Task
 
-    has_many :categories, through: [:project_categories, :category]
-    has_many :skills, through: [:project_skills, :skill]
     has_many :stripe_connect_subscriptions, through: [:stripe_connect_plan, :stripe_connect_subscriptions]
 
+    many_to_many :categories, CodeCorps.Category, join_through: CodeCorps.ProjectCategory
+    many_to_many :skills, CodeCorps.Skill, join_through: CodeCorps.ProjectSkill
```
begedin, line 49: Probably wanna do an `on_delete: :delete_all` here. Default is `:nothing` according to the docs.

If this works out, I wouldn't mind doing it with tasks and users where applicable to.

lib/code_corps/model/project.ex
```diff
@@ -71,12 +74,69 @@ defmodule CodeCorps.Project do
...
+
+    changeset |> put_assoc(:skills, skills)
+  end
+  defp associate_skills(changeset, _) do
+    do_associate_skills(changeset, [])
+  end
+
+  defp do_associate_skills(changeset, _) do
+    changeset |> put_assoc(:skills, [])
+  end
```
begedin: We could save us a clause, and a `do_associate` helper, by using these:

```Elixir
defp associate_categories(changeset, %{"categories_ids" => ids}) when is_binary(ids) do
  categories =
    ids |> Enum.join(",") |> coalesce_id_string() |> find_categories()
  changeset |> put_assoc(:categories, categories)
end
defp associate_categories(changeset, _), do: changeset |> put_assoc(:categories, [])
```

test/lib/code_corps/model/project_test.exs
```diff
@@ -83,8 +83,17 @@ defmodule CodeCorps.ProjectTest do
...
     test "with valid attributes" do
+      category = insert(:category)
+      skill = insert(:skill)
       organization = insert(:organization)
-      attrs = %{title: "A title", organization_id: organization.id}
+      attrs = %{
+        "categories_ids" => [category.id],
+        "cloudinary_public_id" => "foo123",
+        "description" => "Description",
+        "skills_ids" => [skill.id],
```
begedin: We are using `coalesce_id_string` above. Wouldn't that mean we're supposed to be sending csv ids here, instead of a list?

lib/code_corps/model/project.ex
begedin: Note: I'm seeing some inconsistencies with `coalesce_ids` string and some of our tests.
If we're supposed to be getting a list here, instead of a csv `ids`string, then the first clause would guard with

```Elixir
when is_list(ids) when length(ids) > 0 do
```

begedin (dismissed):
I like the potential simplification we're getting out of this. Seeing some inconsistencies, though.

test/lib/code_corps/model/project_test.exs
joshsmith: No, because this is how JaSerializer modifies the data.

lib/code_corps/model/project.ex
```diff
@@ -86,22 +85,13 @@ defmodule CodeCorps.Project do
...
+    categories = ids |> Enum.map(&String.to_integer/1) |> find_categories()
     changeset |> put_assoc(:categories, categories)
   end
   defp associate_categories(changeset, _) do
-    do_associate_categories(changeset, [])
-  end
-
-  defp do_associate_categories(changeset, _) do
     changeset |> put_assoc(:categories, [])
   end
```
begedin: I'm reasonably sure you could still rewrite this

```Elixir
defp associate_categories(changeset, %{"categories_ids" => []}) do
  changeset |> put_assoc(:categories, [])
end
defp associate_categories(changeset, %{"categories_ids" => ids}) do
  categories = ids |> Enum.map(&String.to_integer/1) |> find_categories()
  changeset |> put_assoc(:categories, categories)
end
defp associate_categories(changeset, _) do
  changeset |> put_assoc(:categories, [])
end
```

as this

```Elixir
defp associate_categories(changeset, %{"categories_ids" => ids}) when is_list(ids) when length(ids) > 0 do
  categories = ids |> Enum.map(&String.to_integer/1) |> find_categories()
  changeset |> put_assoc(:categories, categories)
end
defp associate_categories(changeset, _) do
  changeset |> put_assoc(:categories, [])
end
```

Heck, since you're just using an enum map, you don't even need the `when length(ids) > 0` part.

lib/code_corps/model/project.ex
```diff
@@ -71,12 +73,46 @@ defmodule CodeCorps.Project do
...
     |> assoc_constraint(:organization)
   end
 
+  defp associate_categories(changeset, %{"categories_ids" => ids}) when is_list(ids) when length(ids) > 0 do
+    categories = ids |> Enum.map(&String.to_integer/1) |> find_categories()
+    changeset |> put_assoc(:categories, categories)
+  end
+  defp associate_categories(changeset, _) do
+    changeset |> put_assoc(:categories, [])
+  end
```
begedin, line 93: 👍 

Same thing applies to `associate_skills`, of course. 😬

begedin: Oh, nvm, it's there.

begedin: Oh, node version manager, it's there.

begedin (approved):
Looking good 👍

PR #1270, opened 2017-11-30
WIP: Add projects context

ghost:
# What's in this PR?

 - New projects context module
 - Updated test units for projects test

# Up for dicussion

 - Rename lib/code_corps/model to lib/code_corps/schemas ? 

# TODO (pending discussion)

- [ ] Update all test units in  CodeCorps.ProjectTest to use projects context module 
- [ ] Update tests unit to use ExMachina 
- [ ] Move application logic out of CodeCorps.Project and into context module
- [ ] Update Project controller and other pertinent areas

### DIFF
--- a/test/lib/code_corps/model/project_test.exs
+++ b/test/lib/code_corps/model/project_test.exs
@@ -3,14 +3,14 @@ defmodule CodeCorps.ProjectTest do
 
   import CodeCorps.Project
 
-  alias CodeCorps.{Project, ProjectUser, Repo}
+  alias CodeCorps.{Projects, Project, ProjectUser, Repo}
 
   describe "changeset" do
     @valid_attrs %{title: "A title"}
     @invalid_attrs %{}
 
     test "with valid attributes" do
-      changeset = Project.changeset(%Project{}, @valid_attrs)
+      changeset = Projects.change_project(%Project{title: "A title"})
       assert changeset.valid?
     end
 
--- a/lib/code_corps/projects/projects.ex
+++ b/lib/code_corps/projects/projects.ex
@@ -1,0 +1,104 @@
+defmodule CodeCorps.Projects do
+  @moduledoc """
+  The Projects context.
+  """
+
+  import Ecto.Query, warn: false
+  alias CodeCorps.Repo
+
+  alias CodeCorps.Project
+
+  @doc """
+  Returns the list of projects.
+
+  ## Examples
+
+      iex> list_projects()
+      [%Project{}, ...]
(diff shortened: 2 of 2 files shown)


ghost:
Note: Do ignore the hard coded attribute in the initial pull :)

test/lib/code_corps/model/project_test.exs
```diff
@@ -3,14 +3,14 @@ defmodule CodeCorps.ProjectTest do
...
-  alias CodeCorps.{Project, ProjectUser, Repo}
+  alias CodeCorps.{Projects, Project, ProjectUser, Repo}
 
   describe "changeset" do
     @valid_attrs %{title: "A title"}
     @invalid_attrs %{}
 
     test "with valid attributes" do
-      changeset = Project.changeset(%Project{}, @valid_attrs)
+      changeset = Projects.change_project(%Project{title: "A title"})
```
joshsmith, line 13: Should this not go into a separate test for the `Projects` context?

ghost: IMHO no, but it's debatable... by changing the tests to use the context module we will hit all the same places we're already directly testing.  And if you gen a phoenix 1.3 project and a few schema under the same context, you'll find that phoenix will put them all in the same test name space (i.e., hammer and chisel tests both go in the the toolbox context test file) . This makes sense to me, but is quite debatable I think.

joshsmith: Interesting. Not sure I'm fully caught up on 1.3 enough to understand the reasoning behind this.

ghost: https://hexdocs.pm/phoenix/contexts.html#content

As far as testing goes.... I would ask, what value do we get out of writing a project context test suite and a project schema test suite? If there is a value, then I'm all about it :)

ghost: @joshsmith thinking on this a few days... I do see the value in having a repo test and context test. At first this seemed like repeating myself... but let's say we only write tests for the context, and not for the repo... then we yank out the context... at that point we would have no tests for our schema module, unless other contexts use it, but this feels brittle. To add to that, we're testing two different things. Our context modules will surely end up having some business logic in them and/or delegate to modules/apps that have that logic, and so just a whole different set of variables and execution paths... Not to mention our context modules may or may not accept or return ecto changesets. TLDR; Different behaviors, inputs, and outputs. Thus it only makes sense to have test units for both. I will rework this.

Issue #1234, opened 2017-11-22
Project messaging


begedin:
This task should be considered the start off point for specs and discussion about the project messaging feature.

Our aim here is to have a basic system to allow users to message projects.

From the UI point of view, something like this could do the trick:

# Model

- Initially, it can be as simple as user-to-user, where posting a message to a project simply posts to organization owner
  - we need a sender, a recipient and a body field

# Reaching the UI

- We can add a link to the user dropdown, to reach the message inbox
- We can add a button to the project page to message a project - this opens the message thread UI and allows the current user to send a message to the project owner

# Message inbox

- we get a list of users we had contact with in the sidebar, ordered by most recent on top. clicking any opens the message thread for that user, on the same page, in the main area next to the sidebar

# Message thread

This is where we get to when starting a new message, or clicking an item in the message inbox. 

This is basically a route which fetches all messages either sent by the recipient to the current user, or sent to the recipient by the current user, from oldest to newest. Potentially, we can list only the last X messages, and fetch more as we try to scroll up.

There's a new message form at the bottom, which can be as simple as a text area. It does not need to support markdown, or anything like that. 

At this point, we have something shippable, and also implementable in a couple of days. 

Past this point we can 

# Enhancements

- support rendering of links in the message thread
- add buttons to the project membership page next to each member, to allow project owner and potentially admins/collaborators message each other
- add a similar button to each user's profile page
- extend the message model to allow for project-to-user, user-to-project. Since ecto prefers flattened relationships instead of polymorphism, this can be easily built upon in iterations like this. That would mean the project message button now targets project
- give each message a "read" status. Sounds simple, but with user-to-project, we need to decide how to handle "reading" messages targeted at projects. Might want to merge it with notifications, once we have them, since they'll likely end up serving the same purpose in this context.
- privacy/do-not-disturb settings
- add organization as a third recipient/sender option, if there's a need for it
- plug notifications into it, once we have them

joshsmith:
We just discussed the model here and came up with a simplified version:

### Message

- subject (required if sent by project)
- body (HTML or text)
- initiated_by ("admin" or "user")
- belongs_to author
- belongs_to project
- has_many conversations

### Conversation

- open
- read_at (last read time by admin in project)
- belongs_to message
- belongs_to user (target of conversation)
- has_many conversation_parts

### Conversation Part

- body
- read_at (by admin in project if author is user, otherwise by user)
- belongs_to author
- belongs_to conversation

[...]

PR #1278, opened 2017-12-05
Add messages

joshsmith:
# What's in this PR?

- Message
  - model
  - policy
  - controller
  - view

## References
Progress on: #1234


lib/code_corps/model/message.ex
```diff
@@ -0,0 +1,48 @@
...
+
+  @type t :: %__MODULE__{}
+
+  schema "messages" do
+    field :body, :string
+    field :initiated_by, :string
+    field :subject, :string
+
+    belongs_to :author, CodeCorps.User
+    belongs_to :project, CodeCorps.Project
```
begedin, line 26: If I'm understanding this correctly, this has no recipient yet? That goes into `Conversation`? The author is the user starting the message, the project is the project the user is a member/admin of.

If this is correct, or if it not correct and I'm misunderstanding it completely, I'm thinking it should be part of the
`moduledoc` either way, since it's not completely intuitive (albeit for good reasons).

lib/code_corps/model/message.ex
```diff
@@ -0,0 +1,48 @@
...
+  @spec maybe_validate_subject(Ecto.Changeset.t) :: Ecto.Changeset.t
+  defp maybe_validate_subject(changeset) do
+    initiated_by = changeset |> Ecto.Changeset.get_field(:initiated_by)
+    changeset |> do_maybe_validate_subject(initiated_by)
+  end
+
+  defp do_maybe_validate_subject(changeset, "admin") do
+    changeset |> validate_required(:subject)
+  end
+  defp do_maybe_validate_subject(changeset, _), do: changeset
```
begedin: I would just call this chain `require_subject_if_admin`. Makes it clearer and foregoes the need for a comment

lib/code_corps/policy/message.ex
```diff
@@ -0,0 +1,26 @@
...
+    true
+  end
+  def show?(%User{} = user, %Message{} = message) do
+    message |> get_project() |> administered_by?(user)
+  end
+  def show?(_, _), do: false
+
+  def create?(%User{id: id}, %{"initiated_by" => "user", "author_id" => author_id}) do
+    id == author_id
+  end
```
begedin: a guard clause would work better here

```Elixir
def create?(%User{id: id}, %{"initiated_by" => "user", "author_id" => author_id}) 
  when id == author_id, do: true
```

lib/code_corps/model/message.ex
joshsmith: The `author` is not necessarily a member of the project. The message could be to the project without the `author` being a member.

lib/code_corps/model/message.ex
joshsmith: I didn't want to tie it to the admin specifically in the method signature, but probably fine for now to do that.

lib/code_corps/policy/message.ex
joshsmith: Only because the fallback is false. I'll make the change, but I'm not sure if I understand the argument.

begedin: Using a guard may be faster (not sure if it actually is here) and, additionally, stresses that there are no side effects to the check since only kernel functions can be used in guards.

It seems arbitrary, and probably doesn't affect much, if anything here, but generally it's preferred to avoid entering the function at all if it's easy to do, which it is here.

begedin:
@joshsmith I've moved the list subclause logic into a separate `Query` module and made it so multiple filters can be applied at once. I think we should do this with most other endpoints, and here's my reasoning.

- Any client hitting the API, including our own will probably expect to have multiple filters being applied
- For example, if we do `GET /messages?project_id=1&author_id=2`, we expect messages from project 1, by author 2, not just **all messages from project 1, ignore the author**
- If a client request an array of ids, the client should worry about not additionally filtering by author or project. If they do specify, we should not assume they made a mistake and just give them those ids, ignoring the rest of the query, because we then effectively disable the ability to do that "mistake" intentionally.

begedin:
@joshsmith I couldn't find an authorization approach for index that wouldn't completely kill our performance, so instead, I pushed up a commit which offers an alternative - `scope`

With this, our index route would not return a 403 and instead just return the data the user is authorized to view. That would mean anything ranging from no data, to some data, to all the data in the case of admin users.

The idea is that we apply the scope first, then apply the filters, although technically, any order would work.

Looking at authorization libraries, scopes are a common solution to the problem of index authorization.

For endpoints where we discriminate purely on user type (admin or not admin) returning a 403 would still be an option, but I really don't think returning a 403 when only some of the results are inaccessible is intuitive, nor is it manageable performance-wise.

However, if you disagree, I tried to write this without affecting the rest of the PR, so it can be reversible.

Even `pundit`, probably one of the better developed authorization libraries out there seems to encourage scopes for index actions. Yes, that's Ruby, not Elixir, but considering how Ecto fetches data (we build the query first, then explicitly execute it using `Repo.all`), it seems even more suited for this approach.

begedin:
To make my previous commit stick, what we need to do is 

- remove `index?` clauses from `Policy.Message`
- remove `index?` clause for `Message` from `Policy`
- remove `index?` tests
- consider creating an issue to switch to a scope approach on the `GithubEvent` index endpoint to, albeit that one is clearer on who gets to access it, regardless of data, so it may make sense to keep it around

begedin:
While working on this, I found a bug, or at the very least, an inconsistency we need to deal with. I created an issue for it in #1282

begedin:
@joshsmith You're the author, so I can't request your review, but this checks all the boxes in the PR description, so it should be good to review/merge

begedin:
@joshsmith Due to findings from other issues I've created, it may be that 

- we don't really need an index endpoint for messages, other than maybe for coalesced id requests. 
- we do need a show endpoint
- we do need a create endpoint
- the create endpoint needs to create a conversation record alongside the message record

That means good chunks of this PR will get removed. However, I would still like to get it merged, since it's possible to reappropriate most of the code for reuse in issues linked above - #1286, #1287, #1288, #1289, #1290, #1291, #1292 

I've created #1294  and #1293 to deal with remaining problems once this is merged

snewcomer:
@joshsmith @begedin This looks great!  So a conversation has many messages?  Is the reason to move the index route to a Conversation due to the desire to create threads?  If so, what role does Messages play in the UI?  Trying to draw out a map of Conversations and Messages and how it is painted in the UI.

begedin:
@snewcomer Actually, it's the other way around.

A project admin could send out a message to multiple users. That would create a single message, belonging to the project and authored by the admin, as well as multiple conversations - one for each user.

Once replies to a conversation start going in, conversation parts get created.

That means you can run separate conversations with different users from one starting message.

On the other side, a user could also message a project. This would create a Message authored by that user, belonging to that project, as well as a single Conversation belonging to that user and that message.

Once replies start going in, it works pretty much the same as above.

As I said in my previous comment, this PR is sort of wrong, and needs to be corrected post merge, but the work done here is useful for solving other problems, so we want to get it in.

joshsmith:
I think we do need an `index` for messages because we'd still want to show which manual messages were created by a project in some other yet-to-be-determined UI.

begedin:
@joshsmith As I said, we might need it anyway, which is one of several reasons why I'd prefer to merge this as is, barring any serious oversights, and move on from there.

joshsmith:
I can't review myself but this is good to merge.

PR #1295, opened 2017-12-13
1286 conversation models

### LINKED ISSUE #1286: Add Conversation and ConversationPart models and schemas
# Problem

These two really go together, so we likely need to add them together.

We should not deal with changesets as part of this issue, as we really aren't sure yet what they might look like.

The idea is that a project can start a message with a user. This creates a `Message` record, with the following fields

- `initiated_by: "admin"`
- `author` - the project member who created it, 
- `project` - the project the author is creating the message in the name of

A `Conversation` record is then also created
- belongs to the newly created message
- belongs to the user the message is sent to

From the other side, a user can also message a project. This, again, creates a `Message` record, with the following fields:

- `initiated_by: "user"`
- `author` - the outside user who messaged the project
- `project` - the project that gotmessaged

Again, a `Conversation` is also created
- belongs to the user who sent the message
- belongs to the project the message was sent to

In both cases, any side can then post further messages to the same conversation. This creates `ConversationPart` records

## Conversation fields

- `status` - `"open"`/`"closed"` - string, defaults to `"open"`
- `read_at` - timestamp, defaults to `null`
- `belongs_to :message` - cannot be `null`
- `belongs_to :user` - cannot be `null`
- `has_many :conversation_parts`

## `ConversationPart` fields

- `body`
- `read_at` (by admin in project if author is user, otherwise by user)
- `belongs_to :author`
- `belongs_to :conversation`

## Subtasks

- [x] Add migration with the `"conversations"` and `"conversation_parts"` tables
- [x] Add `lib/code_corps/model/conversation.ex`
- [x] Add `lib/code_corps/model/conversation_part.ex`

## References

- we need to get #1278 merged, so we have a `Message` model to link `Conversation` with

begedin:
# What's in this PR?

__Setting the base to #1278  for the time being, since it's not merged yet, so reviews can be easier. Base should be changed to develop once #1278 is merged.__

This PR adds Conversation and ConversationPart schemas and models, with correct associations with `CodeCorps.Message`, `CodeCorps.User` as well as each other.

I also implemented the factories for the two new models and have added indexes based on what we intend to filter the records by.

## References
Fixes #1286 

Progress on: https://github.com/code-corps/code-corps-api/milestone/22


joshsmith:
I think I'd specified either `open` or `closed` as a boolean rather than a string. There are perf implications to one vs the other and if it's only anticipated to be a boolean, we should treat it as a boolean.

joshsmith (approved):

joshsmith:
Coverage went down because factories are unused, but that will change.

PR #1296, opened 2017-12-13
Conversation index endpoint with supporting infrastructure

### LINKED ISSUE #1288: Add ConversationController index endpoint and scope
# Problem

Our client `/project/messages` and `/user/messages` routes will be in fact fetching conversations.

That means the current `:index` endpoint being implemented in #1278 is over-implemented and par of the behavior will have to be moved into the new endpoint. 

That being said, all of the work being done there will be transferable to the solution for this issue.

The `ConversationController` needs potentially just one, but likely two endpoints. 

This issue is for the `:index` endpoint.

## Subtasks

- [ ] Implement `Policy.Conversation.scope`, with tests

This is a custom query involving the message scope, in that, the user has access to all conversations belonging to messages they initiated, as well as all conversations belonging to them directly (meaning they were the target of a message sent by a project admin).

```Elixir
def scope(query, %User{id: id} = current_user) do
  scoped_message_ids = 
    Message 
    |> Policy.Message.scope(current_user)
    |> select([m], m.id)
    |> Repo.all
  
  query
  |> where(user_id: ^id)
  |> or_where([c], c.message_id in scoped_message_ids)
end
```

- [ ] tests for `Policy.Conversation.scope`
- [ ] Implement `:index` endpoint with tests
  - needs to scope correctly
  - needs to apply correct status filter
  - needs to apply `project_id` filter for the purposes of /:project/messages
  - no need to apply a `user_id` filter yet, the scope will be enough for `/:user/messages`
  - no need for coalesced id support - the client should fetch conversations directly, not through `message.conversations`. If there's a need later, we can plug it in

From https://github.com/code-corps/code-corps-api/issues/1234#issuecomment-350776870

> the messages/conversations index page should, by default query conversations initiated by users with any amount of parts, as well as convs. initiated by projects with at least one part (so we don't render unreplied to messages sent out to multiple users)

This is a weird route/endpoint which in a way is indexing two schemas at once.

We'd have to scope the conversations using the message scope, apply `project_id` filter to that scope, then fetch conversations from those messages, which fulfil the "status" rule and potentially any additional filter.

```Elixir
# in CodeCorpsWeb.ConversationController

def index(conn, params) do
  Message
  |> Policy.Conversation.scope(current_user) 
  |> Messages.list_conversations(params)
end

# in CodeCorps.Messages

def list_conversations(scope, %{} = params) do
  scope
  |> ConversationQuery.project_filter(params)
  |> ConversationQuery.status_filter(params)
end

# in CodeCorps.Messages.ConversationQuery

def project_filter(query, %{"project_id" => project_id}) do
  query
  |> join(:left, [c], m in Message, c.messsage_id == m.id)
  |> where([_c, _m], m.project_id == ^project_id)
end
def status_filter(query, %{}), do: query

# "status" => "active" is what the client is querying for by default,
# and is the default query parameter for the route, so it isn't rendered in the URL
def status_filter(query, %{"status" => "active"}) do
  query
  |> where([initiated_by: "user"]) 
  |> or_where([initiated_by: "admin, part_count > 0])
end
def status_filter(query, %{}), do: query
```

- [ ] tests for index endpoint
- [ ] tests for `Messages.list_conversations/2`

## References

- needs #1286 done

begedin:
# What's in this PR?

___This PR basess off of #1295 to facilitate review. Should be rebased to `develop` once #1295 is merged.___

This PR adds a ConversationController with an `:index` endpoint as well as all associated infrastructure
- `Messages.list_conversations/2`
- `ConversationView`
- `Policy.Conversation.scope/2`

Tests have also been added.

## References

Fixes #1288

Progress on: https://github.com/code-corps/code-corps-api/milestone/22

### DIFF
--- a/lib/code_corps/messages/conversation_query.ex
+++ b/lib/code_corps/messages/conversation_query.ex
@@ -1,0 +1,46 @@
+defmodule CodeCorps.Messages.ConversationQuery do
+  @moduledoc ~S"""
+  Holds helpers to query `CodeCorps.Conversation` records using a map of params.
+  """
+
+  import Ecto.Query
+
+  alias CodeCorps.{Conversation, ConversationPart, Message, Repo}
+  alias Ecto.Queryable
+
+  @doc ~S"""
+  Narrows down a `CodeCorps.Conversation` query by `project_id` of the parent
+  `CodeCorps.Message`, if specified in a params map
+  """
+  @spec project_filter(Queryable.t, map) :: Queryable.t
+  def project_filter(queryable, %{"project_id" => project_id}) do
+    queryable
+    |> join(:left, [c], m in Message, c.message_id == m.id)
+    |> where([_c, m], m.project_id == ^project_id)
+  end
+  def project_filter(queryable, %{}), do: queryable
+
+
+  @doc ~S"""
+  Narrows down a `CodeCorps.Conversation` query to return only those records
+  considered to have a specific status.
+
+  The status of `active` means that only those records are included which either
+  - belong to a `CodeCorps.Message` initiated by user
+  - belong to a `CodeCorps.Message` initiated by admin, with at least a single
+    reply in the form of a `CodeCorps.ConversationPart`
+  """
+  @spec status_filter(Queryable.t, map) :: Queryable.t
+  def status_filter(queryable, %{"status" => "active"}) do
+    prefiltered_ids = queryable |> select([c], c.id) |> Repo.all
+
+    Conversation
+    |> where([c], c.id in ^prefiltered_ids)
+    |> join(:left, [c], m in Message, c.message_id == m.id)
+    |> join(:left, [c, _m], cp in ConversationPart, c.id == cp.conversation_id)
+    |> group_by([c, m, _cp], [c.id, m.initiated_by])
+    |> having([_c, m, _cp], m.initiated_by == "user")
+    |> or_having([c, m, cp], m.initiated_by == "admin" and count(cp.id) > 0)
+  end
+  def status_filter(query, %{}), do: query
+end
--- a/lib/code_corps/messages/messages.ex
+++ b/lib/code_corps/messages/messages.ex
@@ -3,7 +3,7 @@ defmodule CodeCorps.Messages do
   Main context for work with the Messaging feature.
   """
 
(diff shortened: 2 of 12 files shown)


lib/code_corps/messages/conversation_query.ex
```diff
@@ -0,0 +1,46 @@
...
+  considered to have a specific status.
+
+  The status of `active` means that only those records are included which either
+  - belong to a `CodeCorps.Message` initiated by user
+  - belong to a `CodeCorps.Message` initiated by admin, with at least a single
+    reply in the form of a `CodeCorps.ConversationPart`
+  """
+  @spec status_filter(Queryable.t, map) :: Queryable.t
+  def status_filter(queryable, %{"status" => "active"}) do
+    prefiltered_ids = queryable |> select([c], c.id) |> Repo.all
```
begedin, line 35: Due to additional joins below, we cannot just pipe the existing query due to potentially applying the project filter above as well. The SQL query ecto builds get's all garbled up with multiple joins and is no longer applying conditions to correct schemas and fields. Unfortunately, this is the only way I could find to support flexibility of multiple queries.

lib/code_corps/messages/conversation_query.ex
```diff
@@ -0,0 +1,46 @@
...
+
+    Conversation
+    |> where([c], c.id in ^prefiltered_ids)
+    |> join(:left, [c], m in Message, c.message_id == m.id)
+    |> join(:left, [c, _m], cp in ConversationPart, c.id == cp.conversation_id)
+    |> group_by([c, m, _cp], [c.id, m.initiated_by])
+    |> having([_c, m, _cp], m.initiated_by == "user")
+    |> or_having([c, m, cp], m.initiated_by == "admin" and count(cp.id) > 0)
```
begedin, line 43: Adding a `reply_count` counter cache field to `Conversation` might speed things up a bit here and it would certainly replace the `having/or_having` with `where/or_where`, but I would do it as part of a separate issue, since updating it would involve `ConversationPart` changesets.

My vote is, separate issue for consideration, and mark it low priority for now. If we find we need it for performance reasons, we can do it then.

joshsmith (approved):

PR #1300, opened 2017-12-15
Moves nil parameter removal from github adapter to user changeset

### LINKED ISSUE #996: Move email nil behavior from GitHub.Adapters.User to Account.create_from_github
# Problem

At the moment, our github user adapter specifically removes the email key if the value in the github payload is nil. The reasoning behind it is that we do not want to unset the user email from github.

However, this is an "input" validation issue and should not be part of the behavior of an adapter. Instead, it should go into the github user changeset in Accounts.create_from_github

The tests should follow suit as well.

iref:
# What's in this PR?

* Moved nil parameter filtering from Github adapter to `Accounts.Changeset.create_from_github_changeset`,
* Fixed `RecordLinker` test setup that was dependent on nil filtering.

## References
Fixes #996


test/lib/code_corps/github/adapters/user_test.exs
```diff
@@ -20,15 +31,15 @@ defmodule CodeCorps.GitHub.Adapters.UserTest do
...
     end
 
     test "maps User type" do
-      assert User.to_user(%{"type" => "User"}) == %{type: "user"}
+      assert User.to_user(%{"type" => "User"}) == expected_payload("user")
     end
 
     test "does not map Organization type" do
-      assert User.to_user(%{"type" => "Organization"}) == %{type: "Organization"}
+      assert User.to_user(%{"type" => "Organization"}) == expected_payload("Organization")
```
begedin, line 42: I'm seeing some inconsistencies here which I'm unsure about.

Any idea why the other two are tested with lowercase, while this one is uppercase?

begedin (commented):
Generally looks good, bit I do have a small question.

test/lib/code_corps/github/adapters/user_test.exs
iref: I am not sure what's reason, but `Github.Adapters.Users` enforces lowercase `user` and `bot` type, but not `Organization`. @joshsmith may know more.

joshsmith: Hmm I'm not sure there's any good reason for that. We probably should enforce it, too.

begedin: @iref Would you be ok with trying to add that enforcement as part of this PR? We may have a potential bug or at least inconsistency and I'm hesitant to merge until we know more.

iref: Type normalization isn't related to original issue, so new PR might be better. Just to keep changeset as small as possible. Although I don't feel super strong about it. :)

Do we want to normalize any type value in future? This change wouldn't need additional changes if Github adds new user type, but I don't know if it's desired behaviour or we rather normalize types per case.

begedin: @joshsmith What do you think? It's definitely out of scope for this PR, so I'd be fine with a new issue, but it's something we probably ought to consolidate.

It might involve migrating data to. Do we ever store "users" of type "Organization"?

joshsmith: I think I’ve seen this come up. Honestly it’s something that should also be normalized at the database level too in these cases by lower casing the values in the database.

In which case I think this should be dealt with as a separate issue.

begedin (approved):

begedin:
Created issue #1345 for the "bug" we've discovered. Merged this since it's good otherwise. Thank you @iref 👍

PR #1301, opened 2017-12-15
Add support for creation of associated conversations with message via params

### LINKED ISSUE #1293: Discussion: Best way to create Conversations when creating Message
# Problem

Our `MessageController` `:create` endpoint, implemented in #1278 is not enough.

It needs to support creating associated `Conversation` records alongside the created `Message` record, for each user specified as the recipient when a project admin is starting the thread.

When a user is starting the thread, a single `Conversation` record needs to be created for that user.

The question is, which side is the smart one here? Do we specify these associations on the client as directly as possible, the way we do with user skills, for example, or do we specify them __dumbly__ and let the API inferr the correct information?

My vote would go for smarter client, simply because it allows for more flexibility.

When starting a thread as a project
- init new `message` record, initiated by admin, authored by current user, belonging to project
- for each user (the UI will probably allow for one initially), create new `conversation` record and put it into `message.conversations`
- save the `message` record
- API casts those asssocations

When starting a thread as a user
- init new `message` record, initiated by user, authored by current user, belonging to project
- create single new `conversation` record for current user and put it into `message.conversations`
- save the `message` record
- API casts that one asssocation

The only alternative that I can think of is

When starting thread as project admin
- send virtual attributes indicating a list of users the message targets
- API inferrs users from data in those attributes

When starting thread as user
- API inferss conversation from current user

This alternative makes for a more fragmented codebase and is more difficult to follow, in my opinion, which is why I prefer the first approach.

## Subtasks
- [ ] Once this discussion is over, close issue and open another one to implement the decided on approach

@joshsmith What do you think?

begedin:
# What's in this PR?

**Update**

Still need:

- [x] tests for `DataToAttributes` plug
- [x] tests for `Messages.create` to check it's possible to create a `Conversation` alongside
- [x] tests for controller create action, to serve as integration between `DataToAttributes` and `Messages.create`

**Original text**

This PR progress on and hopefully closes #1293 

It's been marked as discussion, but due to requiring the set of features described there for client PR code-corps/code-corps-ember#1597, I was forced to start on some type of implementation.

Will comment on specific changes, but to put it in short

`ember-data`, by default, does not support including records during post. It has a basic level of support for associating with existing has-many/many-to-many children in the form of the basic `{type, id}` identifier map, but not a full include, with all the child record's attributes.

ja_serializer does not support it either, so I was forced to provide my own implementation on both ends.

The alternatives are

- **create conversation manually from client, after creating message** - To me, this may actually make sense. It makes the API less opinionated and allows potential other clients later on more flexibility on how they implement their own flow. On the downside, there's no clean way to clean up a message without child conversations if the process is interrupted.

- **Let the API infer which child records to create** - This makes the API extremely opinionated, forces the client down an inflexible route and creates coupling between the two, so I'm against it. We would have to agree on some non-standard way to provide a list of user ids to create conversations for. I guess, we could define a `message has many users through conversations` relationship and do what we do with project skills, project categories, etc, but really, which is one way where I could agree to do it that way

- Can't think of any other approach at the moment

## References

Progress on: #1293
Tied to: code-corps/code-corps-ember#1597

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -43,7 +43,7 @@ config :code_corps, CodeCorps.Repo,
   pool_size: 10
 
 # CORS allowed origins
-config :code_corps, allowed_origins: ["http://localhost:4200"]
+config :code_corps, allowed_origins: ["http://localhost:4200", "chrome-extension://pfdhoblngboilpfeibdedpjgfnlcodoo"]
 
 config :code_corps, CodeCorps.Guardian,
   secret_key: "e62fb6e2746f6b1bf8b5b735ba816c2eae1d5d76e64f18f3fc647e308b0c159e"
--- a/lib/code_corps/messages/conversations.ex
+++ b/lib/code_corps/messages/conversations.ex
@@ -1,0 +1,19 @@
+defmodule CodeCorps.Messages.Conversations do
+  @moduledoc ~S"""
+  Subcontext aimed at managing `CodeCorps.Conversation` records aimed at a
+  specific user belonging to a `CodeCorps.Message`.
+  """
+
+  alias Ecto.Changeset
+
+  alias CodeCorps.{Conversation}
+
+  @doc false
+  @spec create_changeset(Conversation.t, map) :: Ecto.Changeset.t
+  def create_changeset(%Conversation{} = conversation, %{} = attrs) do
+    conversation
+    |> Changeset.cast(attrs, [:user_id])
+    |> Changeset.validate_required([:user_id])
+    |> Changeset.assoc_constraint(:user)
+  end
+end
--- a/lib/code_corps/messages/messages.ex
+++ b/lib/code_corps/messages/messages.ex
@@ -67,6 +67,11 @@ defmodule CodeCorps.Messages do
   def create(%{} = params) do
     %Message{}
     |> Message.changeset(params)
+    |> Changeset.cast(params, [:author_id, :project_id])
+    |> Changeset.validate_required([:author_id, :project_id])
+    |> Changeset.assoc_constraint(:author)
+    |> Changeset.assoc_constraint(:project)
+    |> Changeset.cast_assoc(:conversations, with: &Messages.Conversations.create_changeset/2)
     |> Repo.insert()
   end
 
--- a/lib/code_corps_web/channels/conversation_channel.ex
+++ b/lib/code_corps_web/channels/conversation_channel.ex
@@ -1,0 +1,34 @@
+defmodule CodeCorpsWeb.ConversationChannel do
+  use Phoenix.Channel
+
+  alias CodeCorps.{Conversation, Policy, Repo, User}
+  alias Phoenix.Socket
+
+  @spec join(String.t, map, Socket.t) :: {:ok, Socket.t} | {:error, map}
+  def join("conversation:" <> id, %{}, %Socket{} = socket) do
+    with %Conversation{} = conversation <- Conversation |> Repo.get(id),
+         %User{} = current_user <- socket.assigns[:current_user],
+         {:ok, :authorized} <- current_user |> Policy.authorize(:show, conversation, %{}) do
+
+        {:ok, socket}
+    else
+      nil -> {:error, %{reason: "unauthenticated"}}
+      {:error, :not_authorized} -> {:error, %{reason: "unauthorized"}}
+    end
+  end
+
+  def event("new:conversation-part", socket, message) do
+    broadcast socket, "new:conversation-part", message
+    {:ok, socket}
+  end
+
+  def broadcast_new_conversation_part(conversation_part) do
+    channel = "conversation:#{conversation_part.conversation_id}"
+    event = "new:conversation-part"
+    payload = %{
+      id: conversation_part.id
+    }
+
+    CodeCorpsWeb.Endpoint.broadcast(channel, event, payload)
+  end
+end
--- a/lib/code_corps_web/plugs/data_to_attributes.ex
+++ b/lib/code_corps_web/plugs/data_to_attributes.ex
@@ -1,9 +1,11 @@
 defmodule CodeCorpsWeb.Plug.DataToAttributes do
   @moduledoc ~S"""
-  Converts params in the JSON api "data" format into flat params convient for
+  Converts params in the JSON api format into flat params convient for
   changeset casting.
 
-  This is done using `JaSerializer.Params.to_attributes/1`
+  For base parameters, this is done using `JaSerializer.Params.to_attributes/1`
+
+  For included records, this is done using custom code.
   """
 
   alias Plug.Conn
@@ -12,13 +14,38 @@ defmodule CodeCorpsWeb.Plug.DataToAttributes do
   def init(opts), do: opts
 
   @spec call(Conn.t, Keyword.t) :: Plug.Conn.t
-  def call(%Conn{params: %{"data" => data} = params} = conn, _opts) do
+  def call(%Conn{params: %{} = params} = conn, opts \\ []) do
     attributes =
       params
       |> Map.delete("data")
-      |> Map.merge(data |> JaSerializer.Params.to_attributes)
+      |> Map.delete("included")
+      |> Map.merge(params |> parse_data())
+      |> Map.merge(params |> parse_included(opts))
 
     conn |> Map.put(:params, attributes)
   end
-  def call(%Conn{} = conn, _opts), do: conn
+
+  @spec parse_data(map) :: map
+  defp parse_data(%{"data" => data}), do: data |> JaSerializer.Params.to_attributes
+  defp parse_data(%{}), do: %{}
+
+  @spec parse_included(map, Keyword.t) :: map
+  defp parse_included(%{"included" => included}, opts) do
+    included |> Enum.reduce(%{}, fn (%{"data" => %{"type" => type}} = params, parsed) ->
+      attributes = params |> parse_data()
+
+      if opts |> Keyword.get(:includes_many, []) |> Enum.member?(type) do
+        # this is an explicitly specified has_many,
+        # update existing data by adding new record
+        pluralized_type = type |> Inflex.pluralize
+        parsed |> Map.update(pluralized_type, [attributes], fn data ->
+          data ++ [attributes]
+        end)
+      else
+        # this is a belongs to, put a new submap into payload
+        parsed |> Map.put(type, attributes)
+      end
+    end)
+  end
+  defp parse_included(%{}, _opts), do: %{}
 end
--- a/test/lib/code_corps/messages/messages_test.exs
+++ b/test/lib/code_corps/messages/messages_test.exs
@@ -2,10 +2,12 @@ defmodule CodeCorps.MessagesTest do
   @moduledoc false
 
   use CodeCorps.DbAccessCase
+  use Phoenix.ChannelTest
 
   import Ecto.Query, only: [where: 2]
 
   alias CodeCorps.{Conversation, ConversationPart, Message, Messages}
+  alias Ecto.Changeset
 
   defp get_and_sort_ids(records) do
     records |> Enum.map(&Map.get(&1, :id)) |> Enum.sort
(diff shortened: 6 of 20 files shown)


lib/code_corps/messages/conversations.ex
```diff
@@ -0,0 +1,19 @@
...
+  alias CodeCorps.{Conversation}
+
+  @doc false
+  @spec create_changeset(Conversation.t, map) :: Ecto.Changeset.t
+  def create_changeset(%Conversation{} = conversation, %{} = attrs) do
+    conversation
+    |> Changeset.cast(attrs, [:user_id])
+    |> Changeset.validate_required([:user_id])
+    |> Changeset.assoc_constraint(:user)
+  end
```
begedin, line 18: This changeset is needed to `cast_assoc`. The `message_id` is provided automatically by `cast_assoc` and is not part of the params.

lib/code_corps/messages/messages.ex
```diff
@@ -67,9 +67,16 @@ defmodule CodeCorps.Messages do
   def create(%{} = params) do
     %Message{}
     |> Message.changeset(params)
+    |> Changeset.cast(params, [:author_id, :project_id])
+    |> Changeset.validate_required([:author_id, :project_id])
+    |> Changeset.assoc_constraint(:author)
+    |> Changeset.assoc_constraint(:project)
+    |> Changeset.cast_assoc(:conversations, with: &Messages.Conversations.create_changeset/2)
```
begedin, line 74: `create` requires more work than the default `changeset` does. We should probably add more tests for the `create` function to make sure these changes work. I don't think there's an explicit need to extract a create changeset, though.

lib/code_corps_web/plugs/data_to_attributes.ex
```diff
@@ -12,13 +12,30 @@ defmodule CodeCorpsWeb.Plug.DataToAttributes do
...
+  defp parse_included(%{"included" => included}, many_mappings) do
+    included |> Enum.reduce(%{}, fn (%{"data" => %{"type" => type}} = params, parsed) ->
+      attributes = params |> parse_data()
+
+      case many_mappings |> Map.has_key?(type) do
+        true -> parsed |> Map.update(many_mappings[type], [attributes], &List.push(&1, attributes))
+        false -> parsed |> Map.put(type, attributes)
+      end
+    end)
+  end
```
begedin: The critical part of the `included` support is here.

I approached this by following the existing doc for this plug, where it's stated that the goal of the plug is to convert JSON API params into a map of params suitable for `Ecto.Changeset`

From that, the goal was

- Included `belongs_to` records are merged with params as a submap property, with the key identical to the belongs to record type
- Included `has_many` records are merged with params as a list property, with the key being a pluralized version of the record type

This allows us to simply call `cast_assoc(changeset, :record_type, opts)` and the casting works out of the box.

I did not work on inferring the pluralized type directly and instead opted to explicitly provide these when calling the plug. It should be sufficient for our needs, but there is a way to infer instead.

begedin: Since the plug has become quite complex, I think we should add tests for it.

begedin:
I've fixed existing tests.

We still need:

- [ ] tests for `DataToAttributes` plug
- [ ] tests for `Messages.create` to check it's possible to create a `Conversation` alongside
- [ ] tests for controller create action, to serve as integration between `DataToAttributes` and `Messages.create`

lib/code_corps/messages/messages.ex
joshsmith: There's an argument to be made here that this belongs in its own `changeset`, akin to `Messages.Conversations.create_changeset`.

lib/code_corps_web/plugs/data_to_attributes.ex
```diff
@@ -12,13 +12,34 @@ defmodule CodeCorpsWeb.Plug.DataToAttributes do
...
+      attributes = params |> parse_data()
+
+      case many_mappings |> Map.has_key?(type) do
+        true ->
+          parsed
+          |> Map.update(many_mappings[type], attributes, fn (records, record) ->
+            records ++ [record]
+          end)
+        false -> parsed |> Map.put(type, attributes)
+      end
```
joshsmith: Is there any reason to do a `case` here over an `if/else`?

lib/code_corps_web/plugs/data_to_attributes.ex
```diff
@@ -12,13 +12,34 @@ defmodule CodeCorpsWeb.Plug.DataToAttributes do
...
   @spec call(Conn.t, Keyword.t) :: Plug.Conn.t
-  def call(%Conn{params: %{"data" => data} = params} = conn, _opts) do
+  def call(%Conn{params: %{} = params} = conn, many_mappings) do
     attributes =
       params
       |> Map.delete("data")
-      |> Map.merge(data |> JaSerializer.Params.to_attributes)
+      |> Map.delete("included")
+      |> Map.merge(params |> parse_data)
+      |> Map.merge(params |> parse_included(many_mappings))
```
joshsmith: How does this work if `Map.delete("included")`?

lib/code_corps/messages/conversations.ex
joshsmith: This will need a test.

lib/code_corps/messages/messages.ex
snewcomer: @begedin this is the nested POST, right?  I remember doing this once and I had to create a changeset.

I think what I did was take the nested ids, pass it through `|> Enum.map(&Changeset.change/1)` and then `put_assoc(changeset, :conversations, conversation_changesets)` or something like that.  But this looks much better if I am understanding it right.

snewcomer: But that required the nested ids to already exist in the DB.  Not sure what the case is here.

snewcomer (commented):
Don't have the full context of what is going on here.  Do the conversations already exist in the DB when posting the message?

joshsmith:
@snewcomer the hierarchy is like:

```
Message
└── Conversation
    └── ConversationPart
    └── ConversationPart
```

A message could be written with multiple recipients in mind: think of a broadcast message as an example. If we wanted to send an update to all the volunteers, it would be done by creating a message that specifies multiple recipients, all of whom have a conversation created individually for the message.

But a message could also be 1:1, for example individually messaging a user, or a user individually messaging a project. In this case there would only be one conversation.

snewcomer:
@joshsmith gotcha.  On the frontend, the conversation-part is not serialized up though, correct?  At one point does the conversation-part get created? Does that happen completely separate from creating a message?

joshsmith:
@snewcomer yeah, the first conversation part is not the first line of the conversation, which I know is confusing, but necessary due to the `1:n` nature of messages. It's similar to a comment thread of a GitHub issue. The initial part of the thread is actually the issue body, while everything else is specifically a comment.

lib/code_corps_web/plugs/data_to_attributes.ex
joshsmith: I had to fix this fn because `Map.update`'s fun parameter only takes a single value (in this case the accumulated records) and we needed to provide the `cast_assoc` with an array with the default argument if the key doesn't yet exist (i.e. `[attributes]`).

I _think_ this is the logic you were looking for @begedin. Spent quite some time debugging and trying to figure this out. More than I'm happy to admit.

lib/code_corps_web/plugs/data_to_attributes.ex
```diff
@@ -12,13 +12,34 @@ defmodule CodeCorpsWeb.Plug.DataToAttributes do
...
+      attributes = params |> parse_data()
+
+      if many_mappings |> Map.has_key?(type) do
+        parsed
+        |> Map.update(many_mappings[type], [attributes], fn (records) ->
+          records ++ [attributes]
+        end)
+      else
+        parsed |> Map.put(type, attributes)
+      end
```
joshsmith: I had to fix this fn because `Map.update`'s fun parameter only takes a single value (in this case the accumulated records) and we needed to provide the `cast_assoc` with an array with the default argument if the key doesn't yet exist (i.e. `[attributes]`).

I _think_ this is the logic you were looking for @begedin. Spent quite some time debugging and trying to figure this out. More than I'm happy to admit.

lib/code_corps/messages/messages.ex
begedin: @snewcomer The ids are not in the database in this case. The child record is created alongside the parent.

lib/code_corps_web/plugs/data_to_attributes.ex
begedin: Actually, the idea is, if there is a `many_mapping` key, then the included record is for a `has_many` relationship, so it needs to be put into a list. 

If there is no `many_mapping` key, then the included record is for a `belongs_to` relationship, so it needs to be a map.

We don't cast a belongs to this way anywhere yet, so it doesn't matter for now, but that was the intended approach. Had I had the chance to add a test yesterday, this would have been clearer.

begedin: Actually, you did the exact thing that was needed. Sorry, this is difficult to follow

begedin:
Instead of unit testing each individual bit, I opted to integration test on Messages.create/1 and the controller action. The individual bits (changesets) aren't that complex and their behavior is covered with these integration tests.

I think this ought to be good to go for now.

joshsmith:
This will need tests fixed for the has_many added to conversations.

config/dev.exs
```diff
@@ -43,7 +43,7 @@ config :code_corps, CodeCorps.Repo,
   pool_size: 10
 
 # CORS allowed origins
-config :code_corps, allowed_origins: ["http://localhost:4200"]
+config :code_corps, allowed_origins: ["http://localhost:4200", "chrome-extension://pfdhoblngboilpfeibdedpjgfnlcodoo"]
```
joshsmith, line 46: This needs removed.

joshsmith:
Will need tests for channels now.

Also, I think we should be propagating new conversation parts to the `conversation.updatedAt` so we can order conversations by their last updated time rather than their created time. The UI will then prefer the updated time.

begedin:
I'll add the required tests today, but I'd say we're getting to a point where i'd consider the API side mergeable. It does not kill existing features, so there's little reason for holding off and it would make our life easier.

lib/code_corps_web/channels/conversation_channel.ex
```diff
@@ -1,20 +1,21 @@
...
+        {:ok, socket}
     else
-      {:error, %{reason: "unauthorized"}}
+      nil -> {:error, %{reason: "unauthenticated"}}
+      {:error, :not_authorized} -> {:error, %{reason: "unauthorized"}}
     end
   end
-  def join(_, _, _) do
-    {:error, %{reason: "unauthenticated"}}
-  end
```
joshsmith: If the `Repo.get` returns `nil` then we return `{:error, %{reason: "unauthenticated"}}`. I think this changes the behavior we had wanted originally, which is determining whether the user is authenticated or not, hence the pattern match on `assigns`.

test/lib/code_corps/messages/messages_test.exs
```diff
@@ -327,6 +328,21 @@ defmodule CodeCorps.MessagesTest do
...
+        author_id: user.id,
+        body: "Test body",
+        conversation_id: conversation.id
+      }
+
+      CodeCorpsWeb.Endpoint.subscribe("conversation:#{conversation.id}")
+      {:ok, %ConversationPart{id: id}} = Messages.add_part(attrs)
+      assert_broadcast("new:conversation-part", %{id: ^id})
+      CodeCorpsWeb.Endpoint.unsubscribe("conversation:#{conversation.id}")
+    end
```
joshsmith, line 345: 👍

test/support/channel_case.ex
```diff
@@ -20,11 +20,12 @@ defmodule CodeCorpsWeb.ChannelCase do
       # Import conveniences for testing with channels
       use Phoenix.ChannelTest
 
-      alias CodeCorps.Repo
-      import Ecto
-      import Ecto.Changeset
+      import CodeCorps.Factories
+      import CodeCorpsWeb.ChannelCase
```
joshsmith, line 24: Is there a reason this module is being imported into its own `using`? I'm guessing it's because it's importing its own namespace into the tests, but is that necessary if the case is used?

joshsmith (changes requested):
Some questions here.

test/support/channel_case.ex
begedin: It's the general pattern for other `XCase` modules. Allows us to declare helper methods within this module's namespace, which the test cases can then use.

PR #1307, opened 2017-12-19
Track member approval

### LINKED ISSUE #1285: Track whether a user's membership was approved
# Problem

Right now our analytics tracking tracks the user who requested project membership and the user who approved it separately. This means we can't do any kind of conversion funnel analysis here.

We need to rework this, likely by:

- tracking a "Project Membership Approved" event for the approved user

or

- tracking somehow on behalf of the _project_ rather than specific users

iref:
# What's in this PR?

* Added member username and member_id to tracking events on `project_user`,
* Added event properties logging to in-memory analytics api for simpler debugging.

These changes should enable conversion funnel analytics using member_id in queries.

## References
Progress on #1285

### DIFF
--- a/lib/code_corps/analytics/in_memory_api.ex
+++ b/lib/code_corps/analytics/in_memory_api.ex
@@ -9,13 +9,14 @@ defmodule CodeCorps.Analytics.InMemoryAPI do
 
   def identify(user_id, _traits), do: log_identify(user_id)
 
-  def track(user_id, event_name, _properties), do: log_track(user_id, event_name)
+  def track(user_id, event_name, properties), do: log_track(user_id, event_name, properties)
 
   defp log_identify(user_id) do
     Logger.info "Called identify for User #{user_id}"
   end
 
-  defp log_track(user_id, event_name) do
-    Logger.info "Called track for event #{event_name} for User #{user_id}"
+  defp log_track(user_id, event_name, properties) do
+    props = Poison.encode!(properties)
+    Logger.info "Called track for event #{event_name} for User #{user_id} and properties #{props}"
   end
 end
--- a/lib/code_corps/analytics/segment_traits_builder.ex
+++ b/lib/code_corps/analytics/segment_traits_builder.ex
@@ -78,10 +78,12 @@ defmodule CodeCorps.Analytics.SegmentTraitsBuilder do
     }
   end
   defp traits(%CodeCorps.ProjectUser{} = record) do
-    record = record |> Repo.preload(:project)
+    record = record |> Repo.preload([:project, :user])
     %{
       project: record.project.title,
-      project_id: record.project_id
+      project_id: record.project_id,
+      member: record.user.username,
+      member_id: record.user.id
     }
   end
   defp traits(%CodeCorps.StripeConnectAccount{} = account) do
--- a/test/lib/code_corps_web/controllers/project_user_controller_test.exs
+++ b/test/lib/code_corps_web/controllers/project_user_controller_test.exs
@@ -54,7 +54,9 @@ defmodule CodeCorpsWeb.ProjectUserControllerTest do
 
       tracking_properties = %{
         project: project.title,
-        project_id: project.id
+        project_id: project.id,
+        member: user.username,
+        member_id: user.id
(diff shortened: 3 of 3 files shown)


joshsmith:
@iref this is a useful addition in other ways so I want this PR to get merged as-is, though I think it's not quite a fix for #1285.

I've had some extensive conversations with a support engineer at Mixpanel who has enlightened me that we may want to track a `project_n`, e.g. `project_25`, as a distinct id. This means we'd actually rewrite the events as `Membership Requested (Project)` and `Approved Membership (Project)` _from the perspective of the project_. But we would _also_ track a `Requested Membership (User)` and `Membership Approved (User)` from the perspective of the user.

We likely need to tweak the semantics here, but does the gist of what I'm saying make sense?

lib/code_corps/analytics/in_memory_api.ex
```diff
@@ -9,13 +9,14 @@ defmodule CodeCorps.Analytics.InMemoryAPI do
...
 
   defp log_identify(user_id) do
     Logger.info "Called identify for User #{user_id}"
   end
 
-  defp log_track(user_id, event_name) do
-    Logger.info "Called track for event #{event_name} for User #{user_id}"
+  defp log_track(user_id, event_name, properties) do
+    props = Poison.encode!(properties)
+    Logger.info "Called track for event #{event_name} for User #{user_id} and properties #{props}"
```
joshsmith, line 20: Good call logging the properties.

lib/code_corps/analytics/segment_traits_builder.ex
```diff
@@ -78,10 +78,12 @@ defmodule CodeCorps.Analytics.SegmentTraitsBuilder do
...
   end
   defp traits(%CodeCorps.ProjectUser{} = record) do
-    record = record |> Repo.preload(:project)
+    record = record |> Repo.preload([:project, :user])
     %{
       project: record.project.title,
-      project_id: record.project_id
+      project_id: record.project_id,
+      member_id: record.user.id,
+      member: record.user.username
```
joshsmith: Would reverse these to keep with consistency of below: `member` and then `member_id`.

iref:
If I understand it correctly, we would like to emit two events on membership request/approval. One from user perspective (this one already exists) and one from project perspective. Both events have same properties.

Can we extract `project_id` from plug connection or resource and set it as segment api `user_id` when emitting `Membership Requested (Project)` and `Approved Membership (Project)` events?
I am not that familiar with Segment API, so I am not sure if this is correct way how to do this.

joshsmith:
I don't think we'd extract if from the `conn` but from the `%ProjectUser{project_id: project_id}`.

joshsmith:
And to clarify @iref, you're right about wanting to emit multiple events.

Without being confusing, here's the message from Mixpanel (modified slightly to fix some errors I spotted):

> Thanks for your response! It doesn't hurt to track more properties than fewer in this case. I'd suggest tracking the following:
> - Event: User Requests
>   - `distinct_id` = the user's `distinct_id`
>   - `project` = project they want to join
> - Event: Project Accepts Request
>   - `distinct_id` = the project's `distinct_id`
>   - `project` = project they want to join
>   - `accepter` = the user who accepted the request
>
> That way, you could track this for a single user: the user who requested to join and when the user's request was accepted. If you want to track the user accepting that other user's request to join, you could trigger another event to document this and have the `distinct_id` be the accepter. The caveat to this is that you'd have 2x events for when a user's request is accepted, but you'd be able to capture this in our existing funnel.

In our case the user's `distinct_id` would be `id`, as it has been, but the project's `distinct_id` would be `project_n`, e.g. `project_1`. This is necessary because otherwise the analytics tools (like Mixpanel) would confuse `1` for the user with `id` of `1`.

All that make sense?

joshsmith:
If you want to squash your commits on this PR down to just a single commit, then I'm happy to merge this as-is since this is a good changeset, and we can continue with the bigger change in another PR.

iref:
Yes, I think I get it now. I'll start to work on new PR that implements suggested changes.

joshsmith (approved):

PR #1309, opened 2017-12-19
Update conversation route

### LINKED ISSUE #1308: Add Conversations update route
# Problem

A conversation needs to cast the status from open to closed or closed to open.

## Subtasks
- [ ] Add update route
- [ ] Add policy for update

snewcomer:
Fixes #1308

### DIFF
--- a/lib/code_corps/messages/messages.ex
+++ b/lib/code_corps/messages/messages.ex
@@ -53,6 +53,10 @@ defmodule CodeCorps.Messages do
     Conversation |> Repo.get(id)
   end
 
+  def update_conversation(conversation, params) do
+    conversation |> Conversation.update_changeset(params) |> Repo.update
+  end
+
   @doc ~S"""
   Gets a `CodeCorps.ConversationPart` record
   """
--- a/lib/code_corps/policy/conversation.ex
+++ b/lib/code_corps/policy/conversation.ex
@@ -41,4 +41,10 @@ defmodule CodeCorps.Policy.Conversation do
     conversation |> get_message() |> get_project() |> administered_by?(user)
   end
   def show?(_, _), do: false
+
+  def update?(%User{admin: true}, _conversation), do: true
+  def update?(%User{} = user, %Conversation{} = conversation) do
+    conversation |> get_message() |> get_project() |> administered_by?(user)
+  end
+  def update?(_, _), do: false
 end
--- a/lib/code_corps_web/controllers/conversation_controller.ex
+++ b/lib/code_corps_web/controllers/conversation_controller.ex
@@ -29,6 +29,18 @@ defmodule CodeCorpsWeb.ConversationController do
     end
   end
 
+  @spec update(Conn.t, map) :: Conn.t
+  def update(%Conn{} = conn, %{"id" => id} = params) do
+    with %Conversation{} = conversation <- Messages.get_conversation(id) |> preload(),
+         %User{} = current_user <- conn |> CodeCorps.Guardian.Plug.current_resource,
+         {:ok, :authorized} <- current_user |> Policy.authorize(:update, conversation),
+         {:ok, %Conversation{} = updated_conversation} <- conversation |> Messages.update_conversation(params)
+      do
+
+      conn |> render("show.json-api", data: updated_conversation)
+    end
+  end
+
   @preloads [:conversation_parts, :message, :user]
 
   def preload(data) do
--- a/test/lib/code_corps/policy/conversation_test.exs
+++ b/test/lib/code_corps/policy/conversation_test.exs
@@ -1,7 +1,7 @@
 defmodule CodeCorps.Policy.ConversationTest do
   use CodeCorps.PolicyCase
 
-  import CodeCorps.Policy.Conversation, only: [scope: 2, show?: 2]
+  import CodeCorps.Policy.Conversation, only: [scope: 2, show?: 2, update?: 2]
 
   alias CodeCorps.{Conversation, Repo}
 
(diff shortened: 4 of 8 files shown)


lib/code_corps/policy/conversation.ex
```diff
@@ -32,4 +32,19 @@ defmodule CodeCorps.Policy.Conversation do
...
+    # scoped_message_ids =
+    #   Message
+    #   |> Policy.Message.scope(user)
+    #   |> select([m], m.id)
+    #   |> Repo.all
+
+    # conversation
+    # |> where(user_id: ^id)
+    # |> or_where([c], c.message_id in ^scoped_message_ids)
+    true
```
snewcomer: @joshsmith what are the rules here when updating a conversations?

joshsmith: Good question. I think it should be updated really only by someone who's an admin or the message's project is administered by a user.

The `read_at` in a conversation is going to be set by an update to the conversation_part, so it won't happen explicitly. The only explicit update to a conversation _should_ be the `status`, which should be admin-only.

snewcomer: lmk what you think!  updated everything.

snewcomer:
Actually not sure why my commits were changed but tons of merge conflicts

joshsmith:
@snewcomer I had rebased for you. Sorry about that. Was trying to be helpful but was not.

snewcomer:
@joshsmith ah no worries!  I thought I did something.  Does this need tracking?  Not totally sure what the strategy as it relates to tracking....

joshsmith:
We can circle back to figure out the tracking/analytics side of this later. I haven't really formed a coherent strategy around it yet for this feature set.

lib/code_corps_web/controllers/conversation_controller.ex
```diff
@@ -29,6 +29,18 @@ defmodule CodeCorpsWeb.ConversationController do
     end
   end
 
+  @spec update(Conn.t, map) :: Conn.t
+  def update(%Conn{} = conn, %{"id" => id} = params) do
+    with %Conversation{} = conversation <- Messages.get_conversation(id) |> preload(),
+         %User{} = current_user <- conn |> CodeCorps.Guardian.Plug.current_resource,
+         {:ok, :authorized} <- current_user |> Policy.authorize(:update, conversation),
+         {:ok, %Conversation{} = updated_conversation} <- conversation |> Conversation.update_changeset(params) |> Repo.update
```
joshsmith: We should make use of the context here, so instead use `Messages.update_conversation(conversation, params)`.

lib/code_corps/policy/conversation.ex
```diff
@@ -41,4 +41,7 @@ defmodule CodeCorps.Policy.Conversation do
     conversation |> get_message() |> get_project() |> administered_by?(user)
   end
   def show?(_, _), do: false
+
+  def update?(%User{admin: true}, _conversation), do: true
+  def update?(_, _), do: false
```
joshsmith: This also needs the following clause:

```elixir
def update(%User{} = user, %Conversation{} = conversation) do
  conversation |> get_message() |> get_project() |> administered_by?(user)
end
```

test/lib/code_corps/policy/conversation_test.exs
```diff
@@ -113,4 +113,22 @@ defmodule CodeCorps.Policy.ConversationTest do
...
+      assert update?(user, conversation)
+    end
+
+    test "returns false when user is not an admin" do
+      user = insert(:user, admin: false)
+      message = insert(:message)
+      conversation = insert(:conversation, message: message, user: user)
+
+      refute update?(user, conversation)
+    end
```
joshsmith: This will need tests added like in the `show?` tests above once the policy is updated per the above comment.

test/lib/code_corps_web/controllers/conversation_controller_test.exs
```diff
@@ -65,4 +65,37 @@ defmodule CodeCorpsWeb.ConversationControllerTest do
       assert conn |> request_show(conversation) |> json_response(403)
     end
   end
+
+  describe "update" do
+    @tag authenticated: :admin
+    test "updates and renders chosen resource when data is valid", %{conn: conn, current_user: current_user} do
+      message = insert(:message)
+      conversation = insert(:conversation, user: current_user, message: message)
```
joshsmith: This test will need modified using logic similar to the `index`'s `conversation_on_user_administered_project` when the policy is changed, rather than relying on `admin` here.

joshsmith (changes requested):
I think I was unclear about the policy earlier. My more detailed notes here should help.

snewcomer:
@joshsmith fixed!  Thanks for reviewing it all!

lib/code_corps/messages/messages.ex
```diff
@@ -53,6 +53,10 @@ defmodule CodeCorps.Messages do
     Conversation |> Repo.get(id)
   end
 
+  def update_conversation(conversation, params) do
+    Conversation.update_changeset(conversation, params) |> Repo.update
+  end
```
joshsmith: Style thing but I would consider leading with the pipe, so `conversation |> Conversation.update_changeset`...

test/lib/code_corps_web/controllers/conversation_controller_test.exs
```diff
@@ -68,13 +68,16 @@ defmodule CodeCorpsWeb.ConversationControllerTest do
...
     @tag authenticated: :admin
-    test "updates and renders chosen resource when data is valid", %{conn: conn, current_user: current_user} do
-      message = insert(:message)
-      conversation = insert(:conversation, user: current_user, message: message)
+    test "updates and renders chosen resource when data is valid", %{conn: conn, current_user: user} do
+      %{project: project} = insert(:project_user, role: "admin", user: user)
+      message_on_user_administered_project = insert(:message, project: project)
+      conversation_on_user_administered_project =
+        insert(:conversation, message: message_on_user_administered_project)
+      assert conn |> request_update(conversation_on_user_administered_project, %{status: "wat"}) |> json_response(422)
```
joshsmith: It looks like you're doing a `422` assert in here. This line probably needs removed.

joshsmith (approved):

joshsmith:
🙌

PR #1321, opened 2017-12-20
Add new message emails

### LINKED ISSUE #1311: Send email to user for new messages / conversation parts
# Problem

In advance of #1283, we can make a simpler version where for all new messages / conversation parts, we send an email notification automatically.

We can later refine this by adding in some read_at logic and a send delay.

joshsmith:
# What's in this PR?

Adds emails for messages.

## References
Fixes #1311

### DIFF
--- a/lib/code_corps/messages/emails.ex
+++ b/lib/code_corps/messages/emails.ex
@@ -1,0 +1,70 @@
+defmodule CodeCorps.Messages.Emails do
+  @moduledoc """
+  Handles email notifications used within the Messages context
+  """
+  alias CodeCorps.{
+    ConversationPart,
+    Emails,
+    Mailer,
+    Message,
+    Repo
+  }
+
+  @message_preloads [:project, [conversations: :user]]
+
+  @doc ~S"""
+  Notifies all the recipients of a new `CodeCorps.Message`.
+
+  Target recipients are found in the `user` relationship of each
+  `CodeCorps.Conversation`.
+  """
+  @spec notify_message_targets(Message.t) :: :ok
+  def notify_message_targets(%Message{initiated_by: "admin"} = message) do
+    message = message |> Repo.preload(@message_preloads)
+
+    message
+    |> Map.get(:conversations)
+    |> Enum.map(&Emails.MessageInitiatedByProjectEmail.create(message, &1))
+    |> Enum.each(&Mailer.deliver_now/1)
+  end
+
+  @part_preloads [
+    :author,
+    conversation: [
+      [conversation_parts: :author],
+      [message: [:author, [project: :organization]]],
+      :user
+    ]
+  ]
+
+  @doc ~S"""
+  Notifies users via email when a `CodeCorps.ConversationPart` has been added
+  to a `CodeCorps.Conversation`.
+
+  Sends to users participating in the conversation, excluding the author of the
+  conversation part.
+  """
+  @spec notify_of_new_reply(ConversationPart.t) :: :ok
+  def notify_of_new_reply(%ConversationPart{} = part) do
+    part = part |> Repo.preload(@part_preloads)
+    part |> send_reply_to_conversation_emails()
+  end
+
+  @spec send_reply_to_conversation_emails(ConversationPart.t) :: :ok
+  defp send_reply_to_conversation_emails(%ConversationPart{} = part) do
+    part
+    |> get_conversation_participants()
+    |> Enum.map(&Emails.ReplyToConversationEmail.create(part, &1))
+    |> Enum.each(&Mailer.deliver_now/1)
+  end
+
+  @spec get_conversation_participants(ConversationPart.t) :: list(User.t)
+  defp get_conversation_participants(%ConversationPart{author_id: author_id} = part) do
+    part.conversation.conversation_parts
+    |> Enum.map(&Map.get(&1, :author))
+    |> Enum.concat([part.conversation.user])
+    |> Enum.concat([part.conversation.message.author])
+    |> Enum.reject(fn u -> u.id == author_id end)
+    |> Enum.uniq()
+  end
+end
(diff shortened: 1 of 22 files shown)


lib/code_corps/emails/message_for_project_email.ex
```diff
@@ -0,0 +1,43 @@
...
+    %Conversation{} = conversation,
+    %User{} = user) do
+
+    BaseEmail.create
+    |> to(user.email)
+    |> template(template_id(), %{
+      conversation_url: project |> conversation_url(conversation),
+      name: user.first_name,
+      project_title: project.title,
+      subject: "#{user.first_name} has posted a reply on #{project.title}"
```
begedin: Different from the user email primarily because this could be a reply by any participant, including another admin/owner. 

Feel free to word it differently.

lib/code_corps/messages/emails.ex
```diff
@@ -0,0 +1,93 @@
...
+  @doc ~S"""
+  Notifies all relevant users of a new `CodeCorps.Message` being created.
+
+  We match against admin initiated messages only, because it's unclear who
+  message.author is for user-initiated messages, or who the email recipient
+  would be, assuming the message.author is the user who initiated the message.
+  """
+  @spec notify_of_new_message(Message.t) :: :ok
+  def notify_of_new_message(%Message{initiated_by: "admin"} = message) do
+    message = message |> Repo.preload(@message_preloads)
```
begedin: These explicit preloads are ugly, but it's an "internal" module, so it would be a matter of doing them here, or up higher, within the main messages context.

lib/code_corps/messages/emails.ex
```diff
@@ -0,0 +1,93 @@
...
+    |> Enum.each(&Mailer.deliver_now/1)
+  end
+
+  @part_preloads [
+    conversation: [
+      [conversation_parts: :author],
+      [message: [:author, [project: :organization]]],
+      :user
+    ]
+  ]
```
begedin, line 38: Again, ugly, same as above.

In this case, I might actually prefer doing an explicit sql query instead.

lib/code_corps/messages/emails.ex
```diff
@@ -0,0 +1,93 @@
...
+    part = part |> Repo.preload(@part_preloads)
+
+    part |> maybe_send_message_for_user_email()
+    part |> send_message_for_project_emails()
+  end
+
+  # we match against admin initiated only, because it's unclear who
+  # message.author is for user-initiated messages, or who the email recipient
+  # would be, assuming the message.author is the user who initiated the message
+  @spec send_message_for_project_emails(ConversationPart.t) :: :ok
```
begedin: As the comment says

- who gets notified of an email sent by a user to a project? I know we don't have it yet, but still
- who gets assigned as the message.author. Having it be identical to conversation.user seems redundant/confusing. Should we allow it blank in that case?

begedin:
@joshsmith The logic is in, but I'm hoping it could be done a bit better. Does post mark allow specifying a separate template "submodel" for each recipient? That would allow us to just fetch a list of recipients and send the project email to all of them in one request.

Sparkpost allows something like that, so I was hoping Postmark would to.

begedin:
@joshsmith Also, I'm not sure if only participants of the conversation should be notified of new replies, or if all project members of admin or higher should be. If we want to change that, `Messages.Emails.get_project_participants/1` is the place to do it.

lib/code_corps/emails/message_for_project_email.ex
joshsmith: I might change some language here but need to think on it.

lib/code_corps/messages/emails.ex
joshsmith: I think our conversation earlier got it right. Can you comment here what we said?

lib/code_corps/emails/message_for_project_email.ex
joshsmith: Maybe:

> "#{user.first_name} replied to a message in #{project.title}"

joshsmith (commented):
This needs the template ids to be added to the config.

This is also generally confusing to me. It looks like we're missing an email for a message from a user to a project (which admittedly we don't have in the front-end yet, so perhaps can come later). But I think the naming conventions here no longer match up with the original intention or the current usage.

It seems like we need three email types:

- conversation replies, which go to all participants in a conversation
- new message to a project (i.e. initiated_by user), which goes to all a project's admins
- new message to some unknown number of `n` users (i.e. initiated_by admin), which goes to all the conversation's `n` users

Given that, we should maybe have emails that are:

- `ReplyToConversationEmail` - currently `MessageForProjectEmail`
- `MessageInitiatedByProjectEmail` - currently `MessageForUserEmail`
- `MessageInitiatedByUserEmail` - currently not implemented

lib/code_corps/messages/emails.ex
begedin: IIRC, the idea was

When a user messages a project, all project admins and owners receive an email.

When a project admin/owner replies to the conversation, from that point on, only participating users receive emails. That means whoever send it initially, whoever replied, and whoever decided to join in at a later time.

begedin:
I think the 3 new types would make sense. In this case, that would mean

`MessageToUserEmail` -> `MessageInitiatedByProjectEmail`
`MessageToProjectEmail` -> `ReplyToConversationEmail`

Then we add `MessageInitiatedByUserEmail`  at a later time

The logic would have to change a bit to. The `Messages.add_part/1` would only send `ReplyToConversationEmail` emails, to all participants, except the author of the conversation.

Also note that with conversation part types, we might have to name it differently and word it a bit differently to.

joshsmith:
I’ll get to work making the modifications then and create an issue to do the `MessageInitiatedByUserEmail`.

joshsmith:
I think the `maybe_send_message_for_user_email` was overthinking what's needed to send a message to all relevant users in a conversation.

For sending a notification for the conversation part, we need to:

- get all the authors for all the conversation parts
- get the author of the message
- get the user for the conversation

That should be everyone who's participating.

begedin:
@joshsmith I figured it would be simpler with the new email types. Agree with the logic 👍

PR #1323, opened 2017-12-20
Add active filter

### LINKED ISSUE #1312: Add query filter for conversation status
# Problem

- We need to change `ConversationQuery.status_filter` to pattern match on `%{"active" => true}` and rename it `active_filter` (likely requiring a change to the client)
- We need to then add a `status_filter` that filters on the open/closed status

snewcomer:
Fixes #1312

### DIFF
--- a/lib/code_corps/messages/conversation_query.ex
+++ b/lib/code_corps/messages/conversation_query.ex
@@ -33,16 +33,16 @@ defmodule CodeCorps.Messages.ConversationQuery do
 
 
   @doc ~S"""
-  Narrows down a `CodeCorps.Conversation` query to return only those records
-  considered to have a specific status.
+  Filters `CodeCorps.Conversation` record queries to return only those
+  considered to be active.
 
-  The status of `active` means that only those records are included which either
-  - belong to a `CodeCorps.Message` initiated by user
-  - belong to a `CodeCorps.Message` initiated by admin, with at least a single
-    reply in the form of a `CodeCorps.ConversationPart`
+  Active conversations belong either:
+  - to a `CodeCorps.Message` initiated by user
+  - to a `CodeCorps.Message` initiated by an admin, with at least a single
+    conversation part
   """
-  @spec status_filter(Queryable.t, map) :: Queryable.t
-  def status_filter(queryable, %{"status" => "active"}) do
+  @spec active_filter(Queryable.t, map) :: Queryable.t
+  def active_filter(queryable, %{"active" => true}) do
     prefiltered_ids = queryable |> select([c], c.id) |> Repo.all
 
     Conversation
@@ -53,5 +53,15 @@ defmodule CodeCorps.Messages.ConversationQuery do
     |> having([_c, m, _cp], m.initiated_by == "user")
     |> or_having([c, m, cp], m.initiated_by == "admin" and count(cp.id) > 0)
   end
-  def status_filter(query, %{}), do: query
+  def active_filter(query, %{}), do: query
+
+  @doc ~S"""
+  Filters `CodeCorps.Conversation` record queries by their status.
+  """
+  @spec status_filter(Queryable.t, map) :: Queryable.t
+  def status_filter(queryable, %{"status" => status}) do
+    queryable
+    |> where([c], c.status == ^status)
+  end
+  def status_filter(query, _), do: query
 end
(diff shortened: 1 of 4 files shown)


lib/code_corps/messages/conversation_query.ex
```diff
@@ -53,5 +53,16 @@ defmodule CodeCorps.Messages.ConversationQuery do
...
+
+  @doc ~S"""
+  Narrows down a `CodeCorps.Conversation` query to return only those records
+  considered to have a specific status.
+  """
+  @spec status_filter(Queryable.t, map) :: Queryable.t
+  def status_filter(queryable, %{"status" => status}) do
+    queryable
+    |> where([c], c.status == ^status)
+  end
```
snewcomer, line 65: This still needs a test

test/lib/code_corps/messages/messages_test.exs
```diff
@@ -180,23 +180,23 @@ defmodule CodeCorps.MessagesTest do
...
       refute conversation_started_by_admin_without_reply.id in result_ids
       assert conversation_started_by_admin_with_reply.id in result_ids
       assert conversation_started_by_user_without_reply.id in result_ids
       assert conversation_started_by_user_with_reply.id in result_ids
 
-      result_ids =
-        Conversation
-        |> Messages.list_conversations(%{"status" => "any"})
-        |> get_and_sort_ids()
+      # result_ids =
```
snewcomer: Should the default be `Repo.all()`....do we want to get all conversations?

lib/code_corps/messages/conversation_query.ex
```diff
@@ -41,8 +41,8 @@ defmodule CodeCorps.Messages.ConversationQuery do
   - belong to a `CodeCorps.Message` initiated by admin, with at least a single
     reply in the form of a `CodeCorps.ConversationPart`
   """
-  @spec status_filter(Queryable.t, map) :: Queryable.t
-  def status_filter(queryable, %{"status" => "active"}) do
+  @spec active_filter(Queryable.t, map) :: Queryable.t
+  def active_filter(queryable, %{"active" => true}) do
```
snewcomer, line 45: @joshsmith is this how active filter should work?  What specifically filters out inactive records?

lib/code_corps/messages/conversation_query.ex
```diff
@@ -53,5 +53,16 @@ defmodule CodeCorps.Messages.ConversationQuery do
     |> having([_c, m, _cp], m.initiated_by == "user")
     |> or_having([c, m, cp], m.initiated_by == "admin" and count(cp.id) > 0)
```
joshsmith, line 54: In response to your question above, this is the query that does it (specifically the `having` and `or_having`), and the documentation above expands on that.

lib/code_corps/messages/conversation_query.ex
```diff
@@ -41,8 +41,8 @@ defmodule CodeCorps.Messages.ConversationQuery do
   - belong to a `CodeCorps.Message` initiated by admin, with at least a single
```
joshsmith: We need to update this documentation.

> Filters `CodeCorps.Conversation` record queries to return only those considered to be active.
>
> Active conversations belong either:

And then update the bullets to remove "belong".

lib/code_corps/messages/conversation_query.ex
```diff
@@ -53,5 +53,16 @@ defmodule CodeCorps.Messages.ConversationQuery do
     |> having([_c, m, _cp], m.initiated_by == "user")
     |> or_having([c, m, cp], m.initiated_by == "admin" and count(cp.id) > 0)
   end
-  def status_filter(query, %{}), do: query
+  def active_filter(query, %{}), do: query
+
+  @doc ~S"""
+  Narrows down a `CodeCorps.Conversation` query to return only those records
+  considered to have a specific status.
```
joshsmith: Would change this to:

> Filters `CodeCorps.Conversation` record queries by their status.

test/lib/code_corps/messages/messages_test.exs
joshsmith: I think the test on L128 still stands: `returns all records by default`.

In this case we just want to test filtering by that status.

joshsmith: And to expand on this, the client really should be responsible for passing the "defaults" in cases where it's not about authorization to see records.

joshsmith (commented):
A few things I'd recommend changing, along with some answers to your questions.

joshsmith (approved):

PR #1325, opened 2017-12-20
Add part_type to conversation_part

### LINKED ISSUE #1318: Add part_type to conversation parts
# Problem

We need to add a `part_type` to the schema for `ConversationPart`.

It should validate inclusion in:

- "comment"
- "note"
- "reopened"
- "closed"

snewcomer:
Fixes #1318

### DIFF
--- a/lib/code_corps/messages/conversation_parts.ex
+++ b/lib/code_corps/messages/conversation_parts.ex
@@ -4,7 +4,8 @@ defmodule CodeCorps.Messages.ConversationParts do
   i.e. a reply to the `CodeCorps.Conversation` by any participant.
   """
 
-  import Ecto.Changeset, only: [assoc_constraint: 2, cast: 3, validate_required: 2]
+  import Ecto.Changeset, only: [assoc_constraint: 2, cast: 3, validate_required: 2,
+                                validate_inclusion: 3]
 
   alias CodeCorps.{
     ConversationPart,
@@ -24,9 +25,14 @@ defmodule CodeCorps.Messages.ConversationParts do
   @spec create_changeset(ConversationPart.t, map) :: Ecto.Changeset.t
   def create_changeset(%ConversationPart{} = conversation_part, attrs) do
     conversation_part
-    |> cast(attrs, [:author_id, :body, :conversation_id])
+    |> cast(attrs, [:author_id, :body, :conversation_id, :part_type])
     |> validate_required([:author_id, :body, :conversation_id])
+    |> validate_inclusion(:part_type, part_types())
     |> assoc_constraint(:author)
     |> assoc_constraint(:conversation)
+  end
+
+  defp part_types do
+    ~w{ closed comment note reopened }
   end
 end
--- a/lib/code_corps/model/conversation_part.ex
+++ b/lib/code_corps/model/conversation_part.ex
@@ -19,6 +19,7 @@ defmodule CodeCorps.ConversationPart do
   schema "conversation_parts" do
     field :body, :string, null: false
     field :read_at, :utc_datetime, null: true
+    field :part_type, :string, default: "comment"
 
     belongs_to :author, CodeCorps.User
     belongs_to :conversation, CodeCorps.Conversation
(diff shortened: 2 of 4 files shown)


lib/code_corps/messages/conversation_parts.ex
```diff
@@ -24,9 +25,14 @@ defmodule CodeCorps.Messages.ConversationParts do
   @spec create_changeset(ConversationPart.t, map) :: Ecto.Changeset.t
   def create_changeset(%ConversationPart{} = conversation_part, attrs) do
     conversation_part
-    |> cast(attrs, [:author_id, :body, :conversation_id])
+    |> cast(attrs, [:author_id, :body, :conversation_id, :part_type])
```
snewcomer, line 28: I'm guessing this will need an update cast as well?  Don't see an update action however.

joshsmith: I don't think we would cast this in an `update`, since the part type should never change.

lib/code_corps/model/conversation_part.ex
```diff
@@ -19,6 +19,7 @@ defmodule CodeCorps.ConversationPart do
   schema "conversation_parts" do
     field :body, :string, null: false
     field :read_at, :utc_datetime, null: true
+    field :part_type, :string
```
joshsmith: I would make this `default: "comment"` so it doesn't have to be specified for most cases.

priv/repo/migrations/20171220154922_add_part_type_to_conversation.exs
```diff
@@ -0,0 +1,9 @@
+defmodule CodeCorps.Repo.Migrations.AddPartTypeToConversation do
+  use Ecto.Migration
+
+  def change do
+    alter table(:conversation_parts) do
+      add :part_type, :string
```
joshsmith: With above comment, should add `default: "comment"`.

test/lib/code_corps/messages/conversation_parts_test.exs
```diff
@@ -8,14 +8,42 @@ defmodule CodeCorps.Messages.ConversationPartsTest do
   }
 
   @valid_attrs %{
-    body: "Test body."
+    body: "Test body.",
+    part_type: "comment"
```
joshsmith: You can remove this line and the tests should still pass, with the changes mentioned above.

joshsmith (changes requested):
Some quick changes and then should be good to go!

lib/code_corps/model/conversation_part.ex
snewcomer: How does it get set to any other type?  If I set it to a default, I can figure out a way to test the part_type since it doesn't get included in the changeset...

test/lib/code_corps/messages/conversation_parts_test.exs
snewcomer: If I set to a default, do I need to do something in the changeset chain to include it in the changes?

joshsmith: No it’ll just get set if it’s not set. Remember Ecto is still just an adapter to SQL.

test/lib/code_corps/messages/conversation_parts_test.exs
```diff
@@ -13,9 +13,44 @@ defmodule CodeCorps.Messages.ConversationPartsTest do
 
   describe "create_changeset/2" do
     test "with valid attributes" do
+      insert(:user, id: 1)
+      insert(:conversation, id: 1)
       attrs = @valid_attrs |> Map.merge(%{author_id: 1, conversation_id: 1})
+      {:ok, conversation} = ConversationParts.create_changeset(%ConversationPart{}, attrs) |> Repo.insert
+      assert conversation.body == "Test body."
+      assert conversation.part_type == "comment"
```
joshsmith, line 21: I think you also probably want to assert the conversation.user_id here is equal to the user.id.

snewcomer: done

joshsmith (approved):

PR #1326, opened 2017-12-20
Added missing reserved routes to slug validator

### LINKED ISSUE #1304: Validate new routes against usernames
# Problem

We have some new resources/routes that have been added to our client and conflict with possible usernames. We should add these to the slug validator.

begedin:
# What's in this PR?

This PR also includes a dynamically generated test which enumerates all routes defined by the API and generates a list of slugs which need to be reserved.

## References

Fixes #1304

### DIFF
--- a/test/lib/code_corps/validators/slug_validator_test.exs
+++ b/test/lib/code_corps/validators/slug_validator_test.exs
@@ -88,6 +88,21 @@ defmodule CodeCorps.Validators.SlugValidatorTest do
     refute changeset.valid?
   end
 
+  test "reserves all api routes" do
+    CodeCorpsWeb.Router.__routes__
+    |> Enum.map(&Map.get(&1, :path))
+    |> Enum.map(&String.split(&1, "/"))
+    |> List.flatten
+    |> Enum.reject(fn fragment -> fragment == "" end)
+    |> Enum.reject(fn fragment -> fragment |> String.at(0) == ":" end)
+    |> Enum.uniq
+    |> Enum.sort
+    |> Enum.each(fn reserved ->
+      changeset = process_slug(reserved)
+      refute changeset.valid?, "#{reserved} should not be allowed as a slug"
+    end)
+  end
+
   defp process_slug(slug) do
     slug
     |> cast_slug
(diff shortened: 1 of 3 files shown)


test/lib/code_corps/validators/slug_validator_test.exs
```diff
@@ -88,6 +88,21 @@ defmodule CodeCorps.Validators.SlugValidatorTest do
     refute changeset.valid?
   end
 
+  test "reserves all api routes" do
+    CodeCorpsWeb.Router.__routes__
```
begedin, line 92: I'm accessing a __private__ function here, but I couldn't find any other way.

test/lib/code_corps/validators/slug_validator_test.exs
```diff
@@ -88,6 +88,21 @@ defmodule CodeCorps.Validators.SlugValidatorTest do
...
+    |> Enum.map(&Map.get(&1, :path))
+    |> Enum.map(&String.split(&1, "/"))
```
begedin, line 94: This means that a route like `/foo/bar/baz` will reserve `foo`, `bar`and `baz`, not just `foo`.

If that's too strict, I can loosen it by getting the first element of each list here.

test/lib/code_corps/validators/slug_validator_test.exs
```diff
@@ -88,6 +88,21 @@ defmodule CodeCorps.Validators.SlugValidatorTest do
...
+    |> List.flatten
+    |> Enum.reject(fn fragment -> fragment == "" end)
+    |> Enum.reject(fn fragment -> fragment |> String.at(0) == ":" end)
+    |> Enum.uniq
+    |> Enum.sort
```
begedin, line 99: Sorting so it's easier to resolve the test failure when it happens.

test/lib/code_corps/validators/slug_validator_test.exs
```diff
@@ -88,6 +88,21 @@ defmodule CodeCorps.Validators.SlugValidatorTest do
...
+    |> Enum.each(fn reserved ->
+      changeset = process_slug(reserved)
+      refute changeset.valid?, "#{reserved} should not be allowed as a slug"
```
begedin, line 102: Specifying a fail message here. Otherwise, the test will fail with just `expected truthy, got nil`, with no information on which slug is failing.

joshsmith (approved):

PR #1332, opened 2017-12-21
Switch to SparkPost

### LINKED ISSUE #1311: Send email to user for new messages / conversation parts

### LINKED ISSUE #1331: Perform switch from Postmark to SparkPost
# Problem

We want to change our email provider to SparkPost, since it better suits our needs.

## Subtasks
- [ ] Implement testing infrastructure
- [ ] Make the switch
- [ ] Create Issue: Optimize email sending to make use of sparkpost's multiple recipient support

## References

Might make #1330 obsolete or require it to be reworded

begedin:
# What's in this PR?

Things to do in this PR:

- [ ] tests for all functions in `CodeCorps.Emails`
- [ ] couple of remaining dialyzer errors. I'm not sure they're fixable as is. Might need to ignore

## References

Fixes #1331 

## Issues to add before merge

- Make template creation/updating idempotent - Single command either updates or creates depending on the template existing or not. - #1334
- Switch reply to conversation to support multiple recipients - #1335
- Make sure all emails use the new name fallback provided by #1306 - #1336
- Clean up `CodeCorps.Emails.Transmissions.Receipt` - all with failure points should go into the event handler performing the email send. `Receipt.build` should not have a failure point. - #1337


begedin:
Process of converting an email:

- convert `lib/code_corps/emails/x_email.ex` to `lib/code_corps/sparkpost/emails/x.ex`
- convert the related test module as well
- add a test case to the converted test module, to ensure sub data has all the required keys - use examples from two existing tests
- find place in code where old email was called and replace with new call
  - update tests failing due to that replacement

lib/code_corps/sparkpost/emails/forgot_password.ex
```diff
@@ -0,0 +1,26 @@
...
+
+  @spec build(User.t, String.t) :: %Transmission{}
+  def build(%User{} = user, token) do
+    %Transmission{
+      content: %Content.TemplateRef{template_id: "forgot-password"},
+      options: %Transmission.Options{inline_css: true},
+      recipients: [%{name: user.first_name, email: user.email}],
+      substitution_data: %{
+        from_name: "Code Corps",
+        from_email: "team@codecorps.org",
```
begedin: These are identical everywhere, as far as I can tell, so maybe we add a helper module which builds the base model. I wouldn't stress about it yet, though.

lib/code_corps/sparkpost/emails/receipt.ex
```diff
@@ -0,0 +1,81 @@
...
+    "https://d3pgew4wbk2vb1.cloudfront.net/emails/images/emoji-1f64c-1f3fe@2x.png",
+    "https://d3pgew4wbk2vb1.cloudfront.net/emails/images/emoji-1f64c-1f3ff@2x.png"
+  ]
+
+  @spec build(StripeConnectCharge.t, Stripe.Invoice.t) :: %Transmission{}
+  def build(%StripeConnectCharge{} = charge, %Stripe.Invoice{} = invoice) do
+    with %StripeConnectCharge{user: %User{} = user} = charge <- Repo.preload(charge, :user),
+         %Project{} = project <- invoice.subscription |> get_project(),
+         {:ok, %DonationGoal{} = current_donation_goal} <- project |> get_current_donation_goal()
+    do
```
begedin: The email building can fail. I think this should go to the event handler. Data should be fetched/preloaded there.

lib/code_corps/sparkpost/tasks.ex
```diff
@@ -0,0 +1,82 @@
...
+      """
+
+    template_content |> String.replace(@linked_style_element, style_element)
+  end
+
+  @spec load_css_file :: String.t
+  defp load_css_file() do
+    File.cwd! |> Path.join("emails") |> Path.join("styles.css") |> File.read!
+  end
+end
```
begedin: This file is probably a dialyzer nightmare right now. Need to clean it up

test/support/sparkpost_helpers.ex
```diff
@@ -0,0 +1,34 @@
...
+    |> (fn template -> Regex.scan(~r/{{\s*[\w_]+\s*}}/, template) end).()
+    |> List.flatten
+    |> Enum.map(&Regex.scan(~r/[\w\.]+/, &1))
+    |> List.flatten
+    |> Enum.uniq
+    |> Enum.map(&String.to_existing_atom/1)
+    |> (fn list -> list -- @reserved_keys end).()
+    |> Enum.concat(@global_keys)
+    |> Enum.sort
+  end
```
begedin, line 33: Not a perfect helper, but should be helfpul

PR #1344, opened 2017-12-26
limit scope to comment part type

### LINKED ISSUE #1322: Scope conversation parts for users who are the target of the conversation
# Problem

Now that we have completed #1318, we have conversation part types that are not intended for a user to see, including a project's notes, the closed/open parts, etc.

We should amend the conversation part policy to scope only by those parts the user should have access to, which as of this writing should be just the `comment` `part_type`.

snewcomer:
## References
Fixes #1322


begedin (approved):
Seems correct to me, and good to go 👍 

Thank you @snewcomer

PR #1351, opened 2017-12-28
User invites

### LINKED ISSUE #1350: Add InvitedUser model, policy, and controller
# Problem

We want to add an `InvitedUser` with the following attributes:

- `email` - string; required
- `role` - string; optional
- `name` - string; optional

And the following relationships:

- `invitee` - optional; eventually set to the `user` created from this invite
- `inviter` - required; the `user` who created the invite
- `project` - optional; the project the invite is for

We’ll need create, update, index and show endpoints.

If inviting to a `project`, the policy should check permissions similar to how `ProjectUser` works.

When the invite flow is complete, ie to claim an invite, we’ll need to:

- create a user
- set the `invited_user`’s `invitee_id`
- create a `project_user` for the `project` with the specified `role`

It may make sense to do this on the user create endpoint by specifying the `InvitedUser` `id` as a relationship.

begedin:
# What's in this PR?

This PR adds API support for `UserInvite` creation and claims.

The Issue this is referring to specifies that we should be adding an `InvitedUser` model, but `UserInvite` seems more consistent in naming, so I opted for that.

## Creation of user  invites

- we have a new controller for this, as well as a new endpoint
- we differentiate between a plain invite and an invite created to invite a user to a project
- when creating a plain invite, the user is authorized to do so if they are the inviter, with no other conditions
- when creating a project invite, the user needs to be an inviter, and additionally, they need to be of a high enough membership in the specified project, in order to invite a user with a specified role
  - pending users and contributors (?) can't do anything
  - admins can invite other pending (not useful) and contributor users
  - owners can do the same as admins and also invite other admin users
- during creation of the invite, we require an email and also validate that a user account with that email doesn't exist already

## Claiming of user invites

This is done through the user controller create action. The user creation was extracted into `Accounts.create/1` and a new, special clause matches against a payload holding an `"invite_id"` key, to differentiate from normal account creation.

When claiming the invite we

 - try to load the invite by specified id - can fail and renders a 404 if it does
 - create the user (this can fail with a validation error if the user has been created regularly since the invite was created)
  - create a project user if a project and role are specified by the invite
  - associate invite with created user through `invitee`

The user controller is where we track any claimed invites.

## Suggested issues to create 

- [ ] Add an extra step to normal account creation, which claims all invites associated with the created user's email
- [ ] Add the same extra step to invite claim account creation, which claims all invites other than the one currently being claimed.
- [ ] Add support for inviting existing users to project

## References

Closes #1350


lib/code_corps/model/user_invite.ex
```diff
@@ -0,0 +1,17 @@
+defmodule CodeCorps.UserInvite do
+  use CodeCorps.Model
+
+  @type t :: %__MODULE__{}
+
+  schema "user_invites" do
+    field :email, :string, null: false
+    field :role, :string, default: "contributor"
```
joshsmith: This shouldn't have a default because it's possible for an invite to _not_ be for a project.

priv/repo/migrations/20171228163712_add_user_invites.exs
```diff
@@ -0,0 +1,17 @@
+defmodule CodeCorps.Repo.Migrations.AddUserInvites do
+  use Ecto.Migration
+
+  def change do
+    create table(:user_invites) do
+      add :email, :string, null: false
+      add :name, :string, null: true
+      add :role, :string, null: true, default: "contributor"
```
joshsmith: Same here re: default.

priv/repo/structure.sql
```diff
@@ -2,8 +2,8 @@
 -- PostgreSQL database dump
 --
 
--- Dumped from database version 10.1
--- Dumped by pg_dump version 10.1
+-- Dumped from database version 10.0
+-- Dumped by pg_dump version 10.0
```
joshsmith, line 6: Can we see if there's a way to avoid the db version numbering while still getting the benefits of the structure.sql?

lib/code_corps/accounts/user_invites.ex
```diff
@@ -0,0 +1,100 @@
...
+  alias Ecto.{Changeset, Multi}
+
+  import Ecto.Query
+
+  @spec create_invite(map) :: {:ok, UserInvite.t} | {:error, Changeset.t}
+  def create_invite(%{} = params) do
+    %UserInvite{}
+    |> Changeset.cast(params, [:email, :name, :role, :inviter_id, :project_id])
+    |> Changeset.validate_required([:email, :inviter_id])
+    |> Changeset.validate_inclusion(:role, ~w(contributor admin owner))
```
joshsmith: Ideally we would source the role types from the `ProjectUser`.

lib/code_corps/accounts/user_invites.ex
```diff
@@ -0,0 +1,100 @@
...
+          ProjectUser
+          |> where(project_id: ^project_id)
+          |> join(:inner, [pu], u in User, pu.user_id == u.id)
+          |> where([_pu, u], u.email == ^email)
+          |> select([pu, _U], count(pu.id))
+          |> Repo.one
+
+        if count > 0 do
+          changeset |> Changeset.add_error(:email, "Already associated with a project member")
+        else
```
joshsmith: I think the requirement wasn't quite clear here: we want to be sure that the email is not associated with a `User`.

lib/code_corps/accounts/user_invites.ex
```diff
@@ -0,0 +1,100 @@
...
+          changeset
+        end
+    end
+  end
+
+  @spec claim_invite(map) :: {:ok, User.t}
+  def claim_invite(%{} = params) do
```
joshsmith: We'll need to accommodate invites for projects and general invites to Code Corps.

lib/code_corps/accounts/user_invites.ex
```diff
@@ -0,0 +1,100 @@
...
+    case UserInvite |> Repo.get(invite_id) |> Repo.preload([:invitee, :project]) do
+      nil -> {:error, :not_found}
+      %UserInvite{} = invite -> {:ok, invite}
+    end
+  end
+  defp load_invite(%{}), do: {:error, :not_found}
+
+  @spec claim_new_user(map) :: {:ok, User.t}
+  defp claim_new_user(%{} = params) do
+    %User{} |> User.registration_changeset(params) |> Repo.insert
```
joshsmith: Not sure if this will change, but in all the 1.6 auto-formatting I've seen everything uses `()`, so `Repo.insert()`.

joshsmith (changes requested):

lib/code_corps/accounts/user_invites.ex
begedin: Several questions then

- will a project admin/owner ever be able to invite an existing codecorps user to their project?
  - would that not create a `UserInvite` and create something else? Maybe a `ProjectUser` record with a status of `"invited"`?

- what happens in the following scenario:
  - invite is created for a "mike@mail.com"
  - later, independently, a user registers with that email
  - even later, that same user tries to claim the invite

lib/code_corps/accounts/user_invites.ex
begedin: Right now, the code sort of does that. It will

- create a user
- join that user with the project, provided the project and role are specified
- assign that user to the assignee

I'm guessing an email will also be involved, which should be pluggable into this process

priv/repo/structure.sql
begedin: I'm not seeing any documented way of doing so with `mix ecto.dump`

joshsmith:
> Several questions then
> 
> - will a project admin/owner ever be able to invite an existing codecorps user to their project?
>   - would that not create a `UserInvite` and create something else? Maybe a `ProjectUser` record with a status of `"invited"`?

Yes, they would be able to invite an existing user. But this would be simply a `ProjectUser` with a different `status`, as you suggested. There wouldn't be a need for a `UserInvite` in this case. I'm not sure of a better name to represent the idea, but the invite is to someone who is _not yet_ a user.

> - what happens in the following scenario:
>   - invite is created for a "mike@mail.com"
>   - later, independently, a user registers with that email
>   - even later, that same user tries to claim the invite

Hadn't thought of this. Ideally upon registration we would look through the `UserInvite` table and create any `ProjectUser` records, thereby "claiming" the invite. And again ideally, we would then let a user know that an invite is claimed when they go to a URL with the invite `id` in the query params, and we would also have a `status` for such a claim that differentiates it from a sign up with the invite context.

lib/code_corps/accounts/accounts.ex
```diff
@@ -6,20 +6,41 @@ defmodule CodeCorps.Accounts do
...
+  @spec create(map) :: {:ok, User.t} | {:error, Changeset.t} | {:error, :invite_not_found}
+  def create(%{"invite_id" => _} = params) do
+    with {:ok, %User{} = user, %UserInvite{} = user_invite} <- params |> Accounts.UserInvites.claim_invite() do
+      user.id |> Analytics.SegmentTracker.track("Claimed User Invite", user_invite)
+      {:ok, user}
+    end
+  end
+  def create(%{} = params) do
+    %User{} |> User.registration_changeset(params) |> Repo.insert()
+  end
```
begedin: We needed to extract creation into a context. I didn't have a better idea other than this multiple clause approach to differentiate between plain account creation and an invite claim.

lib/code_corps/model/project_user.ex
```diff
@@ -55,7 +54,11 @@ defmodule CodeCorps.ProjectUser do
...
   end
 
-  defp roles do
-    ~w{ pending contributor admin owner }
+  @doc ~S"""
+  Returns supported project user role types
+  """
+  @spec roles :: list(String.t())
+  def roles do
+    ~w{pending contributor admin owner}
```
begedin, line 62: Did an auto-formatting of the file as I was making the `roles` function public (to be used when validating a `UserInvite`

lib/code_corps_web/controllers/user_controller.ex
```diff
@@ -28,64 +28,71 @@ defmodule CodeCorpsWeb.UserController do
...
+    :slugged_route,
+    :stripe_connect_subscriptions,
+    :stripe_platform_card,
+    :stripe_platform_customer,
+    :user_categories,
+    :user_roles,
+    :user_skills
   ]
 
   def preload(data) do
```
begedin, line 118: Another auto-formatting of file as I was modifying the `create` function. Actual changes are in the function above.

lib/code_corps/accounts/user_invites.ex
```diff
@@ -0,0 +1,114 @@
...
+        changeset |> Changeset.add_error(:role, "Needs to be specified for a project invite")
+
+      [_role, nil] ->
+        changeset
+        |> Changeset.add_error(:project_id, "Needs to be specified for a project invite")
+
+      [_role, _project_id] ->
+        changeset
+    end
+  end
```
begedin, line 55: I'm hoping there's a shorter way to write this, but couldn't think of anything.

lib/code_corps/accounts/user_invites.ex
```diff
@@ -0,0 +1,114 @@
...
+
+  @spec join_project(User.t(), UserInvite.t()) :: {:ok, ProjectUser.t()} | {:error, Changeset.t()}
+  defp join_project(%User{} = user, %UserInvite{role: role, project: %Project{} = project}) do
+    %ProjectUser{}
+    |> Changeset.change(%{role: role})
+    |> Changeset.put_assoc(:project, project)
+    |> Changeset.put_assoc(:user, user)
+    |> Changeset.unique_constraint(:project, name: :project_users_user_id_project_id_index)
+    |> Repo.insert()
+  end
```
begedin, line 95: Technically, validation can fail due to the `unique_constraint`. However, since the only case that can fail is if the user already exists, and the multi will fail on the previous step in that case, this can never happen.

lib/code_corps/accounts/user_invites.ex
```diff
@@ -0,0 +1,114 @@
...
+      %UserInvite{} = invite -> {:ok, invite}
+    end
+  end
+
+  defp load_invite(%{}), do: {:error, :not_found}
+
+  @spec claim_new_user(map) :: {:ok, User.t()}
+  defp claim_new_user(%{} = params) do
+    %User{} |> User.registration_changeset(params) |> Repo.insert()
+  end
```
begedin, line 85: As we discussed, we assume new user. 

Existing users will eventually be invitable into the project by creating a special type of `ProjectUser`.

Also, as a user registers for an account directly, not via invite, there should be a step where any existing invites are claimed.

lib/code_corps/accounts/user_invites.ex
```diff
@@ -0,0 +1,114 @@
...
+  defp associate_invitee(%UserInvite{invitee: nil} = invite, %User{} = user) do
+    invite
+    |> Changeset.change(%{})
+    |> Changeset.put_assoc(:invitee, user)
+    |> Repo.update()
+  end
+
+  @spec marshall_response(tuple) :: tuple
+  defp marshall_response({:ok, %{user: user, user_invite: user_invite}}) do
+    {:ok, user |> Map.put(:claimed_invites, [user_invite])}
```
begedin, line 109: This path creates a user by claiming a single invite. This and the normal account creation path should be extended further by adding a step where all other invites are also claimed.

However, I think that's a scope for a different PR.

Right now, we just assume a single invite.

lib/code_corps/model/user.ex
```diff
@@ -43,6 +43,7 @@ defmodule CodeCorps.User do
     has_one :github_app_installation, CodeCorps.GithubAppInstallation
     has_one :slugged_route, SluggedRoute
 
+    has_many :claimed_invites, CodeCorps.UserInvite, foreign_key: :invitee_id
```
begedin, line 46: We could add a `:created_invites` to, but it isn't used yet

lib/code_corps/policy/user_invite.ex
```diff
@@ -0,0 +1,42 @@
...
+      when user_id == inviter_id do
+    user_role =
+      params
+      |> Helpers.get_project()
+      |> Helpers.get_membership(user)
+      |> Helpers.get_role()
+
+    new_role = Map.get(params, "role")
+    do_create?(user_role, new_role)
+  end
```
begedin, line 27: Two conditions for the case of a project invite

- user is the inviter
- user has a role with authority to invite a user with the specified role

lib/code_corps/policy/user_invite.ex
```diff
@@ -0,0 +1,42 @@
...
+
+  def create?(%User{id: user_id}, %{"inviter_id" => inviter_id})
+      when user_id == inviter_id,
+      do: true
```
begedin, line 31: One condition for a plain invite

- user is the inviter

lib/code_corps/policy/user_invite.ex
```diff
@@ -0,0 +1,42 @@
...
+
+  def create?(%User{}, %{}), do: false
+
+  @spec do_create?(String.t() | nil, String.t() | nil) :: boolean
+  defp do_create?("admin", "pending"), do: true
+  defp do_create?("admin", "contributor"), do: true
+  defp do_create?("owner", "pending"), do: true
+  defp do_create?("owner", "contributor"), do: true
+  defp do_create?("owner", "admin"), do: true
+  defp do_create?(_, _), do: false
```
begedin, line 41: Could a contributor also invite a "pending" user, or even another contributor?

lib/code_corps_web/controllers/user_controller.ex
```diff
@@ -28,64 +28,81 @@ defmodule CodeCorpsWeb.UserController do
...
+    with {:ok, %User{} = user} <- params |> Accounts.create() do
+      case user |> Map.get(:claimed_invites) do
+        user_invites when is_list(user_invites) ->
+          user_invites |> Enum.map(fn invite ->
+            user.id |> Analytics.SegmentTracker.track("Claimed User Invite", invite)
+          end)
+
+        _other ->
+          nil
+      end
```
begedin: Possibly we can extract this into a `maybe_track_invite_claim`, but other than that, not sure how to improve on this. Note that user creation itself is tracked implicitly via a plug. We should switch that as well at some point, to add clarity.

begedin:
@joshsmith I think this is finally in a mergeable state. There are issues to be created to expand this further, pending feedback, but other than that, I think our initial bases are covered. I'll update the PR description with the current state of the PR

joshsmith:
@begedin remember to explicitly re-request a review from me so that it will show up in my reviews.

begedin:
Additional tracking to implement here:

- When an invite gets created
  - [x] `current_user.id |> track("Created User Invite", user_invite)`
  - [ ] `user_invite.email |> track("User Invited", user_invite)`
  - if invite is for project as well
    - [ ] `current_user.id |> track("Invited to Project", user_invite)`
    - [ ] `user_invite.email |> track("Invited to Project (Invitee)", user_invite)`

- When an invite gets claimed
  - [x] `current_user.id |> track("Claimed User Invite", user_invite)`
  - [ ] `user_invite.email |> track("User Invited", user_invite)`
  - if invite is for project as well
    - [ ] `current_user.id |> track("Invited to Project", user_invite)`
    - [ ] `user_invite.email |> track("Invited to Project (Invitee)", user_invite)`

I actually think the project specific ones could be ignored. For those, we would could have funnels with custom events instead, since the `distinct_id` is the same as the regular event and the difference is in the `project_id` property being present or not.

I'm not sure if we might also need a 

- [ ] `user_invite.project_id |> track("Invited to Project", user_invite)`

begedin:
@joshsmith I got a question on tracking based on info from #1364 

The rest of the code is good for another review.

PR #1357, opened 2017-12-31
Alias user on create

joshsmith:
# What's in this PR?

Allows us to alias the user on creation.

This is a WIP and very much incomplete.

A few of the issues we have with this:

- it's not recommended to alias on the server-side
- we're sending too many events close to the `alias`, leading to a race condition
- it's not clear whether we should be aliasing to the segment ID or the mixpanel ID

### DIFF
--- a/lib/code_corps/analytics/in_memory_api.ex
+++ b/lib/code_corps/analytics/in_memory_api.ex
@@ -7,9 +7,15 @@ defmodule CodeCorps.Analytics.InMemoryAPI do
 
   require Logger
 
+  def alias(user_id, previous_id), do: log_alias(user_id, previous_id)
+
   def identify(user_id, _traits), do: log_identify(user_id)
 
   def track(user_id, event_name, properties), do: log_track(user_id, event_name, properties)
+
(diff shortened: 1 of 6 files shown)


begedin:
Re: 

>- we're sending too many events close to the `alias`, leading to a race condition

We already agreed we should weer of off implicit tracking through `Plugs.Segment` and instead track explicitly from the controller action.

We did this in other places, but it looks like here, we still implicitly track user created using the plug. This may cause some parallelism, which may be what's causing race conditions.

If we moved tracking to be explicit, we may eliminate this, but I'm not sure.

We would do this by 

- removing 

```Elixir
def includes?(:create, %CodeCorps.User{}), do: true
```

from  `lib/code_corps/analytics/segment_tracking_support.ex`

- adding 

```Elixir
user_id |> SegmentTracker.track("Signed Up", user)
```

to the create action

I also think we should add a function clause to `CodeCorps.Analytics.SegmentTracker` which accepts the user struct as the first argument, instead of just the id, but that's a whole other issue.

PR #1362, opened 2018-01-03
Upgrade comeonin in to 4.0 close #1208

### LINKED ISSUE #1208: Add bcrypt_elixir to dep in mix.exs
# 1208

The code_corps_api uses the Comeonin.Bcrypt module, but the bcrypt_elixir module is not included as a dependency in the main mix.exs file.  The comeonin module's mix.exs file does list bcrypt_elixir as a dependency, but that is not sufficient in all configurations (Mac OSX Sierra is a known issue) to properly retrieve and build the bcrypt_elixir module.  Adding the bycrypt_elixir as a dependency at the top level of the project is known to ensure its inclusion.

See the Wiki page here: 

https://github.com/code-corps/code-corps-api/wiki/Building-the-app-%E2%80%93-OSX-%E2%80%93-Potential-issues

for a larger discussion of the issues with the build on Mac OSx Sierra and the Comeonin.Bcrypt module.

snewcomer:
Closes #1208


config/test.exs
```diff
@@ -22,7 +22,6 @@ config :code_corps, site_url: "http://localhost:4200"
 
 # speed up password hashing
 config :comeonin, :bcrypt_log_rounds, 4
-config :comeonin, :pbkdf2_rounds, 1
```
joshsmith, line 25: Can you clarify why this is removed and what it changes?

joshsmith (commented):
Is this all that's needed to actually upgrade? I thought the upgrade process sounded more involved, but maybe it's not?

config/test.exs
snewcomer: Oh I guess I didn't see it used anywhere.  `pbkdf2_rounds` specifically in the app anywhere.

joshsmith: I think it's because there's a default that this overrides. Am I wrong on that? This just makes the password hashing faster.

begedin: It's the recommended approach to speed up hashing in test environment only. However, since we use bcrypt and not pbkdf2, it's unnecessary.

begedin (approved):

begedin:
@joshsmith I can confirm that the changes here are all that's required to upgrade. The only real change was that we need to include the bcrypt package as a dependancy.

PR #1364, opened 2018-01-08
Track project events

### LINKED ISSUE #1285: Track whether a user's membership was approved

iref:
# What's in this PR?

Segment plug tracker now also tracks events for project ids if id is extractable from request resource.
There is more work to be done. Mostly to add `acceptor` property in `Approved Project Membership` event. This change might require passing of optional current user to `SegmentTraitBuilder.build`

## References

Closes #1285
Progress on: #927

### DIFF
--- a/lib/code_corps/analytics/segment_data_extractor.ex
+++ b/lib/code_corps/analytics/segment_data_extractor.ex
@@ -23,4 +23,12 @@ defmodule CodeCorps.Analytics.SegmentDataExtractor do
   def get_user_id(%Plug.Conn{assigns: %{current_user: %CodeCorps.User{id: id}}}, _), do: id
   def get_user_id(_, %CodeCorps.User{id: id}), do: id
   def get_user_id(_, %{user_id: user_id}), do: user_id
+
+  @doc """
+  Tries to extract project id from given resource.
+  Returns `nil` if project id can't be extracted.
+  """
+  @spec get_project_id(CodeCorps.ProjectUser.t) :: String.t | nil
+  def get_project_id(%CodeCorps.ProjectUser{project_id: id}), do: "project_#{id}"
+  def get_project_id(_), do: nil
 end
--- a/lib/code_corps/analytics/segment_data_extractor.ex
+++ b/lib/code_corps/analytics/segment_data_extractor.ex
@@ -7,6 +7,14 @@ defmodule CodeCorps.Analytics.SegmentDataExtractor do
   def get_action(%Plug.Conn{private: %{phoenix_action: action}}), do: action
   def get_action(_), do: nil
 
+  @doc """
+  Tries to extract project id from given resource.
+  Returns `nil` if project id can't be extracted.
+  """
+  @spec get_project_id(CodeCorps.ProjectUser.t) :: String.t | nil
+  def get_project_id(%CodeCorps.ProjectUser{project_id: id}), do: "project_#{id}"
+  def get_project_id(_), do: nil
+
   @spec get_resource(Plug.Conn.t) :: struct
   def get_resource(%Plug.Conn{assigns: %{data: data}}), do: data
   # these are used for delete actions on records that support it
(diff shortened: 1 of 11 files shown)


iref:
@joshsmith please let me know if these changes match what we discussed in #1307 .

lib/code_corps/analytics/segment_data_extractor.ex
```diff
@@ -23,4 +23,12 @@ defmodule CodeCorps.Analytics.SegmentDataExtractor do
...
   def get_user_id(_, %CodeCorps.User{id: id}), do: id
   def get_user_id(_, %{user_id: user_id}), do: user_id
+
+  @doc """
+  Tries to extract project id from given resource.
+  Returns `nil` if project id can't be extracted.
+  """
+  @spec get_project_id(CodeCorps.ProjectUser.t) :: String.t | nil
+  def get_project_id(%CodeCorps.ProjectUser{project_id: id}), do: "project_#{id}"
+  def get_project_id(_), do: nil
```
joshsmith: I would move this up above `get_resource` just to keep things alpha-ordered. We try to generally alpha-order things for quick scannability.

joshsmith (commented):
This is generally moving in the right direction, though refer back to this comment I had about the event names needing to change to be explicit about the directionality:

> This means we'd actually rewrite the events as `Membership Requested (Project)` and `Approved Membership (Project)` from the perspective of the project. But we would also track a `Requested Membership (User)` and `Membership Approved (User)` from the perspective of the user.

iref:
I added direction to tracked events. I moved project user tracking to controller. It allows us to create events name and payloads without changes to analytics api, which wouldn't be possible if we kept tracking in plug
(`SegmentTraitsBuilder.build` function would need action and current user as additional parameters passed from plug) these changes aren't required by other implicitly tracked events and would complicate api. It also helps with #927.

begedin:
@iref That was a good call. We're actually gradually shifting away from the plug and and adding explicit tracking in controllers in other places as well.

lib/code_corps/analytics/segment_data_extractor.ex
begedin: This is resolved, but github did not mark it outdated.

begedin:
Looking at this further, I'm not so sure we should be tracking an event identifying as a project. 

The way I see it, from the project's POV, we should still track which user (project admin or owner), approved the membership, or invited a user. 

We know which project it happened on from the traits themselves. If we suddenly start recording events triggered by users which are actually project records, our segment dashboard will become confusing.

@joshsmith What do you think?

begedin (commented):
Looking at this further, I'm not so sure we should be tracking an event identifying as a project. 

The way I see it, from the project's POV, we should still track which user (project admin or owner), approved the membership, or invited a user. 

We know which project it happened on from the traits themselves. If we suddenly start recording events triggered by users which are actually project records, our segment dashboard will become confusing.

@joshsmith What do you think?

joshsmith:
@begedin you're misunderstanding the problem that led to this. Please read https://github.com/code-corps/code-corps-api/pull/1307#issuecomment-353224704

begedin:
@joshsmith I assumed, incorrectly, that it would be possible to define a funnel which does not revolve around the `user_id`, using custom events, so that's where my confusion comes from.

From what I can tell after looking into it more deeply, that isn't possible, so I guess this is our only option.

Since, with #1351 we also have membership invites to projects, I guess we'll have to do something similar there. I'll add a comment to that PR about how I think we should do it.

iref:
Did you get a change to review changes? Is there anything else I should improve before merge?

begedin:
@iref sorry for the delay on this. We had some things to tace care of before merging this one. It looks good to go to me 👍

iref:
No problem. I just wanted update to make sure that I didn't forget anything. :)

Issue #1092, opened 2017-10-19
Reorganize GitHub modules


joshsmith:
# Problem

Much of the GitHub syncing can be reorganized as follows:

- [x] `CommentSyncer` and `TaskSyncer` need to move out of `Event`; structure could be like: 
  - `sync/issue/issue.ex` `Sync.Issue` (main entry point for handling the syncing of an issue)
  - `sync/issue/body_parser.ex` `Sync.Issue.BodyParser` (parses issue body)
  - `sync/issue/task/task.ex` `Sync.Issue.Task` (syncs the issue to a `Task`)
  - `sync/issue/task/changeset.ex` `Sync.Issue.Task.Changeset` (deals with changesets)
  - `sync/issue/github_issue/github_issue.ex` `Sync.Issue.GithubIssue` (syncs the issue to a `GithubIssue`)
  - etc.
- [x] `ChangesetBuilder`s
  - [x] need to move out event to wherever syncing goes
  - [x] `map` can be removed
  - [x] naming needs to be rethought to reflect that these are changesets for Code Corps models (`Task`, `Comment`, etc)
- [ ] `GithubRepo` installation events need to have `common/repo_finder` (renamed to `repo_linker`)
- [x] `CommentDeleter` needs to move out of `Event`
- [x] `CommentDeleter` needs to delete `GithubComment`
- [x] `ChangesetBuilder`s need `create_` and `update_` functions made public, `build_` fns removed, unit tests added
- [ ] Replace instances of `find_or_init` and `commit` with insert_or_update and find

begedin:
Re:

> `GithubRepo` installation events need to have `common/repo_finder` (renamed to `repo_linker`)

Not sure if this is still applicable.

The main Sync module makes extensive use of the existing `Sync.Utils.RepoFinder`. The approach towards using it is that a sync step is expected to fail if none is found.

The InstallationRepositories event has an internal method for finding a repo, which works differently due to a find or create approach which is necessary here.

The Installation event preloads all Repos as the final step and then processes each of them individually. There is no `find_or_create` in there. We explicitly now which repo is to be created, which updated, which deleted.

The Matched/Unmatched user helper modules do not deal with processing repos.

If we try really hard, we could force the installation repos and installation repo loading into `RepoFinder`, but from my POV, unless I'm missing something, it will make the code less manageable.

Re:

> renamed to `repo_linker`

[...]

PR #1365, opened 2018-01-09
Reorganize GitHub modules

begedin:
# What's in this PR?

Performs some long needed cleanup for some of our Github modules. While this does not tackle everything listed under #1092, I added a comment into the issue where I'm explaining why we might not want to do all of it.

## Done - Changesets

- all changeset modules now follow the same pattern of having a `create_changeset` and `update_changeset` function
- fixed formatting in all of them
- removed redundant reliance on API payloads where applicable - only the Installation changeset actually needs it
- moved them all into a common namespace
- no more passing in unsaved struct as first param for create - shorter function calls, less information hidden

## Also done

Moved `InstallationRepositories` event logic into the `Sync` namespace
Moved `Installation` event logic into the `Sync` namespace

## References

Progress on: #1092 

## Notes

The `Sync` module is moving towards unmanageable. It seems like we're just shifting the weight from one end to the other, instead of distributing it.

Couple of things I suggest

- Split up the `Sync` context into `Sync` and `Event`. We don't need individual event modules, just a single `Event` module with all the `x_event` functions. That being said, I'd wait with this, since everything might change with adding the queue.

- Split out the single `outcome` type into `x_outcome` for each method. While there are some overlaps, the types our mostly different, so it's just hard to keep track when we push them into a single type.

- Same for `marshall_response`. 

- I already did both for `installation_event/1`. It makes it much cleaner, IMO. The output and output type are kept closer together and do not touch on any other.


joshsmith:
@begedin I see you added some commits after your `needs code review` label. Is this awaiting review for sure?

joshsmith:
It also looks like there are test failures with a coverage drop.

begedin:
Found more stuff to cleanup here. Basically, I've decoupled and flattened out the sync module structure.

I'll be able to remove some of the code which is unused now, as well as moving some tests around.

The line count in the sync module is higher now, but the benefit is that we have a much clearer picture on how to break things down into the queue system, for example.

We also have a much clearer picture of what's tested and what's untested. One look at the coverals report should explain that.

Basically, we were unit testing bits, but not all integrated paths are covered properly.

I can continue work by cleaning up unused code and moving tests around/adding tests as part of this PR next week, or we can merge this one and do it separately. It should not be put on the backlog, though, since there will be chunks of unused code in the codebase now, which are bound cause confusion and wasted time down the line.

joshsmith (dismissed):

joshsmith (approved):

begedin:
Created #1367

PR #1371, opened 2018-01-13
Ensure organization user type is downcased

### LINKED ISSUE #1345: Consolidate normalization of github user type field
# Problem

The github user payload has a `type` field which can be `"User"`, `"Bot"` or `"Organization".` 

With `"User"` and `"Bot"`, we normalize these values during conversion into `"user"` and `"bot"`.

However, `"Organization"` remains capitalized and is not changed when inserting/updating into our database. Any case where where the type is being saved as `"Organization"` should be saved as `"organization"` (lowercased).

## Subtasks

- make switch in code so `"Organization"` is normalized into the lowercase `"organization"`
  - in `lib/github/adapters/user.ex`, the function `transform_type/1` is likely the only culprit
- change any tests affected by this
- add an existing data migration, for staging and production, which changes `"Organization"` to `"organization"` on all user records of type `"Organization"`

```elixir
from u in User, where [type: "Organization"]
|> Repo.update_all(set: [type: "organization"])
```

malandrina:
* Update `CodeCorps.GitHub.Adapters.User#transform_type/1` to downcase the "Organization" GitHub
  user type
* Add migration for transforming existing data

Fixes https://github.com/code-corps/code-corps-api/issues/1345

Feedback sought:

- For transforming the data, would a one-off Mix task (or something like that) be preferable to a migration?
- If a migration is preferable, what should the `down` method do in an irreversible migration? My implementation defines an empty `down` method but I'm not sure if that's conventional.
- I wrote the migration query in SQL. Would you prefer a more Elixir-heavy implementation like the one shown in the description of https://github.com/code-corps/code-corps-api/issues/1345?


begedin (changes requested):
@malandrina This looks good to me overall.

On the matter of a down migration, I'd expect we'd just do the opposite transformation.  

The way I see it, the only case where we'd need a down migration is if we're reverting the commit itself.

malandrina:
> On the matter of a down migration, I'd expect we'd just do the opposite transformation.

Ah yes, that makes sense. I amended my previous commit and force-pushed.

begedin (approved):
Nice and clean. Awesome work 👍

joshsmith:
🙌

PR #1381, opened 2018-03-07
Add analytics to track when an installation is created from Github

### LINKED ISSUE #1038: Add analytics tracking for core actions

zacck-zz:
# What's in this PR?
In this PR We are tracking when installations are created from GitHub,  This enables us to track installations that don't begin with the Code Corps website as an entry point. At the moment this is just tracking the installation however later we should add tracking for the event that this user creates an account on the website

## References
Fixes #1038 

Progress on: #1105

### DIFF
--- a/lib/code_corps/github/sync/github_app_installation/github_app_installation.ex
+++ b/lib/code_corps/github/sync/github_app_installation/github_app_installation.ex
@@ -1,7 +1,7 @@
 defmodule CodeCorps.GitHub.Sync.GithubAppInstallation do
   import Ecto.Query
 
-  alias CodeCorps.{GithubAppInstallation, GitHub.Sync, Repo, User}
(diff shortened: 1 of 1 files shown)


begedin (changes requested):
I'm sorry we had a misunderstanding here, but while the tracking is correct, it would preferable if it were to happen as, the end of the transaction that handles the `installation_created` event.

That means, in the `Sync` module, when you have the outcome of  `Repo.transaction(multi)`.

That would ensure we do not track at a point where the process might still fail and makes the behavior as explicit and clear as it can be.

zacck-zz:
cool will handle this

